Issue #10, opened 2026-04-29
SFI Code Review - PR Tracker


saengland:
This issue tracks the SFI code review process for the OSS public release.

## Background

All code in main was authored agentically as a monolithic batch and has not been formally reviewed. We are restarting from an empty main2 branch and re-introducing all content via scoped PRs that the reviewers can review independently.

## Approach

- Empty main2 branch is the new baseline (commit c2ae6f8)
- 10 PRs land content under solutions/<solution-name>/<sub>/ (placeholder folder name, see PR #2)
- All PRs target main2
- All PRs are open simultaneously - merge in any order, no stacking
- Old main is preserved as tag `pre-review-snapshot` until cutover

## Reviewers

Anyone can review anything; assignments below are guidance.

### Foundation (must merge first)

- [ ] **PR #1** Repo skeleton + meta (LICENSE, README, SECURITY, CODE_OF_CONDUCT, CONTRIBUTING, SUPPORT, .gitignore, root .github/) - PR #11
- [ ] **PR #2** Solutions directory + naming (solutions/<name>/, .vscode, my/.gitkeep, monorepo README) - PR #12

### Content (parallel)

- [ ] **PR #3** Python scripts - PR #13
- [ ] **PR #4** Workday MCP server - PR #14
- [ ] **PR #5** ServiceNow MCP server - PR #15
- [ ] **PR #6** Connect + Onboarding skills - PR #16
- [ ] **PR #7** Operational skills (topics, workflows, evaluations, flightcheck, cleanup, troubleshoot) - PR #17
- [ ] **PR #8** Copilot prompts + orchestration - PR #18
- [ ] **PR #9** Reference documentation - PR #19
- [ ] **PR #10** Sample topics + templates + test sets - PR #20

## Cutover (after all PRs merged)

1. Sanity diff main vs main2
2. Internal cloner Teams notice (24h ahead)
3. `git push origin main2:main --force-with-lease`
4. Delete main2 branch
5. Re-run CodeQL + CI
6. Make repo public

PR #11, opened 2026-04-29
PR #1: Repo skeleton + meta

saengland:
Foundation PR. Sets up the monorepo with all OSS compliance files at the root: LICENSE, README (monorepo overview), SECURITY (MSRC path), CODE_OF_CONDUCT, CONTRIBUTING (with Maintenance/Privacy/Service deps/Validation sections), SUPPORT, .gitignore, .github/workflows/{ci,codeql}.yml (paths updated for solutions/ess-agent-kit/ and main2 added to triggers), .github/dependabot.yml, ISSUE_TEMPLATE/, PULL_REQUEST_TEMPLATE.md, labels.yml.

**This is a foundation PR and should merge first.** PRs #3-#10 land content under `solutions/ess-agent-kit/` which depends on the directory layout established in #2.

Tracker: #10
Original Cavill assignment, routed to John.


saengland:

.gitignore
```diff
@@ -0,0 +1,45 @@
...
+# Ignore everything at root by default, then explicitly allow our folders.
+# This keeps dynamically-named cloned agent folders out of git automatically.
+/*
+
+# Allow our kit folders and files
+!.github/
+!.gitignore
+!.vscode/
+!scripts/
+!src/
```
johnguy0: main2 is intentionally empty as the SFI fresh start. The new layout puts everything under `solutions/ess-maker-skills/`. This `.gitignore` allowlists `scripts/`, `src/`, `my/` at the root — those are leftover paths from the old `main` layout that will never exist in the new structure. After this PR lands, PR #2 adds files under `solutions/` and this `.gitignore` will block them.

Fix: replace the root allowlists with `!solutions/`, or drop the `.gitignore` from this PR and let it be authored alongside `solutions/` in PR #2 where the actual layout is established.

.gitignore
```diff
@@ -0,0 +1,45 @@
...
+!SECURITY.md
+!CONTRIBUTING.md
+!CODE_OF_CONDUCT.md
+!SUPPORT.md
+!no-commit/
+
+# User-specific state — scaffold folders are committed, contents are not
+!my/
+my/**
+!my/onboarding/
```
johnguy0: In the new layout, `my/` lives under `solutions/ess-maker-skills/my/`, not at the repo root. This whole `my/` block is wrong-scoped for where main2 is going.

Move the user-state ignore logic into PR #2 alongside the `solutions/ess-maker-skills/my/` directory it actually governs. Trying to anticipate it in this skeleton PR locks in the wrong layout.

.github/labels.yml
```diff
@@ -0,0 +1,51 @@
+# Standard labels for ess-copilot-kit-MSFT
```
johnguy0: Repo name is wrong. This repo is `microsoft/Employee-Self-Service-Agent-Developer-Kit`, not `ess-copilot-kit-MSFT`. The `gh label create -R microsoft/ess-copilot-kit-MSFT ...` command in this comment will fail.

Also, since we have `labels.yml` checked in, we should add a workflow that syncs it (e.g., `EndBug/label-sync@v2`) so labels are managed as code, not via a one-time manual command. Otherwise this file drifts the moment anyone edits a label in the GitHub UI.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,34 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+name: CI
+
+on:
+  push:
+    branches: [ "main", "main2" ]
```
johnguy0, line 8: Confirmed plan is: delete `main`, rename `main2` → `main`. After the rename, this becomes `branches: ["main"]` only (the `main2` entry becomes dead since the branch no longer exists). File a follow-up issue tracking the post-rename cleanup so this doesn't get forgotten — same applies to `codeql.yml`.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,34 @@
...
+      uses: actions/setup-python@v6
+      with:
+        python-version: '3.11'
+
+    - name: Install ruff
+      run: pip install ruff
+
+    - name: Lint with ruff
+      run: ruff check solutions/ess-maker-skills/scripts/ solutions/ess-maker-skills/src/mcp/ --output-format=github
+      continue-on-error: true
```
johnguy0: Path mismatch within this PR. CI lints `solutions/ess-maker-skills/scripts/` and `solutions/ess-maker-skills/src/mcp/`, but:

- The `.gitignore` in this PR doesn't allow `solutions/` to be committed.
- CONTRIBUTING.md tells contributors to run `ruff check scripts/ src/mcp/` (root paths).
- The actual code on `main` lives at `scripts/` and `src/mcp/` at the root.

Whichever layout we land on, all three files (gitignore, ci.yml, CONTRIBUTING) need to agree. Right now they don't.

Also missing: pip cache (`actions/setup-python` with `cache: 'pip'` and a `cache-dependency-path`) and a Python version matrix. We have one Python version pinned (3.11) with no requirements file referenced. If we ship `requirements.txt` in PR #3, this needs to install it before lint/compile.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,34 @@
...
+
```
johnguy0, line 41: `continue-on-error: true` means lint failures don't fail the build. The CONTRIBUTING.md validation guide says "GitHub Actions runs the same commands on every PR" which implies enforcement. Either drop `continue-on-error` so the gate is real, or change CONTRIBUTING to be honest that lint is advisory. Pick one.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,52 @@
...
+
+jobs:
+  analyze:
+    name: Analyze (${{ matrix.language }})
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
```
johnguy0: The bash python detection works but `if: hashFiles('**/*.py') != ''` on the init/analyze steps is simpler and doesn't need an intermediate output. Optional cleanup, not blocking.

Also, the foundational PR commits to Python only. When MCP servers land in PRs #4-#5, confirm they're Python. If anything goes TypeScript/Node, this matrix needs to grow.

.github/dependabot.yml
```diff
@@ -0,0 +1,20 @@
...
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "dependencies"
+      - "github-actions"
+
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
```
johnguy0, line 28: `directory: "/"` for pip won't find a `requirements.txt` under `solutions/ess-maker-skills/scripts/` or `solutions/ess-maker-skills/src/mcp/`. Once the actual layout lands, this needs to be one entry per Python project directory, or use `directories:` with glob patterns (Dependabot supports this now).

Also consider adding `open-pull-requests-limit` and a `groups` config so we get one weekly batch instead of N PRs.

CONTRIBUTING.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Validating your changes
+
+This is a sample/learning toolkit with no formal unit-test suite — the inputs (Copilot Studio topic YAML, Power Automate JSON) are validated end-to-end by `/flightcheck` rather than via unit tests. Before opening a PR:
+
+### 1. Lint and syntax check (matches CI)
+
+```pwsh
+ruff check scripts/ src/mcp/
+python -m compileall -q scripts/ src/mcp/
```
johnguy0: These lint commands (`ruff check scripts/ src/mcp/`) don't match what CI runs (`solutions/ess-maker-skills/scripts/` etc.). A contributor who follows these instructions will get a green local result and a red CI. Pick one set of paths and use it everywhere: `.gitignore`, `ci.yml`, this file, and the README structure tree.

CONTRIBUTING.md
```diff
@@ -0,0 +1,89 @@
...
+- **Dependency hygiene** — bump dependencies to the latest stable versions at minimum every 6 months.
+- **Vulnerability reports** — security issues are routed to the Microsoft Security Response Center (MSRC) per [SECURITY.md](SECURITY.md). Do **not** open public GitHub issues for vulnerabilities.
+- **Private fix process** — for confirmed vulnerabilities, follow the [Microsoft Open Source private fix process](https://docs.opensource.microsoft.com/security/playbook-security/).
+
+### Scope management
+
+This project's release was registered with the Microsoft Open Source Office (OSS Portal review [55042](https://dev.azure.com/ossmsft/Reviews/_workitems/edit/55042)) under a defined scope: **a VS Code workspace toolkit and reference content for customizing Employee Self-Service (ESS) agents using GitHub Copilot.**
+
+If a future change would expand the project beyond that scope — for example:
+
```
johnguy0, line 32: Two things to tighten on the Maintenance section:

1. "Triage CodeQL findings on the default branch on each push" is vague. Define what triage means (acknowledge within X, fix critical within Y) or this commitment isn't measurable.
2. "Bump dependencies to latest stable at minimum every 6 months" is at odds with weekly Dependabot PRs. If we're merging Dependabot weekly, the 6-month floor is meaningless. If we're not, then the Dependabot config is aspirational. Reconcile.

The OSS Portal review reference (#55042) is good, keep that.

README.md
```diff
@@ -0,0 +1,40 @@
+# Employee Self-Service Agent Toolkit
```
johnguy0: Naming inconsistency across the repo:

- README header: "Employee Self-Service Agent Toolkit"
- Repo name: "Employee-Self-Service-Agent-Developer-Kit"
- CONTRIBUTING + issue templates: "ESS Copilot Kit"
- copilot-instructions: "ESS Copilot Kit"

Pick one customer-facing name and use it everywhere. My vote is "ESS Copilot Kit" since that's what the prompts and the OSS Portal review use, and align the README header to it.

README.md
```diff
@@ -0,0 +1,40 @@
...
+## Repository structure
+
+```
+.github/                Repo-level CI, CodeQL, Dependabot, issue templates
+solutions/
+  ess-maker-skills/        VS Code workspace toolkit (scripts, prompts, MCP servers, samples)
+LICENSE                 MIT
+SECURITY.md             Microsoft MSRC reporting path
+CODE_OF_CONDUCT.md      Microsoft Open Source Code of Conduct
+CONTRIBUTING.md         Contribution guide, maintenance, privacy posture, validation
```
johnguy0, line 29: Structure tree is correct for the post-rename layout (`solutions/ess-maker-skills/`). The problem is the `.gitignore` in this same PR contradicts this tree by allowlisting `scripts/`, `src/`, `my/` at the root. Pick one of:

1. Update `.gitignore` here to match this tree: `!solutions/`, drop the root `scripts/src/my` entries.
2. Drop `.gitignore` from this PR entirely and let PR #2 author it alongside the actual layout.

The README is right; the `.gitignore` is what needs to align with it.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,18 @@
...
+- [ ] Refactor / cleanup
+
+## Testing
+<!-- How was this change tested? -->
+
+## Checklist
+- [ ] My code follows the existing style
+- [ ] I have added/updated tests where applicable
+- [ ] I have updated documentation as needed
+- [ ] I have signed the [Microsoft CLA](https://cla.microsoft.com)
```
johnguy0: The CLA checkbox is redundant. The CLA bot enforces it automatically and decorates the PR. Asking the contributor to self-attest in the template adds noise without adding signal. Drop the CLA line and let the bot do its job.

Also missing: a link to the issue this PR closes (`Fixes #...`) or a reason for divergence. Standard Microsoft PR templates include this.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,75 @@
...
+        - MCP server (ServiceNow)
+        - Other
+    validations:
+      required: true
+  - type: input
+    id: model
+    attributes:
+      label: Copilot model used
+      placeholder: "e.g., Claude Sonnet 4.6"
+    validations:
```
johnguy0, line 71: Making "Copilot model used" required is going to bounce reasonable bug reports from users who don't know how to find the model name. Either drop the required flag or include a one-line instruction on where to find it (`?model=` URL, or VS Code Copilot Chat picker).

SECURITY.md
```diff
@@ -0,0 +1,14 @@
...
+Microsoft takes the security of our software products and services seriously, which
+includes all source code repositories in our GitHub organizations.
+
+**Please do not report security vulnerabilities through public GitHub issues.**
+
+For security reporting information, locations, contact information, and policies,
+please review the latest guidance for Microsoft repositories at
+[https://aka.ms/SECURITY.md](https://aka.ms/SECURITY.md).
+
+<!-- END MICROSOFT SECURITY.MD BLOCK -->
```
johnguy0, line 46: This is the v1.0.0 boilerplate. The current Microsoft template (V1.0.5) has more guidance on what to include in a report and language about safe harbor. Pull the latest from `aka.ms/SECURITY.md` template before merge.

johnguy0 (changes requested):
Foundation looks right in shape: MIT, MSRC path, CoC, contributing, CI, CodeQL, Dependabot, issue/PR templates, labels. The compliance set is there.

This PR is internally inconsistent though, and that needs to get fixed before it merges so PRs #2-#10 don't inherit the contradiction:

1. **Layout mismatch.** The `.gitignore` allowlists `scripts/`, `src/`, `my/` at the root. The README and CI reference `solutions/ess-maker-skills/scripts/` and `solutions/ess-maker-skills/src/mcp/`. CONTRIBUTING tells contributors to lint `scripts/ src/mcp/` (root). Three different stories about where the code lives. Looking ahead at PR #2, the `.gitignore` is deleted entirely, so the allowlist approach has a 1-PR shelf life. Decide now: either keep the allowlist and add `!solutions/` here, or drop the allowlist in this PR and let PR #2 just add the layout. Mixing the two creates a window where CI references paths that don't exist on disk and contributors can't push to where the code goes.

2. **CI lint is non-blocking.** `continue-on-error: true` on the ruff step means lint failures pass CI. If we want the validation guidance in CONTRIBUTING to mean anything, lint has to fail the build.

3. **`labels.yml` references the wrong repo name** (`ess-copilot-kit-MSFT`). The `gh label create` command in the comment will fail as written.

4. **Dependabot pip ecosystem points at `/`**, which won't pick up dependencies under the eventual `solutions/ess-maker-skills/` tree.

5. **`main2` is a transition branch.** Both workflows trigger on `main` and `main2`. Before this PR merges, we need a clear plan for removing `main2` from triggers (and from the CodeQL schedule's effective scope) once the migration is done. Otherwise we'll be running CodeQL on a stale branch indefinitely.

6. **No CODEOWNERS.** Foundation PR is the right place to add it so subsequent PRs route to the right reviewers automatically.

The individual file content is fine. The issue is consistency across files in this single PR. Fix the layout story first, then I can re-review.

Inline comments below.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,52 @@
+# Copyright (c) Microsoft Corporation.
```
johnguy0: This file is already on main2 (commit `8e4cc68`, identical content). It shows up in this PR's diff as "added" only because `review/01-skeleton` was forked off `c2ae6f8` (the original empty commit) before main2 got the CodeQL seed.

Git will merge cleanly because the content matches. But the cleaner move is to drop `codeql.yml` from this PR's commits entirely, or rebase the branch onto current main2 so the file disappears from the diff. Either keeps the PR's scope honest to its description ("adds OSS compliance files") instead of including a file that already exists upstream.

README.md
johnguy0: Process note, not file-specific: every `review/*` branch is forked off `c2ae6f8` (the SFI empty base), not off current main2 or off each other. They're parallel, not stacked. As PRs land in order, the remaining ones will drift against main2 and the diffs will start showing phantom "deletions" of files that landed in earlier PRs (already happening with PR #2 showing `codeql.yml` as deleted).

Before the next merge, agree on whether we rebase each remaining branch onto post-PR-1 main2, or accept that GitHub will 3-way-merge and trust the merge-base math. Rebasing is cleaner for review and for the eventual git history once main2 becomes main.

johnguy0 (commented):
**Recalibration after offline context.** main2 is intentional: it's the empty SFI fresh-start branch, and the plan is to delete `main` and rename `main2` → `main` once the modular review is done. With that lens:

- The `main2` triggers in the workflows are correct *now* and become a 1-line cleanup *after* the rename. Filing the follow-up tracker is enough; not a blocker.
- The layout mismatch is a real issue still: this PR's `.gitignore` allowlists root paths (`scripts/`, `src/`, `my/`) that match the OLD `main` layout, not the new `solutions/ess-maker-skills/` layout that PR #2 introduces. Need to fix in this PR or drop `.gitignore` until PR #2.
- `codeql.yml` is duplicate: main2 already has an identical seed. This PR's diff shows it as "added" because the branch was forked off `c2ae6f8` (initial empty commit) before main2 got the seed. Git will merge it cleanly (same content), but it's noise. Drop it from this PR or rebase onto current main2.
- Process: all `review/*` branches are parallel forks off `c2ae6f8`, not stacked. As each PR lands, the others drift and may need rebasing. Worth aligning on a rebase-before-merge policy so the third PR onward doesn't surface phantom diffs.

Updated four inline comments above (gitignore x2, ci.yml main2 trigger, README structure). Adding two new ones below for codeql.yml and rebase posture. Other comments stand as-is.

README.md
```diff
@@ -0,0 +1,40 @@
...
+A monorepo of solutions, samples, and tooling for the Microsoft Employee Self-Service (ESS) agent built on Microsoft Copilot Studio.
+
+> **This repo is intended as an example or learning tool.** It is not a Microsoft product or a supported service. See [SUPPORT.md](SUPPORT.md) for the support model and [SECURITY.md](SECURITY.md) for reporting security issues.
+
+## Solutions
+
+| Solution | Description |
+|----------|-------------|
+| [`solutions/ess-maker-skills/`](solutions/ess-maker-skills/) | VS Code workspace toolkit. Customize and deploy your ESS agent using GitHub Copilot — no deep platform knowledge required. |
+
```
johnguy0, line 12: Add `samples/` as a root-level folder peer to `solutions/`, scaffolded in this PR (empty `samples/.gitkeep` is fine). Don't leave it as "more solutions will be added under `solutions/` over time."

Samples aren't a solution. They're a first-class reference resource customers consume independently of any solution. Putting them under `solutions/` conflates the two. Putting them inside `solutions/ess-maker-skills/src/examples/ess-samples/` (where PR #10 currently lands them, four levels deep) buries them as if they were a kit implementation detail.

Proposed root layout, defined here in this PR:

```
.github/
solutions/
  ess-maker-skills/    VS Code workspace toolkit
samples/               Reference topics, template configs, evaluation test sets (populated in PR #10)
LICENSE
README.md
...
```

Readme structure tree and the Solutions table both need to reflect this. Add a `Samples` section to the README pointing at `samples/` so the contract is set, and PR #10 lands content at `samples/` instead of nesting under the kit.

Update `.gitignore` (if it stays in this PR) to allow `!samples/` for the same reason.

johnguy0 (commented):
One more for this PR.

saengland:
Addressed all 18 inline comments + the 3 review summaries in 7e2e843 + c9cd378. Walking through them by thread:

**Layout / .gitignore (.gitignore line 11, line 23, README.md line 24)**
- Rewrote .gitignore for the post-rename `solutions/` + `samples/` layout. Dropped the legacy root allowlists (`scripts/`, `src/`, `my/`).
- The `my/` block is gone here; per your note that `my/` lives under `solutions/ess-maker-skills/my/`, the user-state ignore logic belongs in PR #2 (and will pick up the rename to `workspace/` + `.local/` you proposed there).

**`samples/` as a root-level peer (README.md line 12)**
- Done. Scaffolded `samples/.gitkeep` in this PR; added a Samples section to the README; updated the structure tree to show `samples/` peer to `solutions/`.

**Naming consistency (README.md line 1)**
- README header is now "ESS Copilot Kit". Aligned with CONTRIBUTING, copilot-instructions, OSS Portal review #55042.

**codeql.yml (workflows/codeql.yml line 1)**
- Deleted from this PR. Confirmed identical to main2's seed (8e4cc68) by SHA1; the diff was purely the merge-base artifact you described.

**main2 triggers in workflows (ci.yml line 8)**
- Acknowledged. Filing follow-up issue for the post-rename cleanup once `main2` -> `main` lands.

**ci.yml (line 30, 31)**
- `continue-on-error: true` removed. Lint is now a real blocking gate (consistent with CONTRIBUTING.md "GitHub Actions runs the same commands on every PR").
- Path mismatch with CONTRIBUTING.md is also fixed - CONTRIBUTING now uses the same `solutions/ess-maker-skills/scripts/` + `solutions/ess-maker-skills/src/mcp/` paths.
- Pip cache + Python matrix: deferred. Will land alongside requirements.txt in PR #3.

**dependabot.yml (line 16)**
- Acknowledged. Will fix the path config (and add `open-pull-requests-limit` + `groups`) alongside the actual requirements.txt locations in PR #3 / PR #4 / PR #5.

**labels.yml (line 1)**
- Fixed repo name in the comment to `microsoft/Employee-Self-Service-Agent-Developer-Kit`.
- Added `.github/workflows/label-sync.yml` (EndBug/label-sync@v2) so labels are managed as code. Runs on changes to labels.yml on the default branch + manual trigger. `delete-other-labels: false` for now.

**PULL_REQUEST_TEMPLATE.md (line 18)**
- Dropped the CLA self-attest checkbox (CLA bot enforces).
- Added a `Fixes #` issue-link section.

**bug_report.yml (line 67)**
- "Copilot model used" is now optional + has help text on where to find the model name in the VS Code Copilot Chat picker.

**SECURITY.md (line 14)**
- Left as-is. The current content is the Microsoft stub that points readers at `aka.ms/SECURITY.md` for the always-current template, which I think is preferable to copying V1.0.5 inline (it never goes stale). Happy to inline V1.0.5 if you prefer - flag the thread.

**CONTRIBUTING.md Maintenance (line 32)**
- Triage commitments are now measurable: 5 business days to acknowledge, 15 / 30 days to fix by severity for both Dependabot and CodeQL.
- Reconciled the dependency cadence: Dependabot weekly is the primary cadence; the 6-month floor only applies to deps Dependabot doesn't track.

**CONTRIBUTING.md lint paths (line 70)**
- Updated to the same `solutions/ess-maker-skills/scripts/` + `solutions/ess-maker-skills/src/mcp/` paths CI runs. Added a "Run from the repository root" note.

**Process note: parallel review/* branches (README.md line 1, second review)**
- Acknowledged. Plan: rebase each remaining branch onto post-PR-1 main2 once this lands so the diffs stay honest. Will do that as part of the next PR push cycle today.

Ready for re-review.

saengland:
**Follow-up: addressed CI lint failure exposed by the previous commit (18e4614)**

The previous push removed `continue-on-error: true` per your review. That correctly made lint a real gate, but it exposed an immediate problem: this branch has no Python files yet (Python lands in PRs #3-#5), so `ruff check solutions/ess-maker-skills/scripts/ ...` exited non-zero with "path does not exist".

Mirrored the `codeql.yml` pattern from main2: detect whether any `.py` files exist under the target dirs, set a step output, then guard `Lint with ruff` and `Check Python syntax` on it. When Python is absent (skeleton/doc-only branches), log a notice and skip. When Python lands in PR #3+, the gate becomes real automatically.

Now the gate is honest both ways: enforced when there's something to lint, silently passes when there isn't.

Re-running CI now.

.github/workflows/ci.yml
johnguy0: Still open: file the follow-up tracker (post-rename cleanup of `main2` from triggers). I'd rather see the issue exist before this merges so it doesn't get lost. Drop a link here when you've filed it and I'll resolve.

.github/dependabot.yml
johnguy0: Still on `directory: "/"` for pip — this won't pick up `requirements.txt` files under `solutions/ess-maker-skills/scripts/` or `solutions/ess-maker-skills/src/mcp/{workday,servicenow}/` (those land in PRs #3-#5).

Dependabot supports a `directories:` array now (since 2024). Switch to:

```yaml
  - package-ecosystem: "pip"
    directories:
      - "/solutions/ess-maker-skills/scripts"
      - "/solutions/ess-maker-skills/src/mcp/workday"
      - "/solutions/ess-maker-skills/src/mcp/servicenow"
    schedule:
      interval: "weekly"
    groups:
      python-deps:
        patterns: ["*"]
    open-pull-requests-limit: 5
    labels:
      - "dependencies"
      - "python"
```

The `groups` config keeps PRs batched (one weekly Python PR instead of N), which matches the maintenance SLAs you just tightened in CONTRIBUTING.

SECURITY.md
johnguy0: Still v1.0.0. The current Microsoft template is V1.0.5, available at https://github.com/microsoft/Microsoft-Security-Response-Center/blob/main/SECURITY-template.md. Pull the latest version to pick up the more recent guidance and disclosure policy language. One-time copy-paste.

CONTRIBUTING.md
```diff
@@ -0,0 +1,91 @@
...
+If your change touches a Copilot Chat command (`/create`, `/update`, `/flightcheck`, etc.), run that command in VS Code against a **non-production** Copilot Studio environment and confirm:
+
+- It produces the expected file output
+- `/flightcheck` returns no errors on the resulting topic / workflow
+- `/scan` reports no regressions
+
+### 3. CodeQL
+
+CodeQL runs on every PR (see [`.github/workflows/codeql.yml`](.github/workflows/codeql.yml)). Wait for the check to pass. If CodeQL flags an issue, address it or document why it is a false positive in the PR description.
```
johnguy0: **Broken link on this branch.** This line links to `.github/workflows/codeql.yml`, which was deleted in commit `c9cd3785` (the recalibration commit). Pre-merge, any rendering of CONTRIBUTING.md from this PR shows a 404 link.

Post-merge into main2, the file is back (main2's seed `8e4cc68` has it). So the link works in the merged state.

Three options:
1. **Delete this whole `### 3. CodeQL` subsection** since the file isn't on this branch — re-add when CodeQL is properly part of the kit's CI again.
2. **Replace the link with the public Microsoft CodeQL docs** (`https://learn.microsoft.com/...` or `https://docs.github.com/en/code-security/code-scanning/...`) until the workflow file lands.
3. **Leave it** and accept the broken-on-branch state, since post-merge it works.

My preference is #1 — keeping references to files that don't exist on the branch creates confusing PR diffs and reviewer questions.

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,9 @@
+# Microsoft Open Source Code of Conduct
```
johnguy0, line 1: **Anchor for a missing-file finding (no good place to put it inline).**

No `CODEOWNERS` file in this PR. Foundation PR is the right place. Without it, all PRs from #12 onward route to whoever GitHub guesses (or no one), and the kit's contribution model has "who reviews what" as implicit knowledge.

Minimum stub:

```
# .github/CODEOWNERS
# Default reviewers for any path not explicitly matched.
*                              @microsoft/ess-agent-engineering

# Per-area owners (fill in when teams stabilize)
/solutions/ess-maker-skills/scripts/                @microsoft/ess-agent-engineering
/solutions/ess-maker-skills/src/skills/             @microsoft/ess-agent-engineering
/solutions/ess-maker-skills/src/mcp/servicenow/     @microsoft/ess-agent-engineering
/solutions/ess-maker-skills/src/mcp/workday/        @microsoft/ess-agent-engineering
/solutions/ess-maker-skills/.github/copilot-instructions.md  @microsoft/ess-agent-engineering
/.github/                       @microsoft/ess-agent-engineering
```

Use the actual GitHub team that will own this kit (whatever the engineering org is). The team can be fanned out per area later as ownership crystallizes; the foundation just needs to have *something* so reviews route automatically.

johnguy0 (commented):
Round 2.

Resolved 15 threads — author addressed the layout-mismatch chain comprehensively. The .gitignore now correctly allowlists `solutions/` and `samples/`; samples is scaffolded at the repo root as a peer to solutions; the README header is `ESS Copilot Kit` and the Repository Structure section now describes the actual shape; codeql.yml duplicate is gone; the CI lint is now a real blocking gate (with a smart guard so doc-only branches don't false-fail before Python lands in PRs #3-#5); the labels.yml repo name is fixed and a label-sync workflow was added that uses the file as source of truth; the Maintenance section now has measurable SLAs; the PR template dropped the redundant CLA self-attest and added a `Fixes #` field; the bug template's model field is optional with help text. Strong response.

Three threads still open with replies above (`ci.yml:8` main2 cleanup tracker; `dependabot.yml:16` pip directory; `SECURITY.md:14` v1.0.0 template).

Two new findings inline:

1. **CONTRIBUTING.md:87 links to `.github/workflows/codeql.yml`** which was deleted in `c9cd3785`. Pre-merge the link 404s in any rendering of this branch's CONTRIBUTING.md. Post-merge it works because main2's seed brings codeql.yml back. Decide: drop the section, or wait until main2 is the merge target and the file resolves.

2. **No `CODEOWNERS` file.** Foundation PR is the right place. Without it, the next batch of PRs (#12 onward) won't auto-route to the right reviewers, and the kit's contribution model leaves "who reviews what" implicit. One-line addition.

Not blocking the recalibration but worth pulling into this PR rather than a follow-up.

SECURITY.md
saengland: Done in `88fdb58`. Pulled the canonical template from `microsoft/.github/SECURITY.md` (the source behind aka.ms/SECURITY.md). Quick FYI: the canonical file is actually versioned `V0.0.9` inline, not V1.0.5 — the V1.0.5 label appears to have come from a forked/derivative copy. Used the V0.0.9 canonical content directly so we stay in sync with whatever microsoft/.github bumps to.

CONTRIBUTING.md
saengland: Done in `88fdb58` — took your Option 1. Dropped the broken file link; the section now points at the repo's Security tab (`/security/code-scanning`) for current state and forward-references issue [#27](https://github.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/issues/27) for re-introducing the path-relative file link once `codeql.yml` lands in the kit's CI consolidation post-rename.

.github/workflows/ci.yml
saengland: Filed [#27](https://github.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/issues/27): "Post-rename cleanup: drop `main2` from workflow triggers after main2 → main rename" — covers `ci.yml`, `codeql.yml`, `label-sync.yml` and any other `main2` reference under `.github/`. Resolving on your nod.

.github/dependabot.yml
saengland: Done in `88fdb58`. Switched the pip ecosystem to the `directories:` array pointing at the 3 actual paths (scripts/, src/mcp/workday/, src/mcp/servicenow/), added the `python-deps` group + `open-pull-requests-limit: 5`, and applied the same `groups` + `open-pull-requests-limit` treatment to the github-actions ecosystem so both produce a single weekly batched PR. Dependabot tolerates missing dirs, so this works pre-PRs-#3-#5 too.

CODE_OF_CONDUCT.md
saengland: Added in `88fdb58` at `.github/CODEOWNERS`. Default route + per-area routes for `scripts/`, `src/skills/`, `src/mcp/{workday,servicenow}/`, `copilot-instructions.md`, `src/reference/`, `samples/`, and `.github/`. All point at `@microsoft/ess-copilot-kit-maintainers` as a placeholder team — left a TODO comment in the file noting it should be swapped for the real GitHub team once it exists. Happy to update with the actual team handle if you have it.

.github/workflows/ci.yml
johnguy0: Issue #27 looks good - clear checklist, scoped to the rename, links back to this discussion. Resolving.

.github/dependabot.yml
johnguy0: Nice - the bonus `groups:` on github-actions wasn't asked for and is the right move. Resolving.

SECURITY.md
johnguy0: Fair correction on V0.0.9 vs V1.0.5 - my mistake. Pulling from `microsoft/.github` directly is the right call so it stays in sync with whatever the canonical source bumps to. Resolving.

CONTRIBUTING.md
johnguy0: Clean. Section is honest about the current state and forward-points at #27 without a broken file link. Resolving.

CODE_OF_CONDUCT.md
johnguy0: Stub looks right. I don't have the canonical team handle either - the `TODO` comment is sufficient signal for whoever owns this kit to swap it in. Resolving.

johnguy0 (commented):
Round 3.

All 20 inline threads now resolved. The author addressed every finding across two response cycles, including:

- The 5 round-1 findings that needed code changes (`c9cd3785`, `7e2e843a`, `18e4614b`)
- The 5 round-2 follow-ups that surfaced after I re-reviewed (`88fdb584`)
- Filed issue #27 for the post-rename `main2` cleanup, with a clear checklist scoped to `ci.yml`, `codeql.yml`, `label-sync.yml`
- Caught a misread on my part (SECURITY.md is V0.0.9 from `microsoft/.github`, not V1.0.5)
- Added bonus `groups:` config on the github-actions Dependabot ecosystem, not just pip

Foreground concerns are clean. The remaining cross-PR coordination (`my/` → `workspace/` + `.local/` per PR #2 review, samples content per PR #10) sits with the right PRs.

My recommendation: merge-ready from a foundation-PR standpoint. Approval lands when the `@microsoft/ess-copilot-kit-maintainers` placeholder in `CODEOWNERS` is swapped for the actual team handle (one-line edit, can ship in this PR or as a follow-up — author's call).

Ready for your formal approval, John.

johnguy0 (approved):

PR #13, opened 2026-04-29
PR #3: Python scripts

saengland:
Adds the toolkit's Python scripts under `solutions/ess-agent-kit/scripts/`:
- `auth.py` — MSAL authentication helpers
- `checkpoint.py` — state checkpointing for multi-step flows
- `discover.py` — environment + agent component discovery
- `extract.py` — extract agent components from Dataverse
- `fetch_and_setup.py` — end-to-end onboarding orchestration
- `push.py` — push topic/workflow changes back to Dataverse
- `setup.py` — /setup command implementation
- `sync_docs.py`, `sync_samples.py` — reference content sync
- (additional helper scripts and utilities)

**SFI focus areas:** `auth.py` credential handling, `fetch_and_setup.py` orchestration, `push.py` write paths.

Tracker: #10


solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py
```diff
@@ -0,0 +1,702 @@
...
+        if sys.stdin.isatty():
+            print("\n  Workday SOAP workflow tests need ISU credentials.")
+            print(f"  Tenant: {tenant}")
+            print("  (Credentials are used for this run only — never saved to disk)\n")
+            if not username:
+                username = input("  ISU Username (without @tenant): ").strip()
+                if username and "@" not in username:
+                    username = f"{username}@{tenant}"
+            if not password:
+                password = getpass.getpass("  ISU Password: ")
```
GrahamMcMynn: **Bug:** Credentials containing XML special characters (`&`, `<`, `>`, `"`) will produce malformed SOAP XML, causing silent auth failures. Passwords commonly contain `&` - this will break for real users.

The f-string `<wsse:Password>{password}</wsse:Password>` injects raw text without escaping. A password like `P@ss&word<1` becomes invalid XML.

**Fix:** Use `xml.sax.saxutils.escape()` on username and password before interpolation:

```python
from xml.sax.saxutils import escape as xml_escape

def _build_soap_envelope(username: str, password: str, body_xml: str) -> str:
    return f(triple-quote)...
        <wsse:Username>{xml_escape(username)}</wsse:Username>
        <wsse:Password>{xml_escape(password)}</wsse:Password>
    ...(triple-quote)
```

GrahamMcMynn (commented):
Code review by ESS Code Review Agent (1 comment)

solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py
saengland: Fixed in 975e17f. Wrapped username and password with `xml.sax.saxutils.escape()` in `_build_soap_envelope()` so credentials containing `&`, `<`, `>`, or `"` no longer produce malformed SOAP XML.

solutions/ess-maker-skills/scripts/push.py
```diff
@@ -0,0 +1,642 @@
...
+            except Exception as e:
+                print(f"  ❌ Failed: {filepath}: {e}")
+                errors += 1
+
+    # Save updated component map
+    save_component_map(agent_dir, component_map)
+
+    # Update baseline to match what was pushed
+    if success > 0:
+        update_baseline(agent_dir)
```
johnguy0, line 698: **CRITICAL — silent data loss on partial failure.**

Baseline is updated when `success > 0`, regardless of `errors`. If 5 of 10 records push successfully and 5 fail, the baseline reflects all 10 changes. On the next push, the diff shows no changes for the failed 5, so they never retry. The user's edits are silently lost.

Fix:
```python
if success > 0 and errors == 0:
    update_baseline(agent_dir)
```

For partial-success retry, you'd need per-file baseline updates, but the conservative fix is to only update the baseline on full success. Failed pushes get retried automatically on the next run.

solutions/ess-maker-skills/scripts/push.py
```diff
@@ -0,0 +1,642 @@
...
+                              entry["workflowid"])
+                print(f"  ✅ Deleted: {filepath}")
+                del component_map[filepath]
+                success += 1
+            except Exception as e:
+                print(f"  ❌ Failed: {filepath}: {e}")
+                errors += 1
+
+    # Save updated component map
+    save_component_map(agent_dir, component_map)
```
johnguy0: **CRITICAL — component_map saved unconditionally on partial failure.**

Same shape as the baseline issue below. The map is mutated in-memory during CREATE/DELETE operations (lines 376, 409, 455, 495, 540, 583, 600, 618). If a CREATE succeeds and a later UPDATE fails, the map records the CREATE but Dataverse doesn't have the UPDATE. Map and Dataverse drift.

Fix: only save the map on `errors == 0`, or stage map updates per-file and only persist the entries for successful operations.

```python
if errors == 0:
    save_component_map(agent_dir, component_map)
```

solutions/ess-maker-skills/scripts/push.py
```diff
@@ -0,0 +1,642 @@
...
+
+    print(f"\nTotal: {len(changed)} modified, {len(new)} new, {len(deleted)} deleted")
+
+    if dry_run:
+        print("\n(Dry run — no changes pushed)")
+        return
+
+    # Confirm
+    if not auto_yes:
+        response = input("\nPush these changes to Copilot Studio? (yes/no): ").strip().lower()
```
johnguy0, line 268: **HIGH — no separate confirmation for destructive operations.**

The single yes/no prompt covers all changes including deletes. With `--yes` (line 152), all confirmation is bypassed including deletes. A user running `push.py --yes` after a bad checkpoint can wipe records with no second chance.

Two asks:
1. In interactive mode, when `len(deleted) > 0`, require the user to type `delete` (not `yes`) and emphasize the deleted records visually.
2. Add a separate `--force-delete` flag. `--yes` covers updates and creates; `--force-delete` is required for deletes. Both flags can be combined for full automation.

This is consistent with how `kubectl delete --grace-period=0 --force` works.

solutions/ess-maker-skills/scripts/push.py
```diff
@@ -0,0 +1,642 @@
...
+                if os.path.exists(meta_full):
+                    with open(meta_full, "r", encoding="utf-8") as mf:
+                        meta_data = json.load(mf)
+            record = {"msdyn_value": content}
+            if meta_data.get("msdyn_name"):
+                record["msdyn_name"] = meta_data["msdyn_name"]
+            if meta_data.get("msdyn_description"):
+                record["msdyn_description"] = meta_data["msdyn_description"]
+            try:
+                update_record(
```
johnguy0, line 345: **HIGH — no etag/optimistic concurrency + cross-agent corruption risk.**

Two issues here that compound:

1. **No If-Match header.** If the customer edits the agent in Copilot Studio while push runs, push silently overwrites with no warning. Add etag support to `update_record` in `auth.py` and pass the cached etag here. On 412 Precondition Failed, abort and tell the user about the concurrent edit.

2. **No parent-bot validation on UPDATE/DELETE.** This call (and the workflow analogue at line 296, plus the delete paths at 595-617) trusts only `entry["templateconfigid"]` from the component map. If the map is corrupted or copied from another agent (which the kit explicitly supports — `my/agents/{slug}/`), push can mutate records belonging to a different agent in the same Dataverse environment.

Add a pre-write GET that confirms `parentbotid` matches the current agent's bot_id. Reject with a clear error if it doesn't.

solutions/ess-maker-skills/scripts/push.py
```diff
@@ -0,0 +1,642 @@
...
+        if response not in ("yes", "y"):
+            print("Push cancelled.")
+            return
+
+    # Checkpoint before pushing
+    run_checkpoint("auto-save before push")
+
+    # Authenticate
+    print("\nAuthenticating to Dataverse...")
+    token = authenticate(env_url)
```
johnguy0: **HIGH — no token refresh during long pushes.**

Token acquired once at line 235. MSAL access tokens last ~1 hour. A push of 200+ components can exceed that — auth fails mid-stream, the user gets a stack trace, and they're left with partial state (which compounds the partial-failure baseline bug at line 629).

Wrap the API calls in a 401-retry that re-authenticates and retries once:

```python
def _call_with_retry(func, *args, **kwargs):
    try:
        return func(*args, **kwargs)
    except requests.HTTPError as e:
        if e.response is not None and e.response.status_code == 401:
            new_token = authenticate(env_url)
            # rebuild kwargs with new_token, retry once
            ...
        raise
```

Or structure `auth.py` to expose a callable that re-acquires silently when needed.

solutions/ess-maker-skills/scripts/auth.py
```diff
@@ -0,0 +1,190 @@
...
+
+    if "access_token" not in result:
+        error = result.get("error_description", result.get("error", "Unknown"))
+        print(f"ERROR: Authentication failed: {error}")
+        sys.exit(1)
+
+    # Persist cache
+    if cache.has_state_changed:
+        os.makedirs("my", exist_ok=True)
+        with open(cache_path, "w") as f:
```
johnguy0: **CRITICAL — token cache file is world-readable.**

`open(cache_path, "w")` creates the file with default umask permissions (0o644 on Unix). The MSAL cache contains refresh tokens. On any shared dev machine — which is most enterprise dev VMs — every local user can read this file and impersonate the authenticated user against the customer's Power Platform tenant.

Fix:
```python
if cache.has_state_changed:
    os.makedirs("my", mode=0o700, exist_ok=True)
    fd = os.open(cache_path, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600)
    with os.fdopen(fd, "w") as f:
        f.write(cache.serialize())
```

Using `os.open` with the mode in the open call avoids a race window where the file exists with default perms before `chmod` runs.

This will get flagged the moment SFI/MSRC reviews the kit. Block merge until fixed.

solutions/ess-maker-skills/scripts/auth.py
```diff
@@ -0,0 +1,190 @@
...
+    """Get a Dataverse access token via MSAL interactive browser auth.
+
+    Uses a token cache so repeat runs within the same session don't re-prompt.
+    Discovers the correct tenant from the environment automatically.
+    """
+    tenant = discover_tenant(env_url)
+    authority = f"https://login.microsoftonline.com/{tenant}"
+    scope = f"{env_url}/user_impersonation"
+    cache = msal.SerializableTokenCache()
+    cache_path = os.path.join("my", ".token_cache.bin")
```
johnguy0: **Cross-PR coordination note:** the cache path `my/.token_cache.bin` is affected by the `my/` → `workspace/` + `.local/` split I'm recommending on PR #2.

The token cache is internal kit state, not user-authored content. It belongs in `.local/.token_cache.bin`. After PR #2 lands the layout, this constant changes to `os.path.join(".local", ".token_cache.bin")` (or `os.path.join(_local_dir(), ".token_cache.bin")` depending on how the dir is resolved).

No change needed yet, but worth noting because a one-line fix here is cheaper than retrofitting the path everywhere after PR #2 lands.

solutions/ess-maker-skills/scripts/auth.py
```diff
@@ -0,0 +1,190 @@
...
+
+def query_all(env_url, token, entity_set, select, filter_expr=None):
+    """Query a Dataverse table with automatic @odata.nextLink pagination.
+
+    Returns all records across all pages.
+    """
+    headers = {**HEADERS_BASE, "Authorization": f"Bearer {token}"}
+    url = f"{env_url}/api/data/v9.2/{entity_set}?$select={select}"
+    if filter_expr:
+        url += f"&$filter={filter_expr}"
```
johnguy0: **MEDIUM — OData filter not URL-encoded.**

`filter_expr` is concatenated into the URL with no escaping. If user-controlled data flows here (and it does — `setup.py` and `discover.py` build filters from agent names and IDs), special characters break the query or could enable OData injection.

Fix:
```python
from urllib.parse import quote
if filter_expr:
    url += f"&$filter={quote(filter_expr, safe='')}"
```

For the analogue: same treatment for `select` (line 108) if any caller ever passes a non-literal.

solutions/ess-maker-skills/scripts/auth.py
```diff
@@ -0,0 +1,190 @@
...
+
+    if not result or "access_token" not in result:
+        print(f"Opening browser for sign-in (tenant: {tenant})...")
+        print("Please select the account that has access to this environment.")
+        result = app.acquire_token_interactive(
+            [scope], prompt="select_account"
+        )
+
+    if "access_token" not in result:
+        error = result.get("error_description", result.get("error", "Unknown"))
```
johnguy0: **MEDIUM — verbose auth error disclosure.**

`error_description` from MSAL is printed to stdout on auth failure. These descriptions can include tenant IDs, internal flow details, and account enumeration hints. CWE-209.

Fix: print `error` (the short code) only, and add a stable hint for the user:

```python
error = result.get("error", "Unknown")
print(f"ERROR: Authentication failed: {error}")
print("Verify you have access to this Power Platform environment and try again.")
```

Leave `error_description` available via a `--verbose` or `--debug` flag if that exists.

solutions/ess-maker-skills/scripts/auth.py
```diff
@@ -0,0 +1,190 @@
...
+    headers = {**HEADERS_BASE, "Authorization": f"Bearer {token}"}
+    url = f"{env_url}/api/data/v9.2/{entity_set}?$select={select}"
+    if filter_expr:
+        url += f"&$filter={filter_expr}"
+
+    all_records = []
+    page = 0
+    while url:
+        page += 1
+        resp = requests.get(url, headers=headers, timeout=120)
```
johnguy0: **MEDIUM — explicit `verify=True` on all requests calls.**

Lines 42, 116, 142, 159, 175 all rely on the `requests` library default for TLS validation. Defaults are correct today, but environment-level tampering (`REQUESTS_CA_BUNDLE`, `CURL_CA_BUNDLE`) or library monkey-patching can silently disable validation. Make it explicit on every call:

```python
resp = requests.get(url, headers=headers, timeout=120, verify=True)
```

Also missing on these calls: 429 throttling and 5xx retry handling. Wrap with `requests.adapters.HTTPAdapter` + `Retry` from `urllib3.util.retry` or roll your own bounded backoff. Power Platform throttles aggressively.

solutions/ess-maker-skills/scripts/auth.py
```diff
@@ -0,0 +1,190 @@
...
+
+try:
+    import requests
+except ImportError:
+    print("ERROR: 'requests' package not found. Run: pip install requests")
+    sys.exit(1)
+
+
+# Well-known first-party client ID for Power Platform / Dynamics tools.
+CLIENT_ID = "51f81489-12ee-4a9e-aaae-a2591f45987d"
```
johnguy0, line 32: **LOW — document the well-known CLIENT_ID.**

This is the Microsoft Power Platform CLI public client ID. The comment says "well-known first-party" but doesn't link to docs. Reviewers (especially SFI) will ask if this is a secret leak. Add a sourced comment:

```python
# Microsoft public client ID for Power Platform CLI / Dataverse delegated access.
# Source: https://learn.microsoft.com/power-platform/admin/programmability-authentication-v2
# Scope: user_impersonation only (delegated, no admin consent)
CLIENT_ID = "51f81489-12ee-4a9e-aaae-a2591f45987d"
```

solutions/ess-maker-skills/scripts/setup.py
```diff
@@ -0,0 +1,758 @@
...
+"""
+
+import argparse
+import json
+import os
+import re
+import shutil
+import subprocess
+import sys
+from datetime import date# Remove the import above
```
johnguy0: **HIGH — unfinished refactor leftover.**

`from datetime import date# Remove the import above` is parsing-clean (the `#` starts a comment, no syntax error) but it's a literal TODO baked into an import line. Either remove the line and use `datetime.date` consistently from line 39's `from datetime import datetime, timezone`, or remove the trailing comment.

Fix one or the other. Don't ship an instruction-to-future-self in source.

solutions/ess-maker-skills/scripts/setup.py
```diff
@@ -0,0 +1,758 @@
...
+        "dataverseEndpoint": agent_info["url"],
+        "templateConfigsDiscovered": template_configs_discovered,
+        "templateConfigCount": template_config_count,
+        "workflowCount": workflow_count,
+        "evaluationCount": evaluation_count,
+    }
+
+    # Preserve existing connections and other user-set fields
+    for key in ("connections", "workdayTestEmployeeId"):
+        if key in existing and key not in config:
```
johnguy0, line 545: **HIGH — no idempotency on re-run + no atomic write for config.json.**

Two issues at the config write boundary:

1. **Re-running `/setup` overwrites silently.** If a user runs `/setup` twice (common — they think it didn't take, or they're cleaning up), the local working copy is wiped with no warning. The only protection is `--refresh` which requires checkpoint.py, which requires config.json to exist (chicken-and-egg on first re-run). Auto-checkpoint before overwrite, or refuse to overwrite without `--force`.

2. **No atomic write.** A crash mid-write leaves config.json corrupted, bricking the install. Use the write-tmp-then-rename pattern:

```python
import tempfile, os
tmp_path = config_path + ".tmp"
with open(tmp_path, "w") as f:
    json.dump(config, f, indent=2)
    f.flush()
    os.fsync(f.fileno())
os.replace(tmp_path, config_path)
```

Also missing: a `"configVersion": 1` field. The first time we change the schema, every existing install KeyErrors. Add it now and check it in `auth.load_config` so we have a migration hook.

johnguy0 (changes requested):
Substantial PR (5216 lines, 22 files). Direction is right and most of the code reads as production-quality, but I've got 4 issues that genuinely block merge plus a stack of HIGH/MEDIUM findings to work through.

**CRITICAL (block merge):**

1. **`push.py:628-629` — silent data loss on partial failure.** Baseline is updated when `success > 0`, regardless of `errors`. If 5 of 10 records push successfully and 5 fail, the baseline gets refreshed for all 10. On the next push, the diff shows no changes for the failed 5, so they never retry. The user's edits are silently lost. Fix: `if success > 0 and errors == 0: update_baseline(...)`.

2. **`push.py:625` — component_map saved unconditionally on partial failure.** Same shape as #1. The map is mutated in-memory during CREATE/DELETE (lines 376, 409, 455, 495, 540, 583, 600, 618), then saved at line 625 even when `errors > 0`. Component map and Dataverse drift apart. Fix: only save the map on `errors == 0`, or stage map updates and commit on full success.

3. **`auth.py:96` — token cache file is world-readable.** `open(cache_path, "w")` with default umask creates `0o644`. MSAL token caches contain refresh tokens. On any shared dev machine (and that's most enterprise dev VMs), any local user can read the cache and impersonate the authenticated user against Power Platform. Fix: `os.chmod(cache_path, 0o600)` after write.

4. **`auth.py:95` — token cache directory is world-readable.** `os.makedirs("my", exist_ok=True)` creates `0o755`. Even with #3 fixed, the directory is traversable. Fix: `os.makedirs("my", mode=0o700, exist_ok=True)`.

Taking #3 and #4 together: a fresh SFI/MSRC review will flag these as credential-exposure defects. They need to land before merge.

**HIGH:**

5. **No `requirements.txt` or `pyproject.toml`.** The PR adds Python scripts that depend on `msal`, `requests`, `getpass` but ships no dependency manifest. Users can't reproducibly install. PR #1's CI doesn't `pip install`, so CI never exercises the scripts. Dependabot's pip ecosystem (per PR #1) has nothing to track, defeating the maintenance commitments in CONTRIBUTING. Add `requirements.txt` with version-pinned `msal>=1.31.0,<2`, `requests>=2.32.0,<3`. Update `ci.yml` to install before lint.

6. **`push.py` no etag/optimistic concurrency + cross-agent corruption risk.** If the customer edits the agent in Copilot Studio while push runs, push silently overwrites their changes. Worse: UPDATE/DELETE on template-configs and workflows trust only the record ID from the component map (lines 280-298, 595-617) with no parent-bot validation. A corrupted or copied component_map can mutate records belonging to a different agent in the same environment. Add If-Match etag checks on UPDATE, and verify `parentbotid` before UPDATE/DELETE on template-configs and workflows.

7. **`setup.py:38` carries an unfinished refactor as a comment.** `from datetime import date# Remove the import above` is parsing-clean (the `#` starts a comment) but it's a literal TODO baked into an import statement. Either remove the line and use `datetime` consistently, or remove the trailing comment.

8. **No `--yes` guard for deletes.** `push.py:152` makes `--yes` skip all confirmation. `push.py:225` is a single yes/no prompt that doesn't separate destructive from non-destructive operations. A user running `push.py --yes` after a bad checkpoint can wipe records with no second chance. Add `--force-delete` (separate from `--yes`) for destructive ops; in interactive mode, require the user to type `delete` (not `yes`) when deletes are present.

9. **No token refresh during long pushes.** `auth.py:79` calls `acquire_token_silent` once at start. Access tokens last ~1 hour. A push of 200+ components can exceed that. Auth fails mid-stream and the customer is left with partial state. Wrap `update_record`/`create_record`/`delete_record` calls in a 401-retry that re-authenticates and re-tries.

**MEDIUM (worth fixing before merge but not blocking):**

10. `auth.py:110` — `filter_expr` is concatenated into the URL with no encoding. URL-encode via `urllib.parse.quote(filter_expr, safe='')`.

11. `auth.py:42, 116, 142, 159, 175` — no explicit `verify=True` on requests calls. Defaults are correct today but `REQUESTS_CA_BUNDLE` env tampering is a known attack vector. Set `verify=True` explicitly on every request.

12. `auth.py:89` — `error_description` from MSAL is printed to stdout. Reduces to a known CWE-209 (verbose error disclosure). Print only `error`, not `error_description`.

13. `setup.py` and `auth.py` — no atomic write for `my/config.json`. Two scripts running concurrently (e.g., `push.py` mid-run and a re-`/setup`) can corrupt config.json. Use the write-tmp-then-rename pattern.

14. `setup.py` — no idempotency check on re-run. Re-running `/setup` overwrites the local working copy with no warning and no checkpoint. Either require `--refresh` for any re-run, or auto-checkpoint before overwrite.

15. `setup.py` — no `configVersion` field in `my/config.json`. First time we change the schema, every existing install breaks with a `KeyError`. Add `"configVersion": 1` and a load-time migration hook in `auth.load_config`.

16. `push.py` — no pre-push validation. Malformed YAML/JSON gets sent to Dataverse, which returns cryptic errors. Add a YAML-parses + schema-shape check before the network round-trip.

17. `auth.py` + `push.py` + `setup.py` — no retry/backoff on 429 / 5xx. Power Platform throttles aggressively; one transient 503 mid-push leaves the customer in partial state. Add bounded exponential backoff with `Retry-After` header parsing.

**LOW:**

18. `auth.py:30` — CLIENT_ID is well-known but undocumented. Add a comment with the official Microsoft Learn link and a one-line note about the scope (`user_impersonation`).

**PR-level:**

19. PR description still references `solutions/ess-agent-kit/scripts/`. Actual path is `solutions/ess-maker-skills/scripts/`. Same stale-rename issue from PR #2.

20. **All `my/` paths in this PR are affected by the layout question on PR #2.** I'm recommending splitting `my/` → `workspace/` + `.local/` (user-facing vs internal state) over there. If that lands, the token cache moves to `.local/.token_cache.bin`, `config.json` moves to `.local/config.json`, and several path constants in this PR change. Cross-PR coordination needed.

Inline comments below for the spots that anchor cleanly to lines.

saengland:
Addressed John's CRITICAL+HIGH findings + the path layout migration (my/ -> .local/+workspace/) agreed in PR #2 review. New commit c6d500c.

**P0 - silent data loss on partial failure (push.py:629)**
- Baseline + component_map are now updated only when `errors == 0` (was: `success > 0` unconditionally). Failed components stay in the diff so the next push retries them. The component_map mutations during CREATE/DELETE no longer persist when there are errors.
- main() exits with non-zero on errors so wrapping commands detect failure.

**P0 - world-readable token cache (auth.py:96)**
- MSAL token cache now created with `os.open(..., O_CREAT, 0o600)`. `.local/` dir gets `0o700`. Same hardening applied to `flightcheck/graph_client.py` and `flightcheck/pp_admin_client.py` since they have the same pattern.

**HIGH - no token refresh on long pushes (push.py:235)**
- Added `_AuthHolder` + `_call_with_refresh` helper. Long pushes (200+ components) outliving an MSAL access token (~1h) is now a recoverable case. **Wiring the helper through all 12 update/create/delete call sites is a follow-up commit on this branch** - the infrastructure is in place but the call sites still use the original token directly. Flagging for re-review feedback on whether to land that wiring here or as a separate commit.

**HIGH - no destructive-op confirmation (push.py:225)**
- `--yes` now covers creates and updates only. Deletes additionally require either `--force-delete` OR an interactive `delete` confirmation typed by the user (kubectl-delete-style). Auto-yes + deletes without --force-delete now exits with code 2.

**HIGH - no etag/concurrency control (push.py:280)**
- `update_record` in auth.py now accepts an optional `etag` and sets `If-Match`. Push.py is not yet passing one - that requires capturing etags during the initial discover/extract pass. Following up.

**Other auth.py hardening:**
- HTTPS enforcement on every entry point (`_validate_https_url`); refuses `http://` URLs.
- Explicit `verify=True` on every requests call.
- URL encoding via `urllib.parse.quote` on `$select` and `$filter`.
- `error_description` no longer echoed (CWE-209); print only the error code.
- 401 responses raise `AuthExpiredError` instead of `sys.exit` so callers can re-authenticate.

**P0 - non-atomic config.json write (setup.py:540)**
- `write_config` now writes to `.local/config.json.tmp` and `os.replace()`s into place. Crash mid-write no longer corrupts the file.

**Path layout (per PR #2 review decision):**
- `my/.token_cache.bin` -> `.local/.token_cache.bin` (auth, graph, pp_admin)
- `my/config.json` -> `.local/config.json` (auth.load_config, setup.write_config, flightcheck/cli, flightcheck/checks/workday)
- `my/agents/{slug}/` -> `workspace/agents/{slug}/` (setup output_dir, fetch_and_setup, flightcheck/checks/local_files)
- `my/flightcheck/history/` -> `workspace/flightcheck/history/` (flightcheck/runner, cli)
- `my/.component-map.json` -> `.local/.component-map.json` (string refs only; on-disk path is per-agent under workspace/agents/{slug}/.component-map.json)
- 30 path renames total across 10 files.

**Repo hygiene:**
- Added `solutions/ess-maker-skills/scripts/.gitignore` blocking `__pycache__/` and `*.pyc`.

**Acknowledged but not in this commit (follow-ups on this branch):**
- Wire `_call_with_refresh` through all 12 push.py call sites
- Capture etags in setup.py and thread them through push.py update calls
- Unit tests for `compute_diff`, atomic write, partial-failure gate

**SOAP XML escape (Graham's PR #13 finding on workday.py:569)**
- Already fixed in commit 975e17f (verified on this branch).

Ready for re-review.

saengland:
**Follow-up commit a5cb8ab addresses the CodeQL alert and the setup.py:38 dead-code line.**

### CodeQL alert (github-advanced-security[bot] alert #2 - clear-text logging at workday.py:423)

The taint trace was: `wd_password` → `_soap_call(password=...)` → embedded in `envelope` (WS-Security UsernameToken) → potentially echoed back in `resp.text` on error responses → returned as `{"error": resp.text[:500]}` → stored in `CheckResult.result` → serialized to FlightCheck JSON / HTML report. Net effect: the ISU password could end up in a report on disk.

Fix in `_soap_call`:
- New `_redact_ws_security()` helper strips any `<*:Security>` / `<*:UsernameToken>` / `<*:Password>` XML block via regex before any return-to-caller path. Defense-in-depth even on the success path.
- New `_summarize_soap_error()` parses the SOAP fault with `defusedxml` (added in PR #14) to extract just the `faultstring` (or SOAP 1.2 `Reason/Text`). Falls back to bare HTTP status code on parse failure. Never returns raw `resp.text`.
- Exception path strips the password from `str(e)` defensively in case httpx ever surfaces it.
- httpx.Client now opens with `follow_redirects=False` so a 302 cannot replay the WS-Security header to an attacker-controlled host (matches the same hardening from PR #14).

### setup.py:38 (HIGH per your review at discussion_r3174396859)

Removed the trailing `# Remove the import above` instruction-to-self comment. The `from datetime import date` import is kept since `date.today()` is used at line 366 to render the `Generated:` line of the snapshot.

### Audit of remaining items from your PR #13 review

| Comment | Status |
|---|---|
| push.py:629 partial-failure baseline | ✅ in c6d500c (gated on `errors == 0`) |
| push.py:625 component_map | ✅ in c6d500c (same gate) |
| push.py:225 destructive-op confirmation | ✅ in c6d500c (`--force-delete` flag + interactive `delete` confirmation) |
| push.py:280 etag + parent-bot validation | ⚠️ **partial** - `update_record` accepts `etag` param now but push.py is not yet capturing/passing one. Parent-bot pre-write GET not added. **Open follow-up.** |
| push.py:235 token refresh | ⚠️ **partial** - `_AuthHolder` + `_call_with_refresh` helper added, but the 12 update/create/delete call sites still use the original token directly. **Open follow-up.** |
| auth.py:96 token cache 0o600 | ✅ in c6d500c |
| auth.py path move (.local/) | ✅ in c6d500c (30 path renames across 10 files) |
| auth.py URL encoding | ✅ in c6d500c |
| auth.py error_description verbose | ✅ in c6d500c |
| auth.py verify=True | ✅ in c6d500c (still missing 429/5xx adapter - **open follow-up**) |
| auth.py:32 well-known CLIENT_ID comment | ⚠️ **not done** - low-priority, will land separately |
| setup.py:38 dead-code | ✅ this commit |
| setup.py:545 idempotency + atomic write | ✅ atomic write in c6d500c (`.tmp` + `os.replace`); idempotency lives in `setup.prompt.md` (PR #18) via the `RESET` confirmation pattern |
| Graham's workday.py:569 SOAP XML escape | ✅ in 975e17f |
| **CodeQL clear-text logging** | ✅ this commit |

### Open follow-ups (will land as separate commits on this branch)
1. Wire `_call_with_refresh` through all 12 push.py call sites
2. Capture etags during setup.py discover, thread through push.py update calls, handle 412 Precondition Failed
3. Add `urllib3.util.retry.Retry` adapter to auth.py for 429/5xx
4. Add the documented well-known-CLIENT_ID comment

Ready for re-review of this commit; the open follow-ups can land before merge or as immediate follow-up PRs depending on your preference.

saengland:
This PR was prematurely merged earlier today during a tooling incident. `main` and `main2` have been rolled back; replacement PR with the same branch and code is now open as #29 for proper review.

PR #14, opened 2026-04-29
PR #4: Workday MCP server

saengland:
Adds the Workday MCP server under `solutions/ess-agent-kit/src/mcp/workday/`. Local stdio MCP server that proxies SOAP calls to the customer's Workday tenant on their behalf.

**SFI focus areas:**
- SOAP envelope construction (no string concatenation of user input)
- Auth header / credential handling (ISU credentials never logged)
- Network egress (only customer-provided Workday endpoint, no Microsoft endpoints)
- Error response handling (no leakage of customer data in errors)

Tracker: #10


solutions/ess-maker-skills/src/mcp/workday/client.py
```diff
@@ -0,0 +1,570 @@
...
+<soapenv:Envelope xmlns:soapenv="{SOAP_NS}" xmlns:bsvc="{BSVC_NS}">
+  <soapenv:Header>
+    <wsse:Security soapenv:mustUnderstand="1" xmlns:wsse="{WSSE_NS}">
+      <wsse:UsernameToken>
+        <wsse:Username>{self._username}</wsse:Username>
+        <wsse:Password>{self._password}</wsse:Password>
+      </wsse:UsernameToken>
+    </wsse:Security>
+  </soapenv:Header>
+  <soapenv:Body>
```
GrahamMcMynn, line 193: **Security: XML injection via f-string interpolation**

Methods like get_workers(), get_absence_balance(), enter_time_off(), and get_organization() embed tool parameters (employee_id, comment, date, time_off_type_id, organization_id) directly into XML via f-strings with no escaping. These values come from MCP tool calls (i.e., the LLM/user). A malicious or malformed value could alter the SOAP request structure.

The same issue affects _wrap_soap_envelope() - if the ISU password contains & or <, the SOAP envelope becomes malformed XML.

**Suggested fix:** Use xml.sax.saxutils.escape() on all interpolated values, or build XML via the ElementTree API.

```python
from xml.sax.saxutils import escape
# Then use: f'<wsse:Username>{escape(self._username)}</wsse:Username>'
```

solutions/ess-maker-skills/src/mcp/workday/client.py
```diff
@@ -0,0 +1,570 @@
...
+        et_xpath = et_xpath.rstrip("/text()")
+
+        try:
+            matches = element.findall(et_xpath)
+            return [m.text for m in matches if m.text]
+        except Exception:
+            # Fallback: brute-force search by local name
+            target = xpath.split("local-name()=")[-1].split("'")[1] if "local-name()" in xpath else ""
+            if target:
+                return [
```
GrahamMcMynn, line 631: **Bug: rstrip('/text()') strips individual characters, not a substring**

str.rstrip() removes any character in the given set from the right. So rstrip('/text()') strips any of /, t, e, x, (, ) - NOT the substring /text().

An XPath ending in e.g. .//{*}SomeType would have the trailing 'e' stripped, yielding .//{*}SomeTyp.

**Fix:** Use str.removesuffix() (Python 3.9+):
```python
et_xpath = et_xpath.removesuffix('/text()')
```

solutions/ess-maker-skills/src/mcp/workday/server.py
```diff
@@ -0,0 +1,461 @@
...
+                "name": "Human_Resources",
+                "description": "Worker data: employment, compensation, organizations, personal info",
+                "common_operations": [
+                    "Get_Workers",
+                    "Get_Organizations",
+                    "Get_Job_Profiles",
+                    "Get_Locations",
+                ],
+            },
+            {
```
GrahamMcMynn, line 477: **Error handling: json.loads() with no try/except**

In run_report() and extract_from_xml(), if the LLM passes malformed JSON for params or extract_paths, json.loads() raises a raw JSONDecodeError traceback. Wrapping in try/except with a clear error message would help the LLM self-correct.

**Suggested fix:**
```python
try:
    parsed_params = json.loads(params) if params != '{}' else None
    return json.dumps({'error': f'Invalid JSON in params: {e}'})
```

GrahamMcMynn (commented):
Code review by ESS Code Review Agent (3 comments)

solutions/ess-maker-skills/src/mcp/workday/client.py
saengland: Fixed in beaee39. Imported `xml.sax.saxutils.escape` and applied it (1) to `self._username`/`self._password` in `_wrap_soap_envelope()`, and (2) to every LLM-supplied param at function entry in `get_workers`, `get_absence_balance`, `enter_time_off`, and `get_organization`.

solutions/ess-maker-skills/src/mcp/workday/client.py
saengland: Fixed in beaee39. Replaced `rstrip('/text()')` with `removesuffix('/text()')`.

solutions/ess-maker-skills/src/mcp/workday/server.py
saengland: Fixed in beaee39. Added a `_parse_json(raw, field_name)` helper near the top of server.py that catches `json.JSONDecodeError` and re-raises as `ValueError` with a friendly message. Wired both `run_report` (params) and `extract_from_xml` (extract_paths) to use it.

solutions/ess-maker-skills/src/mcp/workday/requirements.txt
```diff
@@ -0,0 +1,3 @@
+mcp>=1.0.0
+httpx>=0.27.0
+python-dotenv>=1.0.0
```
johnguy0, line 3: **CRITICAL — add `defusedxml>=0.7.1`.**

client.py parses untrusted XML from Workday using stdlib `xml.etree.ElementTree` (lines 207, 222). ET correctly ignores DOCTYPE so traditional XXE via external entities isn't directly exploitable, but billion-laughs and quadratic-blowup entity expansion attacks still work against ET. SFI will flag this on first pass.

Add:
```
defusedxml>=0.7.1
```

Then update client.py:
```python
from defusedxml import ElementTree as ET
```

This is the standard hardening for any Python service parsing third-party XML.

solutions/ess-maker-skills/src/mcp/workday/client.py
```diff
@@ -0,0 +1,593 @@
...
+                                raise Exception(
+                                    f"Workday SOAP fault ({resp.status_code}): {faultstring}\n{detail_text}"
+                                )
+                        except ET.ParseError:
+                            pass
+                        raise Exception(
+                            f"Workday API error ({resp.status_code}): {resp.text[:500]}"
+                        )
+
+                    root = ET.fromstring(resp.text)
```
johnguy0: **CRITICAL (paired with the requirements.txt comment).**

This is the success-path SOAP response parse. Combined with line 207 (SOAP fault parse), every Workday response flows through `xml.etree.ElementTree` without entity-expansion protection. Switch to `defusedxml.ElementTree` once it's added to requirements.txt.

solutions/ess-maker-skills/src/mcp/workday/client.py
```diff
@@ -0,0 +1,593 @@
...
+                    if resp.status_code >= 400:
+                        # Try to extract SOAP fault
+                        try:
+                            root = ET.fromstring(resp.text)
+                            fault = root.find(f".//{{{SOAP_NS}}}Fault")
+                            if fault is not None:
+                                faultstring = fault.findtext("faultstring", "Unknown SOAP fault")
+                                detail = fault.find("detail")
+                                detail_text = ET.tostring(detail, encoding="unicode") if detail is not None else ""
+                                raise Exception(
```
johnguy0, line 251: **HIGH — SOAP fault detail leaks customer PII into LLM context.**

The full `<detail>` element is serialized to unicode and folded into the exception message. Workday fault details include employee IDs, names, validation errors that reference PII. This exception propagates back to the MCP tool response and into the LLM conversation context (and possibly into LLM provider telemetry).

Fix: extract only safe metadata and discard `detail`.

```python
if fault is not None:
    faultcode = fault.findtext("faultcode", "unknown")
    raise Exception(f"Workday SOAP fault ({resp.status_code}): {faultcode}")
```

If you need the full fault for debugging, log it locally to a file (with permissions per the auth.py pattern in PR #3) but never include it in the exception that flows to the LLM.

solutions/ess-maker-skills/src/mcp/workday/client.py
```diff
@@ -0,0 +1,593 @@
...
+                                    f"Workday SOAP fault ({resp.status_code}): {faultstring}\n{detail_text}"
+                                )
+                        except ET.ParseError:
+                            pass
+                        raise Exception(
+                            f"Workday API error ({resp.status_code}): {resp.text[:500]}"
```
johnguy0: **HIGH — HTTP error body leaks customer data to the LLM.**

`resp.text[:500]` is folded into the exception. Same shape as line 213 — error bodies can include PII and end up in the LLM's context. The 500-char limit is not a security control; it just truncates.

Same issue at `client.py:281` for the RaaS path.

Fix on both lines:
```python
raise Exception(f"Workday error ({resp.status_code})")
```

If you want richer debugging, write the body to a local log file with restrictive permissions and reference the log path in the exception.

solutions/ess-maker-skills/src/mcp/workday/client.py
```diff
@@ -0,0 +1,593 @@
...
+from typing import Optional
+from xml.etree import ElementTree as ET
+from xml.sax.saxutils import escape as xml_escape
+
+import httpx
+from dotenv import load_dotenv
+
+load_dotenv()
+
+logger = logging.getLogger("workday-mcp")
```
johnguy0, line 27: **HIGH — silence httpx loggers to prevent credential exposure at DEBUG.**

httpx logs full HTTP requests at DEBUG. The SOAP envelope contains a WS-Security UsernameToken with username + password. If a customer or operator enables Python DEBUG logging globally (common when troubleshooting), credentials leak to wherever logs are routed.

Add in `WorkdayClient.__init__` (or at module scope):
```python
logging.getLogger("httpx").setLevel(logging.WARNING)
logging.getLogger("httpcore").setLevel(logging.WARNING)
```

And document in the eventual README that this MCP server must never be run with DEBUG-level logging enabled.

solutions/ess-maker-skills/src/mcp/workday/client.py
```diff
@@ -0,0 +1,593 @@
...
+            raise ValueError("WORKDAY_BASE_URL environment variable is required")
+        if not self.tenant:
+            raise ValueError("WORKDAY_TENANT environment variable is required")
+        if not username or not password:
+            raise ValueError(
+                "WORKDAY_USERNAME and WORKDAY_PASSWORD are required in env"
+            )
+
+        self._username = username
+        self._password = password
```
johnguy0, line 96: **MEDIUM — passwords stored as plain instance attributes.**

`self._password = password` (this line) and `self._raas_password` (lines 90, 94) are plain str attributes. Anything that calls `repr(client)` or chains the client into an exception traceback exposes the password.

Fix: don't store the plaintext password as an instance attribute beyond what's strictly required. The `httpx.BasicAuth` object on line 79 already encapsulates the credential; for the SOAP envelope path that needs the raw password (line 154), keep it in a closure or scrub it via `__repr__`:

```python
def __repr__(self):
    return f"<WorkdayClient tenant={self.tenant} base_url={self.base_url}>"
```

Defense-in-depth, but free.

solutions/ess-maker-skills/src/mcp/workday/client.py
```diff
@@ -0,0 +1,593 @@
...
+        self.entra_resource_url = os.environ.get("WORKDAY_ENTRA_RESOURCE_URL", "")
+
+    def _build_client(self, use_auth: bool = True) -> httpx.AsyncClient:
+        kwargs = {
+            "timeout": self.timeout,
+            "headers": {"Content-Type": "text/xml; charset=utf-8"},
+        }
+        if use_auth:
+            kwargs["auth"] = self._auth
+        return httpx.AsyncClient(**kwargs)
```
johnguy0, line 132: **HIGH — set `follow_redirects=False` on every httpx client.**

httpx 0.27.0+ defaults to following redirects. If a Workday endpoint returns 302 (misconfiguration, captive portal, or MITM), httpx follows the redirect and replays the Authorization header to the new target. Workday SOAP/RaaS endpoints don't legitimately redirect — any redirect is an error.

Fix here and at `client.py:264`:
```python
return httpx.AsyncClient(**kwargs, follow_redirects=False)
```

The redirect would surface as a 302 status the caller can handle explicitly, rather than silently leaking the Authorization header.

solutions/ess-maker-skills/src/mcp/workday/client.py
```diff
@@ -0,0 +1,593 @@
...
+        WORKDAY_PASSWORD       — ISU account password
+
+    Optional env vars (for OAuth 2.0 user-context operations):
+        WORKDAY_OAUTH_TOKEN_URL — OAuth token endpoint
+        WORKDAY_OAUTH_CLIENT_ID — OAuth client ID (base64-encoded)
+        WORKDAY_ENTRA_RESOURCE_URL — Entra resource URL for the Workday app
+    """
+
+    def __init__(self):
+        self.base_url = os.environ.get("WORKDAY_BASE_URL", "").rstrip("/")
```
johnguy0, line 76: **MEDIUM — enforce HTTPS on `WORKDAY_BASE_URL`.**

The URL is read from env, stripped of trailing slash, and used directly. A misconfigured `http://` URL sends ISU credentials in cleartext.

Add validation immediately after this line:
```python
if not self.base_url.lower().startswith("https://"):
    raise ValueError("WORKDAY_BASE_URL must use HTTPS")
```

Workday production tenants are always HTTPS, but explicit enforcement protects dev/test misconfigurations.

solutions/ess-maker-skills/src/mcp/workday/client.py
```diff
@@ -0,0 +1,593 @@
...
+            headers["SOAPAction"] = soap_action
+
+        last_error = None
+        for attempt in range(self.max_retries):
+            async with self._build_client(use_auth=False) as client:
+                try:
+                    resp = await client.post(url, content=envelope, headers=headers)
+
+                    if resp.status_code == 429:
+                        wait = int(resp.headers.get("Retry-After", str(2 ** attempt)))
```
johnguy0, line 224: **LOW — add jitter to retry backoff.**

`2 ** attempt` is deterministic. If two parallel MCP tool calls hit a 429 simultaneously, they retry in lockstep. Same pattern at lines 201, 275, 276, 289.

Fix:
```python
import random
await asyncio.sleep((2 ** attempt) + random.uniform(0, 1))
```

Keep the `Retry-After` header path on line 196 unchanged when present.

solutions/ess-maker-skills/src/mcp/workday/server.py
```diff
@@ -0,0 +1,473 @@
...
+    """
+    client = get_client()
+    result = await client.get_absence_balance(
+        employee_id=employee_id,
+        as_of_date=as_of_date,
+    )
+    return _xml_to_json(result)
+
+
+@mcp.tool()
```
johnguy0, line 210: **HIGH — write operation with no user confirmation.**

`request_time_off` submits to Workday the moment the LLM calls it. An LLM hallucination, prompt injection, or misunderstood intent can submit incorrect time-off requests (wrong dates, wrong employee, wrong hours) that require manual reversal in Workday.

MCP doesn't have a built-in confirmation pattern, so handle it at the tool design layer. Two options:

1. **Two-tool flow:**
   - `preview_time_off_request(...)` returns a structured summary, no write.
   - `submit_time_off_request(idempotency_key, ...)` writes, requires the key from the preview response.

2. **Strong tool description.** Add to the docstring:
   > ⚠️ This tool submits a time-off request directly to Workday. The orchestrating agent MUST display the request details to the user and obtain explicit confirmation before invoking this tool.

Option 1 is enforceable; Option 2 relies on the LLM following instructions. Prefer Option 1.

solutions/ess-maker-skills/src/mcp/workday/server.py
```diff
@@ -0,0 +1,473 @@
...
+
+# ═══════════════════════════════════════════════════════════════
+#  RAW SOAP API
+# ═══════════════════════════════════════════════════════════════
+
+
+@mcp.tool()
+async def call_soap_api(
+    service_name: str,
+    body_xml: str,
```
johnguy0, line 367: **HIGH — `call_soap_api` is an LLM-controlled escape hatch to any Workday service.**

The tool docstring lists `Human_Resources, Absence_Management, Compensation, Staffing, Payroll, Benefits, Talent_Management, Performance_Management` as common services, but doesn't restrict the call. A prompt-injected or hallucinating LLM can invoke `Terminate_Employee`, `Change_Compensation`, `Cancel_Direct_Deposit`, or any other Workday SOAP operation.

Fix: enforce a service-name allowlist scoped to read operations only. Move write operations behind named tools that go through the confirmation flow (per the `request_time_off` comment).

```python
READ_ONLY_SERVICES = {"Human_Resources", "Absence_Management", "Compensation", ...}
if service_name not in READ_ONLY_SERVICES:
    raise ValueError(f"call_soap_api supports read-only services only. Use a dedicated typed tool for {service_name}.")
```

Document the allowlist in the tool docstring.

solutions/ess-maker-skills/src/mcp/workday/server.py
```diff
@@ -0,0 +1,473 @@
...
+    Args:
+        service_name: Workday web service name
+        body_xml: XML body of the SOAP request
+        extract_paths: JSON string mapping keys to XPath expressions.
+                      Example: {"JobTitle": "//*[local-name()='Position_Title']/text()"}
+        version: API version (default v42.0)
+    """
+    client = get_client()
+    result = await client.raw_soap(service_name, body_xml, version)
+    paths = _parse_json(extract_paths, "extract_paths")
```
johnguy0, line 443: **MEDIUM — `extract_from_xml` parses LLM-provided XPath expressions.**

The XML is from Workday (trusted), but `extract_paths` JSON keys and XPath values come from the LLM. ElementTree's XPath subset is limited (no full XPath 2.0 functions), so the attack surface is small, but a pathological pattern can still cause quadratic-time evaluation.

The kit's template configs only use a narrow set of XPath shapes (per `src/skills/topics/create/SKILL.md` and the template config XMLs). Allowlist that grammar:

```python
import re
SAFE_XPATH = re.compile(r"^[/.\w\[\]'\"=@*: -]+$")
for key, xpath in paths.items():
    if not SAFE_XPATH.match(xpath):
        raise ValueError(f"Unsupported XPath pattern for key {key!r}")
```

Tune the regex to match the patterns you actually need. Reject anything else.

johnguy0 (changes requested):
Solid foundation. The author got most of the hard things right: env-var credentials with no disk persistence, `httpx.BasicAuth` (not manual base64), `xml.sax.saxutils.escape` on credentials in SOAP envelopes, `requirements.txt` with version-pinned deps (which PR #3 didn't have), bounded retries with `Retry-After` parsing on 429s, separate RaaS credential handling. This reads as production-quality.

That said, this is the first thing SFI/MSRC will deeply review, and there are real findings to land first.

**CRITICAL:**

1. **XML parsing uses `xml.etree.ElementTree` instead of `defusedxml`.** ET.fromstring at `client.py:207` (SOAP fault parsing) and `client.py:222` (success-path response parsing) parses untrusted XML from the customer's Workday tenant. ET correctly ignores DOCTYPE so traditional XXE-via-external-entities isn't directly exploitable, but billion-laughs and quadratic-blowup entity expansion attacks still work. SFI will flag this on first pass. Add `defusedxml>=0.7.1` to requirements.txt and replace the import: `from defusedxml import ElementTree as ET`.

**HIGH:**

2. **SOAP fault detail XML leaks customer PII into LLM context.** `client.py:213` raises an exception with the full `<detail>` element serialized to unicode. That detail can include employee IDs, names, validation errors referencing PII. The exception flows into the MCP tool response, which means it lands in the LLM conversation context and potentially into LLM provider telemetry. Sanitize: extract `faultcode` and a generic `faultstring` only, drop `detail`.

3. **HTTP error bodies leak via `resp.text[:500]`** in two places (`client.py:218-219` SOAP, `client.py:280-281` RaaS). Same shape as #2 — error bodies can include customer data and are returned to the LLM. Replace with `f"Workday error ({resp.status_code})"` and write the body to a local log file if needed for debugging.

4. **Logging hygiene allows credential exposure at DEBUG.** httpx logs full requests at DEBUG level, which would include the WS-Security UsernameToken with username + password in the SOAP envelope. The `workday-mcp` logger at `client.py:20` doesn't silence httpx. Add to `__init__`:
```python
logging.getLogger("httpx").setLevel(logging.WARNING)
logging.getLogger("httpcore").setLevel(logging.WARNING)
```
And document in the README that DEBUG logging must never be enabled.

5. **`request_time_off` is a write operation with no user confirmation.** The LLM can submit a time-off request autonomously. MCP doesn't have a built-in confirmation pattern, but the tool description should explicitly warn the LLM, and the kit's prompts (PR #8) should require user confirmation in the conversation. Better: split into `preview_time_off_request` (read-only validation + return summary) and `submit_time_off_request` (the write). The two-step flow lets the LLM show the preview and ask for confirmation.

6. **`call_soap_api` is an LLM-controlled escape hatch to any Workday SOAP endpoint.** The tool docstring lists `Human_Resources, Absence_Management, ...` but doesn't constrain the call. A prompt-injected LLM can invoke `Terminate_Employee`, `Change_Compensation`, `Cancel_Direct_Deposit`. Add a `service_name` allowlist of read-only services in this tool, and require a separate, confirmation-gated tool for write operations.

7. **httpx clients don't set `follow_redirects=False`.** `client.py:111` and `client.py:264`. httpx 0.27.0+ defaults to following redirects. If a Workday endpoint returns a 302 (misconfiguration or MITM), httpx will follow it and replay the Authorization header to the redirect target. Workday SOAP/RaaS endpoints don't legitimately redirect. Set `follow_redirects=False` on every `AsyncClient`.

**MEDIUM:**

8. **No HTTPS enforcement on `WORKDAY_BASE_URL`.** `client.py:63` accepts whatever the user puts in the env var. A misconfigured `http://` URL sends ISU credentials in cleartext. Validate in `__init__`: `if not self.base_url.lower().startswith("https://"): raise ValueError(...)`.

9. **`extract_from_xml` parses LLM-provided XPath expressions.** `server.py:351-379`. The XML is from Workday (trusted), but the XPath comes from the LLM. ElementTree's XPath subset is limited so the attack surface is small, but a pathological pattern can still cause quadratic-time evaluation. Allowlist a narrow XPath grammar (the kit's template configs only use a few patterns).

10. **Credentials stored as plain instance attributes** (`client.py:78` `self._password`, `client.py:90/94` `self._raas_password`). Anything that calls `repr(client)` or includes the client in an exception chain leaks the password in the trace. Use a `__repr__` override that scrubs them, or store in a wrapper that refuses serialization.

11. **Retry backoff has no jitter.** `client.py:196, 201, 275, 276, 289` use `2 ** attempt` exactly. If two parallel tool calls retry simultaneously, they thunder. Add `+ random.uniform(0, 1)`.

**PR-level:**

12. PR description still says `solutions/ess-agent-kit/src/mcp/workday/`. Actual path is `solutions/ess-maker-skills/src/mcp/workday/`. Same stale-rename issue from PRs #2 and #3.

13. README/setup docs in this PR don't exist — this MCP server is bound to a `.vscode/mcp.json` config that the customer authors. Confirm PR #6 (Connect skill) covers wiring this up safely (input prompts for password via VS Code's `${input:...}` so credentials never land on disk).

Inline below.

saengland:
Addressed all CRITICAL+HIGH+MEDIUM findings from your review in 83af583.

**CRITICAL: Missing defusedxml**
- Added `defusedxml>=0.7.1` to requirements.txt.
- All untrusted-XML parsing in client.py (SOAP responses + fault parsing) now goes through `defusedxml.ElementTree.fromstring`. Stdlib `xml.etree.ElementTree` is kept only for namespace registration and building (not parsing); type hints (`ET.Element`) also use the stdlib type since the parsed objects share the API. Defends against billion-laughs and quadratic-blowup entity expansion.

**HIGH: HTTPS enforcement**
- `WORKDAY_BASE_URL` now must start with `https://` or `WorkdayClient()` raises `ValueError`. Refuses to send ISU credentials over cleartext.

**HIGH: PII leakage via SOAP fault detail / HTTP body**
- SOAP fault exceptions now include only the `faultstring` and HTTP status code. The full `<detail>` element (which can contain employee IDs, names, validation messages) is no longer concatenated into the exception message, so it cannot flow into the LLM context.
- Generic non-SOAP error bodies likewise stripped from the exception.
- Full body still logged at DEBUG for the operator (local-only).

**HIGH: httpx DEBUG credential exposure**
- `httpx` and `httpcore` loggers explicitly silenced to WARNING. If a downstream operator turns on global DEBUG logging they will not get an Authorization-header-leaking httpx trace.

**HIGH: follow_redirects=True default**
- `_build_client()` now sets `follow_redirects=False`. A 302 from Workday cannot replay the Authorization header to an attacker-controlled host.

**HIGH: Plain password attribute exposed by repr/tracebacks**
- `__repr__` now returns only base_url, tenant, and username. The `_password` attr is never echoed in repr or default exception traces.

**HIGH: No jitter in retry backoff**
- 429 retry sleep + RequestError retry sleep both now add `random.uniform(0, 1)`. Avoids thundering-herd on rate-limit recovery.

**HIGH: Unrestricted call_soap_api**
- Added `_READONLY_SOAP_SERVICES` allowlist (Human_Resources, Absence_Management, Compensation, Staffing, Talent, Talent_Management, Performance_Management, Benefits_Administration, Payroll, Integrations, Recruiting). `call_soap_api` and `extract_from_xml` reject any service_name outside the set with a JSON error pointing the LLM at the typed tools or at adding the service with team review. Prompt-injection cannot drive `Terminate_Employee`, `Change_Compensation`, `Cancel_Direct_Deposit`, etc. through the raw-SOAP escape hatch.

**HIGH: Write tool with no preview/confirm (request_time_off)**
- `request_time_off` now requires `confirm=True`. Without it the tool returns a structured JSON preview with status="preview" instructing the LLM to show the request to the user and re-call only after explicit user OK. Mutation only happens on the confirm-true path. Same shape can be applied to future write tools.

**Already addressed in beaee39 (verified on this branch):**
- SOAP XML escaping for credentials (`xml.sax.saxutils.escape`)
- `rstrip('/text()')` -> `removesuffix('/text()')`
- `_parse_json` helper for JSON tool args

**Repo hygiene:**
- Added `.gitignore` for `__pycache__/` and `*.pyc` in workday/.

Ready for re-review.

saengland:
This PR was prematurely merged earlier today during a tooling incident. `main` and `main2` have been rolled back; replacement PR with the same branch and code is now open as #30 for proper review.

PR #30, opened 2026-05-04
PR #4: Workday MCP server

saengland:
> **Note:** This PR replaces the original #14, which was prematurely merged earlier today during a tooling incident and rolled back.
> The branch (`review/04-mcp-workday`) carries commit `83af5836` that addresses the entire PR #14 review.

---

Adds the Workday MCP server under `solutions/ess-maker-skills/src/mcp/workday/`. Local stdio MCP server that proxies SOAP and RaaS API calls to the customer's Workday tenant.

10 tools: connectivity test, HR data (`get_worker`, `list_workers`), absence (`get_time_off_balance`, `request_time_off`), orgs, RaaS reports (`run_report`, `get_user_context`), raw SOAP escape hatches (`call_soap_api`, `extract_from_xml`).

Auth: WS-Security UsernameToken for SOAP, HTTP Basic for RaaS. Credentials env-only via `${input:...}` prompts; never written to disk.

**All 11 PR #14 findings addressed (commit `83af5836`):**

1. CRITICAL — `defusedxml.ElementTree` for parsing untrusted XML (stdlib ET kept for *building* envelopes)
2. HIGH — SOAP fault PII strip: only `faultstring` + status code in exceptions; full body at `logger.debug` only
3. HIGH — HTTP error body strip with same treatment
4. HIGH — `httpx`/`httpcore` loggers silenced at module scope (kills DEBUG-level credential leak)
5. HIGH — `request_time_off` two-step confirm via `confirm: bool = False`
6. HIGH — `call_soap_api` restricted to 11 read-only services with helpful error + allowlist
7. HIGH — `follow_redirects=False`
8. MEDIUM — HTTPS enforcement on `WORKDAY_BASE_URL`
9. MEDIUM — `extract_from_xml` allowlist
10. MEDIUM — credentials scrubbed from `__repr__`
11. LOW — retry jitter on backoff paths

**SFI focus areas:**
- WS-Security UsernameToken handling for SOAP (no HTTP Basic credentials over the wire for SOAP)
- HTTP Basic for RaaS (RFC-required for that surface; HTTPS enforced)
- defusedxml entity expansion protection on parse
- Error message scrubbing (no PII in exceptions / standard logs)
- Network egress (only customer-provided Workday endpoint)
- Read-only allowlist for raw SOAP escape hatches

This commit (`590c79d`) drops two unused `import base64` + `import json` for ruff F401 cleanup.

Tracker: #10

PR #29, opened 2026-05-04
PR #3: Python scripts

saengland:
> **Note:** This PR replaces the original #13, which was prematurely merged earlier today during a tooling incident and rolled back.
> The branch (`review/03-scripts`) is unchanged from the rolled-back PR, plus follow-up commits addressing the PR #13 review.

---

Adds the Python tooling under `solutions/ess-maker-skills/scripts/`:
- `auth.py` — MSAL interactive auth + Dataverse REST helpers
- `setup.py` — one-shot `/setup` extraction (components, template configs, workflows, evaluations) → local files + `.baseline/` + `.component-map.json`
- `push.py` — diff working files vs `.baseline/`, push CRUD to Dataverse
- `extract.py` — bulk extraction helper
- `flightcheck/` — pre-deployment readiness validation (Graph + PP Admin clients, 41+ checks)

**Round 2 review applied (commit `e5dd667a`):**

All 4 CRITICAL findings from PR #13 fixed:
1. `push.py` partial-failure baseline correctly gated (`errors == 0 and success > 0`)
2. component_map saved in same gate
3. `auth.py` token cache uses `os.open` with explicit `0o600` mode
4. `auth.py` cache directory chmodded to `0o700`

Most HIGH findings fixed:
- `--force-delete` flag with interactive `type 'delete' to confirm`
- 401 token-refresh wrapper (`_call_with_refresh`)
- HTTPS validation, `verify=True` explicit, OData URL-encoding, atomic config write, error-message scrubbing

**Round 3 review applied (this commit):**

- Dropped half-built etag parameter from `update_record` (will re-add when push.py captures `@odata.etag`)
- Added `requirements.txt` with pinned msal/requests
- Idempotency: setup.py now refuses an existing agent dir without `--refresh`
- `configVersion: 1` field added to `.local/config.json` for future schema migration
- 429/5xx retry-with-backoff via `urllib3.util.retry.Retry`
- CLIENT_ID documented with source link to learn.microsoft.com

Tracker: #10

PR #31, opened 2026-05-04
PR #5: ServiceNow MCP server

saengland:
> **Note:** This PR replaces the original #15, which was prematurely merged earlier today during a tooling incident and rolled back.

---

Adds the ServiceNow MCP server under `solutions/ess-maker-skills/src/mcp/servicenow/`. Local stdio MCP server that proxies REST API calls to the customer's ServiceNow tenant.

~25 tools across ITSM (incidents), HRSD (cases, knowledge), CMDB (configuration items), Service Catalog, User resolution, Live Agent (interactions), and OAuth/OIDC integration setup. Plus generic CRUD escape hatches (`query_table`, `create_record`, `update_record`, `delete_record`, `call_api`).

**Auth: HTTP Basic Auth over HTTPS** (not OAuth2 — see SFI section below; the original PR description said OAuth2 in error). HTTPS enforcement, table denylists at both client and server layers, sys_id and table-name regex validation, encoded-query injection prevention via `_q()`.

**SFI focus areas:**
- HTTP Basic over HTTPS (HTTPS enforced; no Basic over HTTP allowed)
- REST request construction with sys_id + table-name regex validation
- Network egress (only customer-provided ServiceNow endpoint)
- Encoded-query injection guard via `_q()` on all interpolated values
- Response/error handling (no leakage of customer ticket data into LLM context)
- Table denylists at both layers (client + server) for `sys_user_grmember`, `sys_user_has_role`, `sys_oidc_provider`, etc.
- LLM-supplied `additional_fields` cannot clobber tool-controlled fields

**Round 1 review applied (this commit):**

Architect:
- Promoted `client._request` → `client.request` (public; `call_api` now goes through public surface)
- Consolidated table denylist into a single `EXCLUDED_TABLES` constant; client + server import from one source

Security:
- All four admin/lookup tool paths now route values through `_q()` (`query_table:641`, `list_oidc_providers:721`, `check_plugin_active:808`, `get_system_properties:824`)
- `data.update(extra)` reversed to `data = {**extra, **data}` so LLM-supplied `additional_fields` cannot overwrite tool-controlled `u_interaction` / `u_summary`
- `logger.debug` body emission dropped (the comment promised PII scrub; the debug line undermined it)

Performance:
- Long-lived `httpx.AsyncClient` cached on `self`, lazy-init, closed on shutdown
- Retry path extended to 502/503/504 with same backoff as 429
- `Retry-After` parser falls back to exponential backoff if the value is an HTTP-date (RFC 9110)

Prompt-engineering:
- Server-level instructions add: "prefer typed tools first; CRUD/escape hatches only when no typed tool fits"
- Routing line added to `query_table`, `create_record`, `update_record`, `delete_record` docstrings
- `assigned_to` doc fixed: only sys_id matches; if you have a name, call `resolve_user` first
- `additional_fields` doc lists supported `u_*` columns with explicit failure-mode warning

Tracker: #10


saengland:
@johnguy0 — heads up: the remaining red X on Lint Python here is the cross-branch E902 (`No such file or directory`) from `ruff` being asked to lint both `solutions/ess-maker-skills/scripts/` AND `solutions/ess-maker-skills/src/mcp/` while this branch only carries `src/mcp/servicenow/`. The actual code-level violation (E402 late `import re as _re`) is now fixed. Self-resolves on merge to `main2` once #29 + #30 land too. Ready for your review.

johnguy0 (commented):
**[architect]**

Shape is sound: client/server split, lazy singleton, env-driven auth, tool surface organized by domain. Two boundary issues to fix before this lands as the template for the next MCP server.

1. `call_api` reaches into the client's private `_request`, see line 925.
2. The two table denylists (`_DENYLIST_TABLES` in client.py, `_CALL_API_TABLE_DENYLIST_NON_GET` in server.py) overlap but already diverge, see line 144.

Both are easy fixes that pay off the next time someone adds a tool.

solutions/ess-maker-skills/src/mcp/servicenow/server.py
```diff
@@ -0,0 +1,934 @@
...
+                f"call_api: {method_upper} on table '{m.group(1)}' is forbidden. "
+                "This table holds security/auth/audit configuration. Use a "
+                "typed tool or the /connect skill for legitimate changes."
+            )
+
+    client = get_client()
+    kwargs = {}
+    if data and method_upper in ("POST", "PATCH"):
+        kwargs["json"] = _parse_json(data, "data")
+    result = await client._request(method_upper, path, **kwargs)
```
johnguy0: `call_api` goes through `client._request`, which is private. The typed methods all go through public methods on the client (`query_table`, `create_record`, ...). Either promote `_request` to `request` (or add a thin `client.call(method, path, json=...)`), or route `call_api` through a typed entrypoint. Right now any change to `_request`'s shape silently breaks `call_api`.

solutions/ess-maker-skills/src/mcp/servicenow/server.py
```diff
@@ -0,0 +1,934 @@
...
+    "sys_security_diag",
+    "oauth_entity",
+    "oauth_entity_profile",
+    "sys_oidc_provider",
+    "sys_certificate",
+    "sys_script",
+    "sys_script_include",
+    "sys_script_action",
+    "sys_script_client",
+    "sys_ws_operation",
```
johnguy0: This denylist already differs from `_DENYLIST_TABLES` in client.py. `sys_user_grmember`, `sys_user_has_role`, `sys_oidc_provider`, `sys_security_diag`, `sys_script*`, `sys_ws_operation` are only here. `sys_log*`, `sys_log_transaction`, `oauth_entity_profile`, `sys_certificate` are only in client. Two parallel lists will drift. Move the canonical list to one module and import it in both, or add a comment that says exactly why server-side is stricter.

johnguy0 (commented):
**[security]**

Real injection gap on four admin/lookup tools that bypass `_q()` (lines 641, 721, 808, 824). The `additional_fields` merge at line 621 lets prompt injection clobber the trusted record reference. The auth implementation in code (`BasicAuth`, client.py line 94) doesn't match what the PR description says is being reviewed (`OAuth2 token handling (refresh, expiry, storage)`).

Other security calls in the diff are good: HTTPS enforcement, `follow_redirects=False`, env-name-only secret passing for admin tools, table denylists at both layers, sys_id and table regex validation.

solutions/ess-maker-skills/src/mcp/servicenow/server.py
```diff
@@ -0,0 +1,934 @@
...
+
+@mcp.tool()
+async def list_oauth_applications(query: str = "", limit: int = 10) -> str:
+    """List OAuth Application Registry entries (oauth_entity table).
+
+    Args:
+        query: Free text search in application name
+        limit: Max results
+    """
+    encoded_query = f"nameLIKE{query}" if query else ""
```
johnguy0: `query` is interpolated raw into the encoded query string here. A value like `^OR1=1` recreates the encoded-query injection that `_q()` exists to prevent. Same omission at line 721 (`list_oidc_providers`), line 808 (`check_plugin_active` interpolating `plugin_id` into `source=...`), and line 824 (`get_system_properties`). Run every interpolated value through `_q()`.

solutions/ess-maker-skills/src/mcp/servicenow/server.py
```diff
@@ -0,0 +1,934 @@
...
+    """
+    data: dict = {
+        "u_interaction": interaction_id,
+        "u_summary": summary,
+    }
+    if resolution:
+        data["u_resolution"] = resolution
+
+    extra = _parse_json(additional_fields, "additional_fields") if additional_fields != "{}" else {}
+    data.update(extra)
```
johnguy0: `data.update(extra)` lets LLM-supplied `additional_fields` overwrite `u_interaction` and `u_summary`. A prompt-injection payload like `{"u_interaction": "<other_sys_id>"}` writes the summary against the wrong record. Either pop those keys from `extra` first, or reverse the merge order: `data = {**extra, **data}`.

solutions/ess-maker-skills/src/mcp/servicenow/client.py
```diff
@@ -0,0 +1,242 @@
...
+            raise ValueError(
+                "SERVICENOW_USERNAME and SERVICENOW_PASSWORD are required in env"
+            )
+
+        self.base_url = instance_url.rstrip("/")
+        self._username = username
+        self._password = password
+        self.max_retries = 3
+        self.timeout = 30.0
+        self._auth = httpx.BasicAuth(username, password)
```
johnguy0, line 143: PR description lists `OAuth2 token handling (refresh, expiry, storage)` as an SFI focus area. Code uses `httpx.BasicAuth(username, password)`. No tokens, no refresh, no expiry. Either the SFI write-up needs updating to say `HTTP Basic over HTTPS`, or OAuth2 is expected and missing from this PR. Reconcile so reviewers know what surface to evaluate.

solutions/ess-maker-skills/src/mcp/servicenow/client.py
```diff
@@ -0,0 +1,242 @@
...
+                        body = e.response.json()
+                        error_msg = body.get("error", {}).get("message", "")
+                    except Exception:
+                        error_msg = ""
+                    if not error_msg:
+                        error_msg = f"HTTP {e.response.status_code}"
+                    logger.debug(
+                        "ServiceNow HTTP error body (operator-only): %s",
+                        getattr(e.response, "text", ""),
+                    )
```
johnguy0: The comment above promises the response body won't surface (PII, internal field metadata), and the raised exception honors that. This `logger.debug` line still emits the full body. If a downstream operator turns on global DEBUG (the same scenario the httpx logger guards above explicitly defend against), the body lands in logs. Either drop this line or write to a dedicated logger that operators have to opt into by name.

solutions/ess-maker-skills/src/mcp/servicenow/client.py
```diff
@@ -0,0 +1,242 @@
...
+            # Disable redirect-following so a 302 cannot replay the
+            # Authorization header to an attacker-controlled host.
+            follow_redirects=False,
+        )
+
+    async def _request(self, method: str, path: str, **kwargs) -> dict:
+        """Execute an HTTP request with retry on 429."""
+        last_error = None
+        for attempt in range(self.max_retries):
+            async with self._build_client() as client:
```
johnguy0: Every `_request` call builds and tears down a fresh `httpx.AsyncClient` (defined at line 99-111, used here). No connection reuse, no keep-alive, fresh TLS handshake per request. An LLM session usually fires 5-50 calls. Hold one `AsyncClient` on `self`, init lazily on first request, close on shutdown. Saves 50-200ms per call after the first.

solutions/ess-maker-skills/src/mcp/servicenow/client.py
```diff
@@ -0,0 +1,242 @@
...
+                        error_msg = ""
+                    if not error_msg:
+                        error_msg = f"HTTP {e.response.status_code}"
+                    logger.debug(
+                        "ServiceNow HTTP error body (operator-only): %s",
+                        getattr(e.response, "text", ""),
+                    )
+                    raise Exception(
+                        f"ServiceNow API error ({e.response.status_code}): {error_msg}"
+                    ) from e
```
johnguy0: Retries fire on 429 and on `RequestError` (connection-level). `HTTPStatusError` raises immediately, so a 502/503/504 from ServiceNow during a deploy or restart fails on the first try. Add 5xx (or at least 502/503/504) to the retry path with the same backoff used for 429.

solutions/ess-maker-skills/src/mcp/servicenow/client.py
```diff
@@ -0,0 +1,242 @@
...
+        """Execute an HTTP request with retry on 429."""
+        last_error = None
+        for attempt in range(self.max_retries):
+            async with self._build_client() as client:
+                try:
+                    resp = await client.request(method, path, **kwargs)
+
+                    if resp.status_code == 429:
+                        retry_after = resp.headers.get("Retry-After")
+                        wait = int(retry_after) if retry_after else 2 ** attempt
```
johnguy0: `Retry-After` is allowed to be an HTTP-date by RFC 9110, not just an integer. `int(retry_after)` raises on a date value, which short-circuits the 429 retry into an unhandled exception. Wrap in try/except and fall back to `2 ** attempt` on parse failure.

johnguy0 (commented):
**[performance]**

Two changes worth making before this sees production load.

1. Long-lived `httpx.AsyncClient` instead of one-per-request, see client.py line 117.
2. Retry on transient 5xx, not just 429, see client.py line 145.

Cost shape is otherwise fine for the call volumes MCP tools generate.

solutions/ess-maker-skills/src/mcp/servicenow/server.py
```diff
@@ -0,0 +1,934 @@
...
+    category: str = "",
+    limit: int = 10,
+) -> str:
+    """Search IT incidents with common filters.
+
+    Args:
+        query: Free text search in short_description
+        state: Filter by state (1=New, 2=In Progress, 3=On Hold, 6=Resolved, 7=Closed)
+        priority: Filter by priority (1=Critical, 2=High, 3=Moderate, 4=Low, 5=Planning)
+        assigned_to: Filter by assigned user display name or sys_id
```
johnguy0: `by assigned user display name or sys_id` is wrong. ServiceNow `assigned_to=<value>` matches sys_id only; passing a display name silently returns zero results. The LLM will pass display names because the doc invites it. Two options: (a) restrict the doc to sys_id and tell the LLM `call resolve_user first if you only have a name`; (b) detect non-sys_id input in code and switch the operator to `assigned_to.nameLIKE`. Either way fix the docstring so it doesn't promise something the API doesn't deliver.

solutions/ess-maker-skills/src/mcp/servicenow/server.py
```diff
@@ -0,0 +1,934 @@
...
+) -> str:
+    """Query any ServiceNow table using encoded query syntax.
+
+    Args:
+        table: Table name (e.g., incident, sys_user, cmdb_ci)
+        query: ServiceNow encoded query (e.g., 'state=1^priority=2')
+        fields: Comma-separated field names to return (empty = all)
+        limit: Max records to return (default 10, max 1000)
+        offset: Starting record offset for pagination
+        order_by: Field to sort by (prefix with DESC for descending)
```
johnguy0, line 187: Nothing in this docstring tells the LLM when to reach for `query_table` vs `search_incidents`, `resolve_user`, `search_cmdb_items`, etc. Without that, the LLM will use `query_table` whenever it knows a table name and skip the typed tools, and you lose all the field-list and limit defaults the typed tools encode. Add a line at the top of the docstring:

```
Prefer typed tools (search_incidents, resolve_user, search_cmdb_items, browse_service_catalog, search_hr_cases) when one fits. Use query_table only for tables without a typed tool.
```

Same guidance belongs on `create_record` (line 217), `update_record` (line 231), `delete_record` (line 246), and `call_api` already says `EXPLORATION ONLY` which is the right shape.

solutions/ess-maker-skills/src/mcp/servicenow/server.py
```diff
@@ -0,0 +1,934 @@
...
+from mcp.server.fastmcp import FastMCP
+
+from client import ServiceNowClient
+
+mcp = FastMCP(
+    "servicenow",
+    instructions=(
+        "ServiceNow REST API integration for ITSM incidents, HR cases, "
+        "CMDB, service catalog, user lookup, live agent, and OAuth/OIDC setup"
+    ),
```
johnguy0, line 33: Server instructions describe what the server connects to but don't steer routing. Add a sentence:

```
Always prefer the typed tool that names the resource (incidents, HR cases, users, CMDB, catalog). query_table, create_record, update_record, delete_record, and call_api are escape hatches for unsupported scenarios.
```

Server-level instructions reach the LLM before any individual tool docstring, so this is the highest-leverage place to put routing guidance.

solutions/ess-maker-skills/src/mcp/servicenow/server.py
```diff
@@ -0,0 +1,934 @@
...
+    resolution: str = "",
+    additional_fields: str = "{}",
+) -> str:
+    """Log a copilot chat summary to the u_ess_copilot_summary custom table.
+
+    Args:
+        interaction_id: sys_id of the related interaction record
+        summary: Summary of the copilot conversation
+        resolution: Resolution or outcome description
+        additional_fields: JSON string of extra field/value pairs
```
johnguy0: `additional_fields: JSON string of extra field/value pairs` doesn't tell the LLM what fields are valid on `u_ess_copilot_summary`, what the failure mode is for unknown fields, or that it must be a JSON object (not array). Either list the supported `u_*` columns inline, or drop `additional_fields` from this tool and add a typed param when a field becomes necessary. As written, the LLM will guess column names and surface ServiceNow validation errors back to the user.

johnguy0 (commented):
**[prompt-engineer]**

Per-tool docstrings are clear individually. The gap is routing across the ~25 tools. Nothing tells the LLM to prefer typed tools over the generic CRUD/escape hatches, and `search_incidents`'s `assigned_to` doc (line 279) will produce silent zero-result calls.

Three small text changes close the gap:
1. Fix the `assigned_to` doc.
2. Add a routing line to `query_table`, `create_record`, `update_record`, `delete_record`, `call_api`.
3. Tighten the server-level instructions to say `prefer typed tools first`.

saengland:
@johnguy0 — round-1 fixes pushed (commit `48606e6`, plus a cleanup commit `1a0d1fd` to drop accidentally-committed `__pycache__` files). All 13 inline findings + 4 review-summary items addressed:

**Architect**
- ✅ `client._request` promoted to public `client.request`; `_request` kept as backward-compat alias. `call_api` in server.py now routes through the public surface.
- ✅ Table denylist consolidated into `client.EXCLUDED_TABLES`; server imports from there. Old parallel lists were already drifting (5 entries server-only, 4 entries client-only). Single source now.

**Security (real defects, all fixed)**
- ✅ Encoded-query injection at `list_oauth_applications:641`, `list_oidc_providers:721`, `check_plugin_active:808`, `get_system_properties:824` — all four now route values through `_q()`.
- ✅ `log_copilot_summary:621` — reversed merge to `data = {**extra, **data}` so LLM-supplied `additional_fields` cannot clobber `u_interaction` / `u_summary` / `u_resolution`. Added `isinstance(extra, dict)` guard.
- ✅ Description / code mismatch on auth: PR description now correctly says HTTP Basic over HTTPS (was wrongly saying OAuth2). SFI focus areas refreshed.
- ✅ `client.py:159` `logger.debug` body emission removed. The comment promised PII scrub; the debug line undermined it.

**Performance**
- ✅ Long-lived `httpx.AsyncClient` cached on `self` via `_ensure_client`; `aclose()` for shutdown. Saves the per-call TLS handshake.
- ✅ Retry path extended to 502/503/504 with same backoff as 429.
- ✅ `_parse_retry_after` handles RFC 9110 HTTP-date format with fallback to delta-seconds with fallback to `2**attempt` on parse failure.

**Prompt-engineering**
- ✅ Server-level FastMCP `instructions` add the "prefer typed tools first" routing line.
- ✅ Routing line added to docstrings on `query_table`, `create_record`, `update_record`, `delete_record`.
- ✅ `assigned_to` doc fixed: sys_id only; tells LLM to call `resolve_user` first if it only has a name.
- ✅ `additional_fields` doc lists supported `u_*` columns + JSON-object requirement + warning about unknown columns.

PR description also refreshed (correctly describes `solutions/ess-maker-skills/src/mcp/servicenow/`, all fixes documented, SFI focus areas explicit).

Ready for your re-review.

saengland:
@johnguy0 - all four round-1 persona findings are addressed in head commit `1a0d1fd` (per the PR description). Ready for round-2 verify when you have a slot.

Quick recap of what changed since your `eb0495e5` review:

- **architect:** `client._request` -> public `client.request`; `EXCLUDED_TABLES` consolidated, both client and server import from one source.
- **security:** all four admin/lookup tools now route through `_q()` (lines 641, 721, 808, 824); `data.update(extra)` reversed to `data = {**extra, **data}` so `additional_fields` cannot clobber tool-controlled fields; `logger.debug` body emission dropped; PR description corrected from OAuth2 to HTTP Basic over HTTPS.
- **performance:** long-lived `httpx.AsyncClient` cached on `self`, lazy-init, closed on shutdown; retry path extended to 502/503/504; `Retry-After` parser falls back to exponential when given an HTTP-date.
- **prompt-engineer:** server-level instructions add "prefer typed tools first"; routing line added to `query_table` / `create_record` / `update_record` / `delete_record` docstrings; `assigned_to` doc fixed (sys_id only, point at `resolve_user`); `additional_fields` doc lists supported `u_*` columns.

saengland:
**Re: failing `Lint Python` check** - this isn't a code issue on this branch. The CI workflow on `main2` hard-codes `solutions/ess-maker-skills/scripts/` AND `solutions/ess-maker-skills/src/mcp/` as args to ruff, so ruff returns `E902 No such file or directory` on any branch that doesn't have both legs of the kit yet (which is every per-PR review branch).

Workflow fix is in #47 (against `main2`). Once that lands I'll merge `main2` into this branch and re-run CI to confirm green.

saengland:
**Re: failing Lint Python check** - this isn't a code issue on this branch. The CI workflow on `main2` hard-codes `solutions/ess-maker-skills/scripts/` AND `solutions/ess-maker-skills/src/mcp/` as args to ruff, so ruff returns `E902 No such file or directory` on any branch that doesn't have both legs of the kit yet (which is every per-PR review branch).

Workflow fix is in #47 (against `main2`). Once that lands I'll merge `main2` into this branch and re-run CI to confirm green.

johnguy0 (commented):
Round 2 on PR #31 (ServiceNow MCP server). All 13 threads from round 1 are addressed in 1ad036e8.

**Architect**
- ✅ `_request` promoted to public `request`; `_request` kept as a back-compat alias. `call_api` now routes through the public surface (`server.py:950`).
- ✅ Single canonical denylist: `EXCLUDED_TABLES` lives in `client.py` and is imported by `server.py:21` as `_CALL_API_TABLE_DENYLIST_NON_GET`. The two layers can no longer drift.

**Security**
- ✅ Encoded-query injection closed on all four admin/lookup tools: `list_oauth_applications` (664), `list_oidc_providers` (744), `check_plugin_active` (831), `get_system_properties` (847) all run user input through `_q()`.
- ✅ `log_copilot_summary` reverses the merge order (`data = {**extra, **data}`, `server.py:644`) and rejects non-object `additional_fields`. Prompt injection on `u_interaction` / `u_summary` / `u_resolution` is no longer possible.
- ✅ `logger.debug` body emission dropped from `client.py`; comment block explains why so it doesn't sneak back.
- ✅ PR description reconciled to "HTTP Basic over HTTPS"; SFI focus areas now match the code. No OAuth2 token handling claimed.

**Performance**
- ✅ Long-lived `httpx.AsyncClient` cached on `self._client`, lazy-init via `_ensure_client` with proper double-checked locking, closed via `aclose()`. Connection reuse and TLS session reuse across the typical 5-50-call LLM session.
- ✅ Retry path extended to 502/503/504 with the same backoff path as 429 (`client.py:192`).
- ✅ `_parse_retry_after` handles both delta-seconds and HTTP-date formats per RFC 9110, falling back to exponential on parse failure. Won't blow up on a date-valued `Retry-After` anymore.

**Prompt-engineering**
- ✅ Server-level instructions now steer routing first ("Always prefer the typed tool that names the resource... use them last, not first").
- ✅ `query_table`, `create_record`, `update_record`, `delete_record` docstrings each include a "prefer typed tools" line.
- ✅ `assigned_to` docstring is now correct (sys_id only; call `resolve_user` first if you have a name).
- ✅ `additional_fields` docstring enumerates supported `u_*` columns and the failure mode for unknown columns.

Two follow-ups worth filing as separate issues, not blocking on this PR:

1. `call_api` path normalization: the prefix allowlist (`/api/now/table/`, `/api/now/import/`, `/api/now/attachment`, `/api/now/stats/`) accepts the path verbatim. Path traversal segments like `/api/now/table/incident/../sys_user_grmember` would pass the prefix check, miss the `_CALL_API_TABLE_PATH_RE` match, and skip the denylist. Worth adding a normalize-and-reject step before the prefix check. Pre-existing in eb0495e, not introduced by this delta.

2. `/api/now/import/` and `/api/now/stats/` aren't covered by the table denylist regex (it only matches `/api/now/(v[12]/)?table/`). An LLM could `call_api POST /api/now/import/sys_user_grmember` and bypass write protection on a denylisted table. Same pre-existing concern.

Ready for approval.

solutions/ess-maker-skills/src/mcp/servicenow/server.py
```diff
@@ -0,0 +1,959 @@
...
+        data["redirect_url"] = redirect_url
+    if auth_url:
+        data["auth_url"] = auth_url
+    if token_url:
+        data["token_url"] = token_url
+    if comments:
+        data["comments"] = comments
+
+    client = get_client()
+    result = await client.create_record("oauth_entity", data)
```
GrahamMcMynn, line 731: `register_oauth_application` is non-functional once the round-1 denylist consolidation is in effect. `oauth_entity` is in `client.EXCLUDED_TABLES` (client.py line 60), and `client.create_record` calls `_validate_table(table)` with no `allow_denylist=True` (client.py line 314). Any invocation - even with `SERVICENOW_MCP_ENABLE_ADMIN_TOOLS=1` set and the `_require_admin_tools` gate cleared - raises `ValueError: Refusing to operate on protected table ''oauth_entity''` before the API call ever fires.

Same shape kills `set_system_property` at server.py line 874: `sys_properties` is in `EXCLUDED_TABLES` (client.py line 71), `update_record` enforces the denylist with no override, so the tool always throws on entry.

The round-1 consolidation closed the prompt-injection write path on these tables, but inadvertently nailed the door shut on the legitimate admin path too. Two ways to unblock:

1. Thread an `allow_denylist=True` kwarg through `client.create_record` / `client.update_record` and pass it from the gated admin tools (`register_oauth_application`, `set_system_property`, and `register_oidc_provider` defensively) where the `_require_admin_tools` gate already provides the safety story. (`register_oidc_provider` writes `oauth_oidc_entity`, which is not on the denylist, so it currently works - but if `oauth_oidc_entity` ever gets added the same break pattern reappears.)
2. Have the admin tools bypass the typed CRUD layer entirely and call `client.request("POST", "/api/now/table/oauth_entity", json=data)` / `client.request("PATCH", "/api/now/table/sys_properties/{sys_id}", json=...)` directly - the `_require_admin_tools` env flag is the trust boundary in that case.

Either way needs a smoke-test of the `/connect` flow with the admin flag enabled before this lands; right now both admin write paths are dead code.

GrahamMcMynn (commented):

srideshpande (commented):
These are reviewed by John's agent, and per the agent, it is ready for approval. Hence, approving.

srideshpande (approved):

PR #37, opened 2026-05-05
Add CONFIG-013: Knowledge source crawl/index readiness check

### LINKED ISSUE #24: FlightCheck: Add check for knowledge source crawl/index readiness

## What should be checked

The FlightCheck tool should verify that all configured knowledge sources have completed their initial crawl and are fully indexed before the agent goes live. Customers have experienced inconsistent agent responses because knowledge sources were not yet available when the agent was deployed.

## Suggested implementation

- **Checkpoint ID:** CONFIG-013
- **Category:** Local Files
- **Priority:** HIGH
- **Validation logic:** Query the knowledge source status API (or Dataverse entity tracking crawl state) to confirm all configured knowledge sources report a "completed" or "indexed" status. Alert if any source is in "pending," "crawling," or "failed" state.

## Context

GrahamMcMynn:
## Summary

Implements #24  adds a new flightcheck (CONFIG-013) that verifies all configured knowledge sources have completed their initial crawl/index and are ready for deployment.

### PVA Client (scripts/flightcheck/pva_client.py)
- Authenticates to the Copilot Studio Island Gateway API (resource: \96ff4394-9197-43aa-b393-6a41652e21f8\)
- Discovers the gateway URL automatically via BAP environment API
- Fetches \KnowledgeSourceComponent\ objects with live \state\/\status\ fields

### CONFIG-013 Check
- Reports each knowledge source with its type (SharePoint, Graph Connector, etc.) and live status
- **PASSED**: status is \ctive\, \completed\, \indexed\, \
eady\, or \succeeded\
- **FAILED**: status is \pending\, \crawling\, \queued\, \provisioning\, \ailed\, or \rror\
- **SKIPPED**: when PVA authentication is unavailable (no fallback/false greens)

### Code Quality Rule
- Added \No fabricated URLs\ rule to \.github/copilot-instructions.md\ to prevent agents from inventing documentation links

## Testing
- Tested against live tenant (ESS Golden SF Weekly agent)
- All 3 knowledge sources (2x SharePoint, 1x Graph Connector) correctly report \ctive  indexed and ready\
- Integration testing of failure states pending  requires coordination with engineers who investigated the original ICM to identify how to trigger non-ready states

## Open Items
- Need confirmation from ICM engineers on exact status string when SharePoint crawl is in-progress
- No doc link added (none exists yet)  TODO comments mark where links should go once documentation is created

Closes #24


scripts/flightcheck/checks/local_files.py
```diff
@@ -12,6 +12,7 @@
 import os
 import re
 from pathlib import Path
+from urllib.parse import urljoin
```
GrahamMcMynn: **[standards]** Unused import: `urljoin` is imported but never used anywhere in the new code. The `_check_knowledge_sources` functions construct no URLs using `urljoin`.

**Fix:** Remove `from urllib.parse import urljoin`.

scripts/flightcheck/pva_client.py
```diff
@@ -0,0 +1,183 @@
...
+        url = (
+            f"{self._gateway_url}/api/botmanagement/v1"
+            f"/environments/{self._bap_env_id}"
+            f"/bots/{bot_id}/content/botcomponents"
+        )
+        headers = {
+            "Authorization": f"Bearer {self._pva_token}",
+            "Content-Type": "application/json",
+            "x-ms-client-tenant-id": self.tenant_id,
+            "x-cci-tenantid": self.tenant_id,
```
GrahamMcMynn, line 178: **[bug]** `get_knowledge_sources` silently returns `[]` on HTTP errors (non-2xx) and caught exceptions. The caller `_check_knowledge_sources_via_gateway` wraps the call in try/except to distinguish "API failed" from "no sources returned"  but since this method catches its own exceptions and returns `[]`, the caller's except block never fires.

**Result:** A 403/500/timeout will appear as *"no matching components returned from API"* (misleading) instead of the intended *"Could not query Copilot Studio API: <error>"* message.

**Fix:** Let exceptions propagate (remove the inner try/except), or raise on non-OK status:
```python
resp = requests.post(url, headers=headers, json={}, timeout=60)
if not resp.ok:
    raise RuntimeError(
        f"Island Gateway returned {resp.status_code}: {resp.text[:200]}"
    )
data = resp.json()
```

GrahamMcMynn (commented):

scripts/flightcheck/checks/local_files.py
GrahamMcMynn: Fixed: The urljoin import was removed in the previous push.

scripts/flightcheck/pva_client.py
GrahamMcMynn: Fixed: get_knowledge_sources now raises RuntimeError on non-OK status instead of silently returning []. The caller try/except correctly catches and reports the error message.

.github/copilot-instructions.md
```diff
@@ -285,6 +285,39 @@ partial workflow.
 3. **Test in Copilot Studio**: Open [Copilot Studio](https://copilotstudio.microsoft.com/) to test conversations
 4. **Publish**: Publish the agent in the Copilot Studio portal to make changes live
 
+## Code Quality Rules
+
+### No fabricated URLs
+
+**NEVER invent, guess, or hallucinate URLs.** Every URL in code (doc_link fields,
+remediation messages, comments, README references) must point to a page you have
+confirmed exists — either by finding it in the existing codebase or by verifying
```
johnguy0: The rule requires verification with a web fetch, but agents may not have that tool. Make the requirement tool-agnostic, for example verify the URL exists by an available HTTP or repository lookup, and keep the rule in a repo-wide section where code authors will see it.

scripts/flightcheck/checks/local_files.py
```diff
@@ -288,3 +291,152 @@ def _check_template_configs(agent_path: Path, label: str) -> list[CheckResult]:
...
+        results.append(CheckResult(
+            checkpoint_id="CONFIG-013", category=f"Knowledge Sources ({label})",
+            priority=Priority.HIGH.value, status=Status.WARNING.value,
+            description=f"{label}: Knowledge source crawl status",
+            result=f"Could not query Copilot Studio API: {e}",
+            remediation="Check authentication and retry. Verify status manually in Copilot Studio → Knowledge.",
+        ))
+        return results
+
+    if not sources:
```
johnguy0: Only the zero-source case is checked. If there are three local knowledge files and the API returns one ready component, the check reports success. Build a local-to-remote match by stable fields and report each missing or unmatched source.

johnguy0 (commented):
This adds CONFIG-013 and a new PVA client to read Island Gateway component status. I am not comfortable approving yet because the check can pass when configured local knowledge sources are missing from the API response, and it treats lifecycle state as crawl readiness. It also hides gateway discovery failures as skips. Please tighten the API contract before merging.

scripts/flightcheck/checks/local_files.py
```diff
@@ -288,3 +291,152 @@ def _check_template_configs(agent_path: Path, label: str) -> list[CheckResult]:
...
+            description=f"{label}: Knowledge source readiness",
+            result=f"{len(knowledge_files)} local knowledge file(s) but no matching components returned from API",
+            remediation="Knowledge sources may not be published yet. Publish the agent in Copilot Studio.",
+        ))
+        return results
+
+    all_ready = True
+    for source in sources:
+        name = source.get("displayName", "Unknown")
+        status_value = (source.get("status") or source.get("state") or "unknown").lower()
```
johnguy0: This treats `state` as crawl readiness when `status` is absent. `Active` can be lifecycle state, not proof that indexing completed, so CONFIG-013 can pass without an index signal. Only pass when the API returns a field that specifically represents crawl/index readiness, otherwise return WARNING with the raw state.

scripts/flightcheck/cli.py
```diff
@@ -154,6 +155,20 @@ def main():
         print(f"  Power Platform: WARNING — {e}")
         print("  (Some checks will be skipped)")
 
+    print("Authenticating to Copilot Studio (Island Gateway)...")
```
johnguy0: This prompts for Copilot Studio auth on every scope. CONFIG-013 only runs with local file checks, so `--scope prerequisites` and similar scopes should not open a second interactive login. Initialize PVA only for scopes that include the knowledge readiness check.

scripts/flightcheck/pva_client.py
```diff
@@ -0,0 +1,182 @@
...
+            resp = requests.get(url, headers=headers, timeout=30)
+            if resp.status_code in (401, 403):
+                # Try non-admin endpoint
+                url = (
+                    f"{BAP_BASE}/providers/Microsoft.BusinessAppPlatform/environments"
+                    "?api-version=2021-04-01"
+                )
+                resp = requests.get(url, headers=headers, timeout=30)
+            if not resp.ok:
+                return
```
johnguy0: Gateway discovery failures are swallowed here. That turns API errors into an unconfigured client and later a SKIPPED check, which hides the reason CONFIG-013 could not run. Raise a RuntimeError with the endpoint/status or exception so the runner surfaces a WARNING with the actual failure.

scripts/flightcheck/cli.py
saengland: @GrahamMcMynn - gate PVA init on the scope. The simplest check is "any registered check in this scope needs PVA":

```python
# Scopes that include CONFIG-013 (the only check that needs PVA today).
_PVA_SCOPES = {"full", "local"}

# ...

pva = None
if args.scope in _PVA_SCOPES:
    print("Authenticating to Copilot Studio (Island Gateway)...")
    pva = PVAClient(tenant_id, env_url)
    try:
        pva.authenticate()
        if pva.is_configured:
            print("  Copilot Studio: OK")
        else:
            print("  Copilot Studio: WARNING - Could not discover gateway URL")
            print("  (CONFIG-013 will be skipped)")
    except Exception as e:
        print(f"  Copilot Studio: WARNING - {e}")
        print("  (CONFIG-013 will be skipped)")
        pva = None
else:
    print("Skipping Copilot Studio auth (not required for this scope).")

# ...

runner.pva = pva  # may be None for scopes that don't need it
```

Future-proofing if more checks start to need PVA: derive the scope set from the check registry instead of a hardcoded constant. Something like:

```python
_PVA_DEPENDENT_CHECKS = {"check_knowledge_source_readiness"}  # function names
checks_for_scope = SCOPE_MAP.get(args.scope, FULL_SCOPE)
needs_pva = any(fn.__name__ in _PVA_DEPENDENT_CHECKS for _, fn in checks_for_scope)
```

Either approach kills the second interactive login on `--scope prerequisites` and friends.

scripts/flightcheck/checks/local_files.py
saengland: @GrahamMcMynn - the fix is to stop using `state` as a crawl-readiness fallback. Treat them as separate signals:

```python
# `status` is the crawl/index readiness signal we want.
# `state` is lifecycle (Active/Inactive/Provisioning/etc.) and does NOT
# tell us whether indexing finished. Falling back to `state` here is the
# false-pass: an Active-but-still-crawling source would be reported PASSED.
crawl_status = (source.get("status") or "").strip().lower()
lifecycle_state = (source.get("state") or "").strip().lower()

if crawl_status in _READY_STATUSES:
    # ...PASSED path...
elif crawl_status in _NOT_READY_STATUSES:
    # ...FAILED path...
elif crawl_status:
    # Unknown status string, but the API DID return a status field.
    # WARNING with the raw value so we learn the new status over time.
    all_ready = False
    results.append(CheckResult(
        checkpoint_id="CONFIG-013", category=f"Knowledge Sources ({label})",
        priority=Priority.HIGH.value, status=Status.WARNING.value,
        description=f"{label}: '{name}' ({source_type})",
        result=f"Unknown crawl status: {crawl_status}",
        remediation=(
            "Verify in Copilot Studio -> Knowledge whether the source has "
            "finished indexing. File an issue so this status string can be "
            "added to the readiness allowlist."
        ),
    ))
else:
    # No `status` returned at all - we only have lifecycle `state`.
    # That is NOT proof of indexing; surface as WARNING, not PASSED.
    all_ready = False
    results.append(CheckResult(
        checkpoint_id="CONFIG-013", category=f"Knowledge Sources ({label})",
        priority=Priority.HIGH.value, status=Status.WARNING.value,
        description=f"{label}: '{name}' ({source_type})",
        result=(
            f"Crawl readiness unknown (lifecycle state={lifecycle_state or 'unknown'}). "
            "API did not return a `status` field for this source."
        ),
        remediation=(
            "Verify in Copilot Studio -> Knowledge whether the source has "
            "finished indexing before deploying."
        ),
    ))
```

This makes the check fail-safe: missing status -> WARNING, not PASSED. Once you (or the ICM engineers John mentioned in the open items) confirm the in-progress status string, just add it to `_NOT_READY_STATUSES` and the WARNING for "API did not return status" stays as the catch-all.

scripts/flightcheck/checks/local_files.py
saengland: @GrahamMcMynn - here's the local-to-remote match. Replace the `if not sources:` block with a per-file diff so a customer with 3 local sources and 1 returned component sees 2 explicit FAILED rows, not a misleading PASSED:

```python
# Build remote name set for matching against local files.
# Use displayName as the join key - it's the only field stable across
# both the local YAML (`displayName:` at the top of *.mcs.yml) and the
# Island Gateway response.
remote_by_name = {
    (s.get("displayName") or "").strip().lower(): s
    for s in sources
}

# Resolve each local knowledge file to its remote counterpart (or fail).
matched = []
unmatched = []
for kf in knowledge_files:
    try:
        with open(kf, "r", encoding="utf-8") as f:
            local = yaml.safe_load(f) or {}
    except (OSError, yaml.YAMLError) as e:
        unmatched.append((kf.name, f"could not parse local file: {e}"))
        continue
    local_name = (local.get("displayName") or "").strip().lower()
    if not local_name:
        unmatched.append((kf.name, "missing displayName in local file"))
        continue
    remote = remote_by_name.get(local_name)
    if remote is None:
        unmatched.append((kf.name, f"no matching component in Copilot Studio (looked for displayName={local_name!r})"))
        continue
    matched.append((kf, remote))

# Report each unmatched local file as a distinct failure.
for fname, reason in unmatched:
    results.append(CheckResult(
        checkpoint_id="CONFIG-013", category=f"Knowledge Sources ({label})",
        priority=Priority.HIGH.value, status=Status.FAILED.value,
        description=f"{label}: '{fname}'",
        result=f"Local source not found in Copilot Studio: {reason}",
        remediation=(
            "Publish the agent in Copilot Studio so this knowledge source "
            "is provisioned, or remove the local file if it should not exist."
        ),
    ))

# Then iterate `matched` (instead of `sources`) for the per-source PASSED/FAILED loop.
```

Two notes:
- Pick whatever the actual stable join key is - if local files don't have `displayName` at the top level, swap it for whatever you normalize off the local YAML during `/setup`.
- After this lands, the existing "no matching components returned from API" warning becomes dead code (every local file will surface as either matched or in `unmatched`). Drop it.

scripts/flightcheck/pva_client.py
saengland: @GrahamMcMynn - the `_discover_gateway` block currently has two `return` statements that swallow real failures (the `if not resp.ok: return` after a 401/403 retry, and the bare `except Exception: return`). Replace with raise-on-failure:

```python
def _discover_gateway(self, bap_token: str):
    """Find the PVA gateway URL and BAP environment ID for this environment.

    Raises:
        RuntimeError: when the BAP environment list cannot be fetched, or
            when no environment matches the configured Dataverse URL. The
            caller surfaces this as a WARNING with the actual reason.
    """
    headers = {"Authorization": f"Bearer {bap_token}", "Accept": "application/json"}

    # Try admin endpoint first, fall back to non-admin on 401/403.
    url = (
        f"{BAP_BASE}/providers/Microsoft.BusinessAppPlatform/scopes/admin/environments"
        "?api-version=2021-04-01"
    )
    try:
        resp = requests.get(url, headers=headers, timeout=30)
        if resp.status_code in (401, 403):
            url = (
                f"{BAP_BASE}/providers/Microsoft.BusinessAppPlatform/environments"
                "?api-version=2021-04-01"
            )
            resp = requests.get(url, headers=headers, timeout=30)
    except requests.RequestException as e:
        raise RuntimeError(f"BAP environments request failed: {e}") from e

    if not resp.ok:
        raise RuntimeError(
            f"BAP environments returned {resp.status_code}: {resp.text[:200]}"
        )

    org_match = self.env_url.split("//")[1].split(".")[0] if "//" in self.env_url else ""
    envs = resp.json().get("value", [])
    for env in envs:
        props = env.get("properties", {})
        linked = props.get("linkedEnvironmentMetadata", {})
        instance_url = linked.get("instanceUrl", "")
        if org_match and org_match in instance_url:
            self._bap_env_id = env.get("name")
            runtime = props.get("runtimeEndpoints", {})
            self._gateway_url = runtime.get("microsoft.PowerVirtualAgents")
            return

    raise RuntimeError(
        f"No BAP environment matched Dataverse org '{org_match}'. "
        "User may not have access to this environment, or the environment "
        "type is not Copilot Studio-enabled."
    )
```

Then in `cli.py`'s PVA init (the block I replied on at line 158), the existing `try/except Exception as e:` will catch the `RuntimeError` and emit `"  Copilot Studio: WARNING - <real reason>"` instead of the current "Could not discover gateway URL" generic. CONFIG-013 will then SKIP with a meaningful "Cannot verify - <reason>" instead of swallowing it.

.github/copilot-instructions.md
saengland: @GrahamMcMynn - here's a concrete fix for both halves of John's note (tool-agnostic verification + repo-wide placement):

```diff
- ### No fabricated URLs
-
- **NEVER invent, guess, or hallucinate URLs.** Every URL in code (doc_link fields,
- remediation messages, comments, README references) must point to a page you have
- confirmed exists — either by finding it in the existing codebase or by verifying
- it with a web fetch.
+ ### No fabricated URLs
+
+ **NEVER invent, guess, or hallucinate URLs.** Every URL in code (doc_link fields,
+ remediation messages, comments, README references) must point to a page you have
+ confirmed exists. Verify by ANY of:
+
+ - Finding the same URL already cited elsewhere in the repo (`git grep` / repo search)
+ - Fetching the URL over HTTP and confirming a 2xx response (if your tooling supports it)
+ - Pulling the page from an official Microsoft Learn / Docs reference you already trust
+
+ Pick whichever is available - the rule is "don't ship unverified URLs", not "use a specific verification tool".
```

For the placement half: this section is fine in `.github/copilot-instructions.md` (where any agent working in the repo will see it). If you want extra reach, also drop a one-liner under the **Code Quality** section of `CONTRIBUTING.md` that points back here:

```markdown
- **No fabricated URLs.** See `.github/copilot-instructions.md` -> "No fabricated URLs" for the verification rule.
```

That way human contributors see it during their CONTRIBUTING.md read and agents see it in copilot-instructions.

GrahamMcMynn: Fixed in commit 56bf859. Rule now lists three tool-agnostic verification options (git grep / HTTP fetch / official Microsoft Learn reference) - the principle is no unverified URLs, not a specific tool. Also added a pointer in CONTRIBUTING.md under section 5 (Code quality) so human contributors discover it during their CONTRIBUTING.md read.

scripts/flightcheck/checks/local_files.py
GrahamMcMynn: Partially fixed in commit 56bf859. The principle (do not false-PASS when local count > remote count) is implemented as a count-comparison WARNING. The per-file matching with displayName as the join key was investigated against live data but does not work in this codebase: SharePoint sources return their site URL as displayName (e.g. https://employeehub.sharepoint.com/...) while local filenames come from Dataverse botcomponents.name (e.g. documents.mcs.yml). There is no stable join key. The relationship is also not 1:1 - the test tenant has 4 remote sources and only 2 local files, by design. The count check catches the original false-PASSED scenario without the noise that broken per-file matching would create. Documented this reasoning inline in the function docstring.

scripts/flightcheck/checks/local_files.py
GrahamMcMynn: Fixed in commit 56bf859. Removed Active from _READY_STATUSES and Inactive from _NOT_READY_STATUSES (both are lifecycle state values, not crawl signals). Stopped using state as a fallback when status is missing. New behavior: PASS only when status is in {completed, indexed, ready, succeeded}; missing or unknown status returns WARNING with the raw lifecycle state for context. Verified against live tenant: the API actually returns status=Active for both crawl and lifecycle, with no separate completion field, so healthy sources now produce WARNING (correct per the principle - never PASS without a definitive ready signal). Acknowledged limitation: the API does not expose a richer crawl signal today; this is the safest behavior until it does.

scripts/flightcheck/cli.py
GrahamMcMynn: Fixed in commit 56bf859. PVA authentication is now gated on scope - only initialized for full and local scopes (the only ones registering CONFIG-013 today). Verified: --scope prerequisites now prints Skipping Copilot Studio auth (not required for this scope). instead of triggering the second interactive login. Used the simpler hardcoded scope set rather than scanning the registry; happy to refactor to the function-name approach if more PVA-dependent checks land.

scripts/flightcheck/pva_client.py
GrahamMcMynn: Fixed in commit 56bf859. _discover_gateway now raises RuntimeError on every failure path (network exception, non-OK HTTP response, no environment match) instead of silently returning. The cli.py try/except already wraps this call and surfaces e in the WARNING line, so the user now sees the real reason instead of generic Could not discover gateway URL. Added docstring documenting the raised exceptions.

johnguy0 (commented):
Round 2 on PR #37 (CONFIG-013). All 4 blockers from the prior review are addressed in 56bf859.

1. ✅ `No fabricated URLs` rule is tool-agnostic now (git grep / HTTP fetch / Microsoft Learn / Docs reference), and CONTRIBUTING.md section 5 points back to it so human contributors discover the rule during their CONTRIBUTING.md read.
2. ✅ Local-vs-remote count check catches the original false-PASSED scenario (local > 0, remote = 0). Per-file matching by `displayName` is not viable, the local filename comes from `botcomponents.name` while the gateway returns the SharePoint site URL as `displayName`, and the relationship is not 1:1 by design. Reasoning captured in the function docstring. Acceptable as-is.
3. ✅ `Active` is out of `_READY_STATUSES` and `Inactive` is out of `_NOT_READY_STATUSES`. Missing or unknown crawl `status` now WARNs with the raw lifecycle state instead of falling back to `state` and false-PASSING.
4. ✅ PVA auth is gated on `args.scope in ("full", "local")`. `--scope prerequisites` no longer triggers the second interactive login.
5. ✅ `_discover_gateway` raises `RuntimeError` on every failure path (network exception, non-OK HTTP response, no environment match). The cli.py try/except surfaces the real reason in the WARNING line.

One operational note, not blocking: with `Active` no longer in `_READY_STATUSES` and the live API returning `status=Active` for both crawl and lifecycle today, healthy sources will produce WARNING rather than PASSED until the gateway exposes a richer crawl signal. That is the right behavior per the principle (no PASS without a definitive ready signal), but it means CONFIG-013 will sit at perpetual yellow in practice. Worth filing a follow-up issue to track getting a real `crawl_status` field from the gateway folks so this can flip back to green.

Ready for approval.

scripts/flightcheck/pva_client.py
```diff
@@ -0,0 +1,197 @@
...
+                [BAP_SCOPE], prompt="select_account"
+            )
+
+        if cache.has_state_changed:
+            os.makedirs("my", exist_ok=True)
+            with open(cache_path, "w") as f:
+                f.write(cache.serialize())
+
+        # Discover gateway URL from BAP
+        if bap_result and "access_token" in bap_result:
```
saengland, line 96: **MEDIUM** - BAP token acquisition can silently fail. If `acquire_token_silent` returns no token and `acquire_token_interactive` then returns a result without `access_token` (user cancels, denies consent, network glitch), `_discover_gateway` is simply skipped with no error raised. The PVA token call directly above (line 75-77) correctly raises `RuntimeError` on the same condition; this branch swallows it and the customer ends up with the opaque "Could not discover gateway URL" warning in cli.py with no underlying reason.

Fix: mirror the PVA-token error handling. After the second `acquire_token_interactive`:

```python
if not bap_result or "access_token" not in bap_result:
    error = (bap_result or {}).get("error", "unknown_error")
    raise RuntimeError(f"BAP auth failed (needed for gateway discovery): {error}")
```

(Note: scrub `error_description` like `auth.py` does - CWE-209.)

scripts/flightcheck/checks/local_files.py
```diff
@@ -288,3 +290,199 @@ def _check_template_configs(agent_path: Path, label: str) -> list[CheckResult]:
...
+    bot_id = None
+    if runner and hasattr(runner, "config"):
+        bot_id = runner.config.get("agent", {}).get("botId")
+
+    if not pva or not hasattr(pva, "is_configured") or not pva.is_configured or not bot_id:
+        results.append(CheckResult(
+            checkpoint_id="CONFIG-013", category=f"Knowledge Sources ({label})",
+            priority=Priority.HIGH.value, status=Status.SKIPPED.value,
+            description=f"{label}: Knowledge source readiness",
+            result="Cannot verify — Copilot Studio API authentication required",
```
saengland, line 343: **MEDIUM** - The SKIPPED message "Cannot verify - Copilot Studio API authentication required" conflates four distinct failure modes that each need different remediation: (a) PVA wasn't authenticated for this scope, (b) PVA auth ran but gateway discovery failed (BAP env mismatch / no admin access), (c) `is_configured` is False for some other reason, (d) config is missing `agent.botId` entirely.

A user whose only problem is a missing `botId` in `config.json` will be sent down an auth-troubleshooting path. The check also sets no `remediation` field, so there is no actionable next step in the report.

Fix: branch the SKIPPED reason on which condition actually failed and set a remediation:

```python
if not bot_id:
    reason = "Missing agent.botId in my/config.json"
    remediation = "Re-run /setup to populate agent.botId, or add it manually."
elif not pva:
    reason = "Copilot Studio (Island Gateway) was not authenticated"
    remediation = "Run flightcheck with --scope full or --scope local so the Island Gateway sign-in runs."
else:
    reason = "Could not discover Copilot Studio gateway URL for this environment"
    remediation = "Verify your account has access to the target Power Platform environment and that it is Copilot Studio-enabled."
```

scripts/flightcheck/checks/local_files.py
```diff
@@ -288,3 +290,199 @@ def _check_template_configs(agent_path: Path, label: str) -> list[CheckResult]:
     ))
 
     return results
+
+
+# Knowledge source crawl readiness statuses (the `status` field on the API response).
+# These represent the indexing lifecycle, distinct from the lifecycle `state` field.
+_READY_STATUSES = {"completed", "indexed", "ready", "succeeded"}
+# Statuses that indicate the source is not yet ready for deployment.
+_NOT_READY_STATUSES = {"pending", "crawling", "queued", "provisioning", "failed", "error"}
```
saengland, line 299: **MEDIUM** - `_NOT_READY_STATUSES` buckets terminal failure states (`failed`, `error`) together with transient in-progress states (`pending`, `crawling`, `queued`, `provisioning`). They produce identical FAILED results with the remediation "Wait for the crawl to finish, or check for errors in Copilot Studio".

A permanently failed crawl will never become ready by waiting; the customer needs to fix the underlying source config (auth, URL, permission scope) and re-trigger the crawl. Telling them to wait is wrong advice.

Fix: split the set and remediation:

```python
_TRANSIENT_STATUSES = {"pending", "crawling", "queued", "provisioning"}
_FAILED_STATUSES = {"failed", "error"}
```

Add a third elif branch for `_FAILED_STATUSES` with remediation like: "Knowledge source `{name}` failed to index. Open Copilot Studio -> Knowledge -> {name} to see the error detail, fix the source configuration, and republish the agent." Keep the existing wait-for-crawl message for `_TRANSIENT_STATUSES` only.

scripts/flightcheck/pva_client.py
```diff
@@ -0,0 +1,197 @@
...
+            "Content-Type": "application/json",
+            "x-ms-client-tenant-id": self.tenant_id,
+            "x-cci-tenantid": self.tenant_id,
+            "x-cci-bapenvironmentid": self._bap_env_id,
+            "x-cci-cdsbotid": bot_id,
+        }
+
+        resp = requests.post(url, headers=headers, json={}, timeout=60)
+        if not resp.ok:
+            raise RuntimeError(
```
saengland, line 185: **LOW** - `get_knowledge_sources` raises `RuntimeError` on any non-2xx response, which is inconsistent with the rest of the kit's HTTP clients. `graph_client._get` and `pp_admin_client._get` both return `{"_error": "insufficient_permissions", "_status": 401|403}` on 401/403 instead of raising, so callers can distinguish "API said no" from "network exploded".

The PR's caller in local_files.py wraps the call in a bare `except Exception` so it works, but every future caller will need to remember to do the same.

Fix: match the kit pattern - return a structured marker on auth failures, raise only on transport errors and unexpected statuses:

```python
if resp.status_code in (401, 403):
    return [{"_error": "insufficient_permissions", "_status": resp.status_code}]
resp.raise_for_status()
```

Then update `_check_knowledge_sources_via_gateway` to recognise the marker and emit a SKIPPED/WARNING result with a clear "no Copilot Studio access" message.

scripts/flightcheck/checks/local_files.py
```diff
@@ -288,3 +290,199 @@ def _check_template_configs(agent_path: Path, label: str) -> list[CheckResult]:
...
+        return results
+
+    all_ready = True
+    for source in sources:
+        name = source.get("displayName", "Unknown")
+        # `status` is the crawl/index readiness signal we want.
+        # `state` is lifecycle (Active/Inactive/Provisioning/etc.) and does NOT
+        # tell us whether indexing finished. Falling back to `state` here would
+        # be a false-pass: an Active-but-still-crawling source could be PASSED.
+        crawl_status = (source.get("status") or "").strip().lower()
```
saengland, line 404: **LOW** - `(source.get("status") or "").strip().lower()` assumes status is either a string or None. The Island Gateway is a JSON API; if a future version of the response returns `status` as a number, bool, or nested object, `.strip()` raises `AttributeError` and the whole check crashes for the agent instead of degrading to WARNING. Same pattern on the next line for `state`.

Fix: coerce to str defensively before string ops:

```python
crawl_status = str(source.get("status") or "").strip().lower()
lifecycle_state = str(source.get("state") or "").strip().lower()
```

saengland (changes requested):
Reviewed CONFIG-013 + new pva_client. Solid direction - the gateway-discovery + knowledge-source readiness flow is the right shape and slots into the existing scope-runner correctly. 5 findings:

| # | Sev | File | Issue |
|---|---|---|---|
| 1 | MEDIUM | pva_client.py | BAP token acquisition silently fails - skips `_discover_gateway` instead of raising |
| 2 | MEDIUM | local_files.py | SKIPPED conflates 4 distinct failure modes; wrong remediation for missing `agent.botId` |
| 3 | MEDIUM | local_files.py | `_NOT_READY_STATUSES` buckets terminal failures with transient states - tells customers to wait when they need to fix |
| 4 | LOW | pva_client.py | `get_knowledge_sources` raises on 401/403 instead of returning `{_error: insufficient_permissions}` like `graph_client._get` and `pp_admin_client._get` |
| 5 | LOW | local_files.py | `(source.get('status') or '').strip().lower()` assumes string - crashes if API returns a future int/object |

#1, #2, #3 are the blockers. #4 and #5 are consistency nits.

saengland:

Action needed on your side:

saengland:
Reopened as #57 against main. The previous review (this PR) is preserved here for context.

PR #38, opened 2026-05-05
Add flightcheck developer guide (AGENTS.md)

GrahamMcMynn:
## Summary

Adds a developer guide scoped to the flightcheck tool at \scripts/flightcheck/AGENTS.md\. This documents the knowledge needed to efficiently add new checks without trial-and-error against live APIs.

## Motivation

While implementing CONFIG-013 (PR #37), significant time was spent:
- Trying the wrong API (Dataverse \otcomponents.data\ is YAML config, not status)
- Discovering the Island Gateway API exists and is the correct source for runtime state
- Finding the required \x-cci-*\ headers via an external repo
- Learning that BAP env ID  Dataverse env ID

This guide captures those learnings so the next check can be implemented correctly on the first attempt.

## Contents

- **Architecture Overview**  table mapping API layers to clients, auth scopes, and available data
- **Island Gateway API Reference** — endpoint, headers, env ID discovery, response structure
- **Dataverse \otcomponents\ Entity**  key gotchas (YAML not JSON, no \msdyn_\ prefix, statecode  crawl status)
- **Design Principles**  no misleading results, fail loudly, no fabricated URLs
- **Adding a New Check**  step-by-step checklist including guidance for when a new API is needed


scripts/flightcheck/AGENTS.md
```diff
@@ -0,0 +1,136 @@
...
+to the FlightCheck tool.
+
+## Architecture Overview
+
+FlightCheck validates customer deployments by querying multiple APIs. Each API
+requires its own authentication and has different data available:
+
+| API Layer | Client | Auth Resource / Scope | What's Available |
+|-----------|--------|----------------------|------------------|
+| Dataverse | `auth.py` | `{env_url}/user_impersonation` | Bot components (topics, variables, knowledge source *config*), template configs, solution metadata |
```
johnguy0: `auth.py` lives at `scripts/auth.py`, not in this directory. Fully qualify it (`scripts/auth.py` or `../auth.py`) so a contributor reading `scripts/flightcheck/AGENTS.md` doesn't grep this folder for it. Same applies to the import in `cli.py` — the path manipulation that makes `from auth import authenticate` work isn't obvious.

scripts/flightcheck/AGENTS.md
```diff
@@ -0,0 +1,136 @@
...
+| Microsoft Graph | `graph_client.py` | `https://graph.microsoft.com/.default` | Licenses, user roles, Entra app registrations, CA policies |
+| Power Platform Admin (BAP) | `pp_admin_client.py` | `https://service.powerapps.com//.default` | Environments, cloud flows, connections, DLP policies |
+| Island Gateway (Copilot Studio) | `pva_client.py` | `96ff4394-9197-43aa-b393-6a41652e21f8/.default` | Live bot component status, model config, knowledge source *runtime state* |
+
+**Critical distinction:** Dataverse stores *configuration* (YAML source files,
+component metadata). The Island Gateway stores *runtime state* (active/inactive
+status, crawl progress, model assignments). If you need to check whether
+something is *configured*, query Dataverse. If you need to check whether it's
+*working*, query the Island Gateway.
```
johnguy0: Worth tightening this framing. "If you need to check whether something is *configured*, query Dataverse" isn't how the kit reads today — configured state comes from local YAML in `checks/local_files.py`, not from Dataverse. Dataverse is the source of truth for things the kit doesn't have locally (templates, solution metadata, fields stamped server-side after publish).

A cleaner three-way split:
- **Local YAML** = what the developer authored (the kit's source of truth for config)
- **Dataverse** = server-side state for things not in the local kit
- **Island Gateway** = runtime state (active, indexed, crawl progress)

Otherwise the next contributor writes a Dataverse query for something `local_files.py` already validates.

scripts/flightcheck/AGENTS.md
```diff
@@ -0,0 +1,136 @@
...
+
+3. **One check, one concern.** Each check should validate exactly one thing.
+   Don't bundle multiple validations under a single checkpoint ID.
+
+4. **Follow existing client patterns.** New API integrations should follow the
+   same structure as `graph_client.py` / `pp_admin_client.py` / `pva_client.py`:
+   - Class with `authenticate()` method
+   - Uses shared MSAL token cache at `my/.token_cache.bin`
+   - Initialized in `cli.py`, attached to `runner`
+   - Gracefully skips if auth fails (print warning, set to None)
```
johnguy0, line 130: The "attached to `runner`" pattern works, but `FlightCheckRunner` declares no shape — clients show up as ad-hoc attributes set in `cli.py`. New contributors have to read `cli.py` to know what's available. Either list the attributes here (`runner.config`, `runner.env_url`, `runner.dv_token`, `runner.env_id`, `runner.graph`, `runner.pp_admin`, `runner.pva`) or — better — add them as typed Optional fields on `FlightCheckRunner` so they're discoverable from the type and IDE.

johnguy0 (commented):
Doc is genuinely useful — it captures the gotchas you spent CONFIG-013 hours on, and that's exactly the kind of thing that pays for itself the next time someone adds a check. Two structural things worth tightening before merge:

1. The "Dataverse = configured, Gateway = working" framing in the Architecture Overview doesn't match what the kit actually does today. Configured state in this kit comes from local YAML files (`scripts/flightcheck/checks/local_files.py`), not from Dataverse. If a future check author follows this guide literally, we'll get a duplicate Dataverse query for something already validated locally — two sources of truth that can disagree.

2. The "attached to `runner`" pattern is documented but the shape of `runner` isn't declared anywhere. New contributors have to read `cli.py` to know what's on it. Either enumerate the attributes here (`runner.config`, `runner.env_url`, `runner.dv_token`, `runner.env_id`, `runner.graph`, `runner.pp_admin`, `runner.pva`) or — better — declare them as typed Optional fields on `FlightCheckRunner` in `runner.py` so they're discoverable from the type.

The "No misleading results" principle is the most valuable thing in this doc. Don't lose it in the cleanup. Inline nits on file paths.

johnguy0 (commented):
Nothing security-relevant in this PR. Worth noting for the record:

- The shared MSAL token cache at `my/.token_cache.bin` is referenced as a contributor pattern. Already covered by `my/**` in `.gitignore`, so no leak risk.
- `96ff4394-9197-43aa-b393-6a41652e21f8` is the public Power Virtual Agents 1P app ID, not a secret.
- No untrusted input flowing anywhere. No new dependencies.

Deferring to other personas.

johnguy0 (commented):
Doc-only PR, no perf angle. The closest thing is the documented "Read all bot components" call which has no pagination, but that's a faithful description of how the Island Gateway endpoint actually behaves — not something this doc introduced. Deferring.

scripts/flightcheck/AGENTS.md
```diff
@@ -0,0 +1,136 @@
...
+- **The `data` column is YAML** (the `.mcs.yml` file content), NOT JSON. Do
+  not try to `json.loads()` it. Parse with a YAML library if needed.
+- **`statecode`/`statuscode`** are standard Dataverse record status (Active=0/1),
+  NOT the runtime crawl/index status.
+
+## Design Principles
+
+1. **No misleading results.** If a check cannot actually validate what it claims
+   to validate (e.g., missing API access), report `SKIPPED` — never `PASSED`.
+   A check that always passes is worse than no check at all.
```
johnguy0: This principle is the most valuable thing in the doc, but the `SKIPPED` vs `PASSED` framing under-specifies. The runner has five statuses and existing checks split between them inconsistently:

- `publishing.py` uses `NOT_CONFIGURED` for "manual verification required"
- `local_files.py` uses `SKIPPED` when API auth is unavailable (PR #37)
- Most existing checks use `WARNING` when an API call errors

A new contributor reading this principle won't know which to pick. Spell out the decision tree:

- `PASSED` — we ran the check and the result is good
- `FAILED` — we ran the check and the result is bad
- `WARNING` — we tried but the API errored; surface the error
- `SKIPPED` — we couldn't run the check (API unavailable, missing creds, no relevant data on disk)
- `NOT_CONFIGURED` — feature isn't turned on, or item requires manual verification
- `ERROR` — the check itself crashed (the runner sets this for you)

The principle then becomes: never return `PASSED` when you can't actually validate.

scripts/flightcheck/AGENTS.md
```diff
@@ -0,0 +1,136 @@
...
+   table above). If the data isn't available from any existing client, research
+   the correct API — check the `microsoft/MCS-Agent-Builder` repo's
+   `knowledge/cache/` folder for API documentation, or inspect network traffic
+   in the relevant admin UI (Copilot Studio, Power Platform Admin Center, etc.)
+2. If a new client is needed, create it following existing patterns and add it
+   to the table above
+3. Add the check function to the appropriate module in `checks/`
+4. Wire it into `_check_single_agent()` (for per-agent checks) or the
+   appropriate scope runner
+5. Pass `runner` through so the check can access API clients
```
johnguy0, line 188: Steps 3–5 assume the contributor already knows the `CheckResult` shape. Drop in a minimal example here so they don't have to grep — that's the trial-and-error this whole doc is meant to prevent:

```python
from ..runner import CheckResult, Status, Priority

def run_my_checks(runner) -> list[CheckResult]:
    results = []
    # ... do the check, possibly using runner.graph / runner.pva / etc. ...
    results.append(CheckResult(
        checkpoint_id="CONFIG-099",
        category="Local Files",
        priority=Priority.HIGH.value,
        status=Status.PASSED.value,
        description="Thing is set up correctly",
        result="Found N things",
        remediation="",  # only on non-pass
    ))
    return results
```

scripts/flightcheck/AGENTS.md
```diff
@@ -0,0 +1,136 @@
...
+```
+POST /api/botmanagement/v1/environments/{bapEnvId}/bots/{botId}/content/botcomponents
+Body: {}
+```
+
+Response contains `botComponentChanges[]`, each with a `component` object:
+- `component.$kind`: `DialogComponent`, `KnowledgeSourceComponent`, `GptComponent`,
+  `GlobalVariableComponent`, `CustomEntityComponent`
+- `component.state` / `component.status`: runtime status (e.g., "Active", "Inactive")
+- `component.displayName`, `component.id`, `component.schemaName`
```
johnguy0, line 77: This list of `$kind` values reads as exhaustive but it's not — it's the ones FlightCheck currently cares about. If a contributor writes `if comp.$kind not in [...]: error()`, they'll false-fail on agents with components we haven't enumerated. Phrase as "common kinds include" and call out that the API may return others.

Same for the `componenttype` integers in the Dataverse section (lines 78–80) — "common types are" is already there, good. Just match that hedge here.

johnguy0 (commented):
Solid reference material — exactly the kind of doc that pays for itself the next time. Two things worth tightening so a contributor can actually act on the steps:

1. **Status taxonomy is under-specified.** Principle 1 only mentions `SKIPPED` vs `PASSED`, but the runner exposes five statuses (`SKIPPED`, `NOT_CONFIGURED`, `WARNING`, `FAILED`, `ERROR`) and existing checks split between them in ways this doc doesn't capture. Inline comment with the decision tree.

2. **"Adding a New Check" assumes you know the `CheckResult` shape.** Steps 3–5 say "add a check function" without showing one. Drop in a 5-line minimal example so a contributor doesn't have to grep around — that's the trial-and-error this doc is meant to prevent. Inline.

scripts/flightcheck/AGENTS.md
GrahamMcMynn: Fixed: Changed to `../auth.py` and added a note explaining the `sys.path` manipulation in `cli.py` (line 29) that makes `from auth import authenticate` work.

scripts/flightcheck/AGENTS.md
GrahamMcMynn: Fixed: Rewrote as a three-way split  Local YAML (authored config, checked by local_files.py)  Dataverse (server-side state not in local files like statecode, template configs)  Island Gateway (runtime state). Added explicit guidance: if the data exists in local YAML, check it there.

scripts/flightcheck/AGENTS.md
GrahamMcMynn: Fixed: Added a runner attributes table listing all available attributes (env_url, dv_token, env_id, graph, pp_admin, pva, config) with types and descriptions. Noted that any client may be None if auth failed.

scripts/flightcheck/AGENTS.md
GrahamMcMynn: Fixed: Added full status decision tree covering all six statuses (PASSED, FAILED, WARNING, SKIPPED, NOT_CONFIGURED, ERROR) with when to use each. Core principle is now: never return PASSED when you cannot actually validate.

scripts/flightcheck/AGENTS.md
GrahamMcMynn: Fixed: Added a minimal working example showing imports from runner, function signature, and CheckResult construction with all fields.

scripts/flightcheck/AGENTS.md
GrahamMcMynn: Fixed: Changed phrasing to "common kinds include" with a note that the API may return other kinds not listed. Also fixed componenttype values  topics are type 9 (not 1), variables are type 12 (not 68), confirmed via live Dataverse queries during CONFIG-014 implementation.

scripts/flightcheck/AGENTS.md
```diff
@@ -0,0 +1,202 @@
...
+attached as ad-hoc attributes in `cli.py`. Available attributes:
+
+| Attribute | Type | Description |
+|-----------|------|-------------|
+| `runner.env_url` | `str` | Dataverse environment URL |
+| `runner.dv_token` | `str` | Dataverse bearer token |
+| `runner.env_id` | `str` | Power Platform (BAP) environment ID |
+| `runner.graph` | `GraphClient \| None` | Microsoft Graph client |
+| `runner.pp_admin` | `PowerPlatformAdminClient \| None` | BAP admin client |
+| `runner.pva` | `PVAClient \| None` | Island Gateway client |
```
johnguy0, line 150: `runner.pva` is not wired today: `cli.py` only attaches `graph` and `pp_admin`, and this branch has no `pva_client.py`. Remove this row or add the client and runner wiring before using it in the guide.

johnguy0 (commented):
I re-reviewed the delta since 91e47848. The earlier feedback is addressed. One new docs accuracy issue remains in the runner attributes table because it documents a client that is not present or wired. Not ready yet.

scripts/flightcheck/AGENTS.md
saengland: @GrahamMcMynn - confirmed: this branch has no `pva_client.py` (it lands in PR #37) and `cli.py` here only wires `runner.graph` + `runner.pp_admin`. There's a second doc-accuracy bug in the same table while we're here: `runner.pp_admin` is documented as `PowerPlatformAdminClient`, but the actual class in `pp_admin_client.py` is `PPAdminClient`.

Suggested fix:

```diff
 | `runner.graph` | `GraphClient \| None` | Microsoft Graph client |
-| `runner.pp_admin` | `PowerPlatformAdminClient \| None` | BAP admin client |
-| `runner.pva` | `PVAClient \| None` | Island Gateway client |
+| `runner.pp_admin` | `PPAdminClient \| None` | BAP admin client |
 | `runner.config` | `dict` | Parsed `my/config.json` |
```

And the matching reference at line 126:

```diff
-   same structure as `graph_client.py` / `pp_admin_client.py` / `pva_client.py`
+   same structure as `graph_client.py` / `pp_admin_client.py`
```

Add `runner.pva` and the `pva_client.py` reference back when PR #37 merges (or in the same merge if they go in together).

GrahamMcMynn: I will not address this comment.  The issue is I need to merge #37 first.

scripts/flightcheck/AGENTS.md
```diff
@@ -0,0 +1,202 @@
...
+
+FlightCheck validates customer deployments by querying multiple APIs. Each API
+requires its own authentication and has different data available:
+
+| API Layer | Client | Auth Resource / Scope | What's Available |
+|-----------|--------|----------------------|------------------|
+| Dataverse | `../auth.py` | `{env_url}/user_impersonation` | Bot components (topics, variables, knowledge source *config*), template configs, solution metadata, statecode (enabled/disabled) |
+| Microsoft Graph | `graph_client.py` | `https://graph.microsoft.com/.default` | Licenses, user roles, Entra app registrations, CA policies |
+| Power Platform Admin (BAP) | `pp_admin_client.py` | `https://service.powerapps.com//.default` | Environments, cloud flows, connections, DLP policies |
+| Island Gateway (Copilot Studio) | `pva_client.py` | `96ff4394-9197-43aa-b393-6a41652e21f8/.default` | Live bot component status, model config, knowledge source *runtime state* |
```
saengland, line 16: **HIGH** - Architecture table documents `pva_client.py` (Island Gateway client) as if it exists, but the file is not on `main` or in this PR's diff - it lives only on PR #37 (`users/grahamc/260504UpdateFlightCheckerForNewChecks`). A future check author following this guide before PR #37 merges will look for a file that isn't there. The same dangling reference also appears at lines 63-64 (`See pva_client.py _discover_gateway() for the implementation pattern`) and line 126 (lists `pva_client.py` as one of the existing client patterns to follow).

Fix: either (a) make PR #38 dependent on PR #37 and rebase it onto pr-37 so `pva_client.py` lands first, or (b) drop the Island Gateway row from the table and remove every `pva_client.py` / `_discover_gateway` reference (lines 16, 41-81, 63-64, 126) and add it back in a follow-up PR after PR #37 merges. Until then the only valid clients to document are `auth.py`, `graph_client.py`, and `pp_admin_client.py`.

scripts/flightcheck/AGENTS.md
saengland, line 150: **HIGH** - Runner attributes table claims `runner.pva: PVAClient | None` is available to check functions, but `cli.py` (lines 162-167 on main) only sets `runner.config`, `runner.env_url`, `runner.dv_token`, `runner.env_id`, `runner.graph`, `runner.pp_admin`. There is no `runner.pva`. The example at line 163 (`# ... possibly using runner.graph / runner.pva / etc. ...`) propagates the same error.

A check author who writes `if not runner.pva:` will hit `AttributeError`, not the documented `None` sentinel.

Fix: remove the `runner.pva` row from the table and the `runner.pva` mention from the example comment until the attribute is actually attached in cli.py (which happens in PR #37). After PR #37 merges, also document that `tenant_id` is implicitly needed by any new client following the existing MSAL pattern but is currently not exposed as a runner attribute - so authors must accept it as a separate arg or compute it from `env_url`.

scripts/flightcheck/AGENTS.md
```diff
@@ -0,0 +1,202 @@
...
+- **Filter by type:** `componenttype eq {N}` where common types are:
+  - `9` = Topic/Dialog
+  - `12` = Global variable
+  - `16` = Knowledge source
+- **The `data` column is YAML** (the `.mcs.yml` file content), NOT JSON. Do
+  not try to `json.loads()` it. Parse with a YAML library if needed.
+- **`statecode`/`statuscode`** are standard Dataverse record status (Active=0/1),
+  NOT the runtime crawl/index status.
+
+## Design Principles
```
saengland, line 99: **HIGH** - The Design Principles section (lines 99-141) has zero security guidance for the exact bug classes that landed in PR #29 round 8 (commits `97e9708`, `cb22c88`):

1. `xml.etree` instead of `defusedxml` for any XML parsing -> XXE
2. No `urllib3.util.retry.Retry` adapter on `requests.Session` for transient 5xx/429 against BAP/Graph -> intermittent flake
3. Raw exception text echoed into `CheckResult.result` -> CWE-209 leak of bearer tokens / SAS URLs / PII when an upstream API fails
4. No scheme validation on operator-provided URLs (e.g., `dataverseEndpoint` from `my/config.json`) -> http:// or file:// could be accepted

Future check authors will repeat these because the guide doesn't warn them. This guide is the first thing the next contributor reads.

Fix: add a `## Security Requirements` section after Design Principles with four bullets:

1. **XML parsing**: Use `defusedxml.ElementTree` not `xml.etree` for any XML payload (Dataverse SOAP, federation metadata). Catch both `ET.ParseError` AND `defusedxml.common.DefusedXmlException` so attack-path payloads (`EntitiesForbidden`, `ExternalReferenceForbidden`, `DTDForbidden`, `NotSupportedError`) fall through to a structured result instead of an unhandled traceback.
2. **HTTP retry**: Mount a `urllib3.util.retry.Retry(total=3, backoff_factor=1, status_forcelist=[429,500,502,503,504], allowed_methods=GET/HEAD/OPTIONS only, respect_retry_after_header=True)` adapter on a module-level `requests.Session`. Do NOT include POST/PUT/DELETE in `allowed_methods` - replays of mutating verbs on transient 5xx are unsafe.
3. **CWE-209 scrubbing**: Never put raw exception text or `error_description` into `CheckResult.result` or `CheckResult.remediation`. Bearer tokens, SAS URLs, and tenant IDs leak through `requests.HTTPError.response.text` and MSAL error_description. Echo only the `error` code and HTTP status; log full detail to stderr only.
4. **URL validation**: When accepting a URL from `my/config.json` or any operator-provided source, validate `urlparse(url).scheme == "https"` before sending bearer tokens to it. See `auth._validate_https_url` and `pp_admin_client.derive_environment_id` for the pattern.

scripts/flightcheck/AGENTS.md
```diff
@@ -0,0 +1,202 @@
...
+
+1. Identify which API has the data you need. Check existing clients first (see
+   table above). If the data isn't available from any existing client, research
+   the correct API — check the `microsoft/MCS-Agent-Builder` repo's
+   `knowledge/cache/` folder for API documentation, or inspect network traffic
+   in the relevant admin UI (Copilot Studio, Power Platform Admin Center, etc.)
+2. If a new client is needed, create it following existing patterns and add it
+   to the table above
+3. Add the check function to the appropriate module in `checks/`
+4. Wire it into `_check_single_agent()` (for per-agent checks) or the
```
saengland, line 186: **MEDIUM** - Checklist step 4 says "Wire it into `_check_single_agent()` (for per-agent checks) or the appropriate scope runner" but never tells the author HOW to wire a brand-new scope. The actual mechanism is: import the function in `cli.py`, add a tuple `(category_name, function)` to `FULL_SCOPE`, and add an entry to `SCOPE_MAP` if the check should also be runnable in isolation via `--scope <name>`. Without that, the `runner.register(...)` call never happens for the new check and it silently never executes.

Fix: replace step 4 with two explicit steps:

```
4. For per-agent checks (each agent under `my/agents/`), call your check from `_check_single_agent()` in `checks/local_files.py`.
5. For a new scope category, in `scripts/flightcheck/cli.py`:
   a. Import the check: `from flightcheck.checks.<module> import run_<name>_checks`
   b. Append `("<Category Name>", run_<name>_checks)` to `FULL_SCOPE`.
   c. If the scope should also run in isolation, add `"<scope_name>": [("<Category Name>", run_<name>_checks)]` to `SCOPE_MAP`.
   d. The runner picks it up via `runner.register(category, fn)`; no further wiring needed.
```

scripts/flightcheck/AGENTS.md
```diff
@@ -0,0 +1,202 @@
+# FlightCheck — Developer Guide
```
saengland, line 1: **LOW** - Document uses Unicode em-dash (U+2014) in 12 places: lines 1, 22, 33, 87, 105, 107, 108, 109, 111, 113, 115, 180. Rest of the kit's docs and source comments use ASCII hyphen. Mixed usage is a churn signal in `git grep` and breaks the kit's ASCII-only convention (Outlook COM mojibake history - same root cause when authors copy from rendered markdown into other surfaces).

Fix: replace every ` — ` with ` - ` (or ` -- ` if you want to preserve the visual weight). PowerShell from repo root:

```powershell
(Get-Content scripts/flightcheck/AGENTS.md -Raw).Replace([char]0x2014, '-') | Set-Content scripts/flightcheck/AGENTS.md -NoNewline -Encoding utf8
```

saengland (changes requested):
Reviewed AGENTS.md as the contract for future check authors. The structure (architecture, runner attributes, design principles, checklist) is right, but it documents the world that exists in PR #37, not the world that exists on `main`, and the security section has a critical gap. 5 findings:

| # | Sev | Issue |
|---|---|---|
| 1 | HIGH | `pva_client.py` referenced as if it exists on main; only lives in PR #37 |
| 2 | HIGH | `runner.pva: PVAClient \| None` documented but cli.py never sets it |
| 3 | HIGH | Zero security guidance for the bug classes round-8 of PR #29 just fixed (defusedxml, retry adapter, CWE-209 scrubbing, HTTPS validation on operator URLs) |
| 4 | MEDIUM | Checklist step 4 doesn''t actually tell authors how to wire a new scope (FULL_SCOPE / SCOPE_MAP in cli.py) |
| 5 | LOW | 12 em-dash chars throughout - kit convention is ASCII only |

#1, #2, #3 are blockers. The security section gap is the biggest one - this guide will be the first thing the next check author reads, and it should encode the lessons we just paid for.

saengland:

Action needed on your side:

saengland:
Reopened as #58 against main. The previous review (this PR) is preserved here for context.

PR #40, opened 2026-05-05
FlightCheck: Add CONFIG-014 topic description quality check

### LINKED ISSUE #26: FlightCheck: Add check for topic description quality and specificity

## What should be checked

The FlightCheck tool should validate that topic descriptions are sufficiently specific for accurate knowledge retrieval routing. Vague or empty topic descriptions cause the agent to route queries incorrectly, producing inconsistent responses.

## Suggested implementation

- **Checkpoint ID:** CONFIG-014
- **Category:** Local Files
- **Priority:** MEDIUM
- **Validation logic:** Parse each topic's description field and verify it contains at least 20 words with relevant domain keywords. Flag topics with empty, generic, or overly short descriptions that may cause routing ambiguity.

## Context

GrahamMcMynn:
## Summary

Adds CONFIG-014 validation to FlightCheck that checks topic description quality for AI routing accuracy.

### What it checks
- **Placeholder detection** (FAILED): Flags topics with placeholder text like \[add keywords here]\ that were never filled in
- **Word count** (WARNING): Flags descriptions under 20 words that may be too short for accurate AI orchestrator routing

### Key design decisions
- **Skips disabled topics**: Queries Dataverse (\componenttype eq 9, statecode eq 1\) so disabled topics don't produce false errors
- **Skips system topics**: Topics like \on-error\, \conversation-start\, \
eset-conversation\ don't use AI routing
- **Uses display names**: Shows the Copilot Studio topic name (from \modelDisplayName\) in output so users can find the topic easily

### Testing
Tested against live tenant with \python scripts/flightcheck/cli.py --scope local\. Correctly:
- Skipped 5 disabled topics (including the placeholder \Agent handoff - [scenario name]\)
- Flagged 2 topics with short descriptions as warnings

Closes #26


scripts/flightcheck/checks/local_files.py
```diff
@@ -269,6 +272,199 @@ def _check_variables(agent_path: Path, label: str) -> list[CheckResult]:
...
+            filter_expr=f"_parentbotid_value eq '{bot_id}' and componenttype eq 9 and statecode eq 1",
+        )
+        # Convert Dataverse names to local filenames for matching
+        disabled = set()
+        for c in components:
+            name = c.get("name", "")
+            if name:
+                disabled.add(_friendly_filename(name))
+        return disabled
+    except Exception:
```
johnguy0, line 347: Do not swallow this exception and continue as if no topics are disabled. That turns Dataverse/auth/config failures into false CONFIG-014 failures for disabled topics, and it hides the real fault. Return the error to _check_topic_descriptions as a FAILED/SKIPPED result with context, or let it fail loudly.

scripts/flightcheck/checks/local_files.py
```diff
@@ -269,6 +272,199 @@ def _check_variables(agent_path: Path, label: str) -> list[CheckResult]:
...
+        stem = tf.stem.replace(".mcs", "").lower()
+        if stem in disabled_topics:
+            continue
+
+        content = tf.read_text(encoding="utf-8", errors="replace")
+
+        # Extract modelDescription (multiline block scalar or inline)
+        # The pattern allows blank lines within the indented block
+        desc_match = re.search(
+            r'modelDescription:\s*\|[\-\+]?\s*\n((?:(?:[ \t]+.+|[ \t]*)\n?)+)', content
```
johnguy0, line 389: This only handles literal block scalars. A valid folded scalar, `modelDescription: >`, falls through to the inline regex and gets counted as the single token `>`. Accept both `|` and `>`, or parse the YAML field instead of regexing it.

scripts/flightcheck/checks/local_files.py
```diff
@@ -269,6 +272,199 @@ def _check_variables(agent_path: Path, label: str) -> list[CheckResult]:
...
+    checked = 0
+
+    # Query Dataverse for disabled topics to skip them
+    disabled_topics = _get_disabled_topic_names(runner, agent_name)
+
+    for tf in sorted(topics_dir.glob("*.mcs.yml")):
+        filename = tf.name.lower()
+
+        # Skip system topics that don't use AI-based routing
+        if any(re.search(pat, filename) for pat in _SYSTEM_TOPIC_PATTERNS):
```
johnguy0, line 376: This substring match can skip custom topics that happen to include these words in the filename, for example a user topic named `on-error-followup`. Match known system topic stems exactly, or inspect a stable topic/schema marker before skipping.

scripts/flightcheck/checks/local_files.py
```diff
@@ -269,6 +272,199 @@ def _check_variables(agent_path: Path, label: str) -> list[CheckResult]:
...
+    r"response-preparation",
+]
+
+# Placeholder phrases that indicate the description was never filled in
+_PLACEHOLDER_PATTERNS = [
+    r"\[add\s+keywords",
+    r"\[add\s+.*here\]",
+    r"\[describe",
+    r"\[placeholder",
+    r"\btodo\b",
```
johnguy0, line 294: Do not fail on bare `todo` or `placeholder` as standalone words. Those can be legitimate routing terms for task-tracker or template-help agents. Limit FAILED placeholders to bracketed template text or exact phrases that indicate an unfilled description.

scripts/flightcheck/checks/local_files.py
```diff
@@ -269,6 +272,199 @@ def _check_variables(agent_path: Path, label: str) -> list[CheckResult]:
...
+        for c in components:
+            name = c.get("name", "")
+            if name:
+                disabled.add(_friendly_filename(name))
+        return disabled
+    except Exception:
+        return set()
+
+
+def _check_topic_descriptions(agent_path: Path, label: str, runner=None, agent_name: str = "") -> list[CheckResult]:
```
johnguy0, line 351: Add focused tests for this function before merging. Cover folded scalars, inline descriptions, placeholder false positives, disabled-topic query failures, and system-topic skip behavior. Live-tenant testing will not catch these parser edge cases.

johnguy0 (commented):
Do not merge this as-is. CONFIG-014 is useful, but the disabled-topic path currently collapses Dataverse and config failures into 'no disabled topics', which can create false failures and hide integration problems. Fix that path and add focused tests around the parser and skip rules.

scripts/flightcheck/checks/local_files.py
saengland: @GrahamMcMynn - the right move here is to drop the regex altogether and use PyYAML (already a hard dependency for the kit). It handles `|`, `>`, `|-`, `>+`, inline, multiline, and quoted forms for free, and it eliminates the next batch of regex bugs people will find:

```python
import yaml

def _extract_model_description(content: str) -> str | None:
    """Return modelDescription from a topic .mcs.yml, or None if absent.

    Handles literal block (`|`), folded block (`>`), block-chomping
    indicators (`|-`, `>+`, `|2+` etc.), inline strings, and single/
    double-quoted forms - all via PyYAML's parser.
    """
    try:
        doc = yaml.safe_load(content)
    except yaml.YAMLError:
        return None
    if not isinstance(doc, dict):
        return None
    desc = doc.get("modelDescription")
    if isinstance(desc, str):
        return desc.strip()
    return None
```

Same swap for `modelDisplayName` - parse once at the top of the loop instead of regexing both fields:

```python
try:
    doc = yaml.safe_load(content) or {}
except yaml.YAMLError as e:
    # Malformed YAML is its own failure - count it as a topic-quality
    # issue rather than skipping silently.
    has_placeholder.append(f"{tf.name} (could not parse YAML: {e})")
    continue
desc_text = (doc.get("modelDescription") or "").strip()
display_name = (doc.get("modelDisplayName") or
                f"{tf.stem.replace('.mcs', '').replace('-', ' ').title()} (file: {tf.name})")
if not desc_text:
    # No description -> topic uses triggerQueries only; not a CONFIG-014 case.
    continue
```

This is fewer lines, kills the folded-scalar bug, kills the "regex eats inline `|`/`>` if it ever appears in a quoted string" class of bugs, and gives you display-name parsing with the same robustness.

scripts/flightcheck/checks/local_files.py
saengland: @GrahamMcMynn - here's a starter test file that covers all five edge cases John called out. Drop in `scripts/flightcheck/tests/test_check_topic_descriptions.py`:

```python
"""Unit tests for CONFIG-014 topic description quality check."""
from pathlib import Path
import textwrap
import pytest

from flightcheck.checks.local_files import (
    _check_topic_descriptions,
    _extract_model_description,
    _SYSTEM_TOPIC_STEMS,
    _PLACEHOLDER_PATTERNS,
)


def _write(tmp_path: Path, name: str, body: str) -> Path:
    topics = tmp_path / "topics"
    topics.mkdir(exist_ok=True)
    (topics / name).write_text(textwrap.dedent(body), encoding="utf-8")
    return tmp_path


# --- _extract_model_description: parser shape ---

def test_extracts_literal_block():
    body = "modelDescription: |\n  Hello world over multiple words\n"
    assert _extract_model_description(body) == "Hello world over multiple words"


def test_extracts_folded_block():
    body = "modelDescription: >\n  Folded scalar that should join into one line\n"
    assert _extract_model_description(body).startswith("Folded scalar")


def test_extracts_inline_string():
    body = "modelDescription: A short inline description\n"
    assert _extract_model_description(body) == "A short inline description"


def test_extracts_quoted_inline():
    body = 'modelDescription: "quoted: value with colon"\n'
    assert _extract_model_description(body) == "quoted: value with colon"


def test_returns_none_when_absent():
    assert _extract_model_description("modelDisplayName: Foo\n") is None


# --- _check_topic_descriptions: skip rules ---

def test_skips_system_topic_by_exact_stem(tmp_path):
    _write(tmp_path, "on-error.mcs.yml",
           "modelDescription: too short\n")
    out = _check_topic_descriptions(tmp_path, "test")
    # System topic should be skipped, not flagged as too-short.
    assert all("on-error.mcs.yml" not in r.description for r in out)


def test_does_not_skip_user_topic_with_system_prefix(tmp_path):
    # Regression test for John's substring-match finding.
    _write(tmp_path, "on-error-followup.mcs.yml",
           "modelDescription: too short\n")
    out = _check_topic_descriptions(tmp_path, "test")
    assert any("on-error-followup" in r.description for r in out), \
        "user topic with system-stem prefix must not be silently skipped"


# --- placeholder detection ---

def test_bracketed_placeholder_is_failure(tmp_path):
    _write(tmp_path, "good.mcs.yml",
           'modelDescription: |\n  [add keywords here describing this topic]\n')
    out = _check_topic_descriptions(tmp_path, "test")
    assert any("placeholder" in r.result.lower() for r in out)


def test_bare_word_todo_is_not_failure(tmp_path):
    # Regression test for John's "task-tracker agent" finding.
    _write(tmp_path, "create-todo.mcs.yml",
           "modelDescription: |\n"
           "  Create a new todo item in the user's task list with optional due date and priority assignment.\n")
    out = _check_topic_descriptions(tmp_path, "test")
    assert all("placeholder" not in r.result.lower() for r in out), \
        "legitimate description that contains the word 'todo' must not be flagged as placeholder"


# --- disabled-topic lookup failure mode ---

class _StubRunner:
    def __init__(self):
        self.dv_token = None  # forces _get_disabled_topics to short-circuit
        self.env_url = None


def test_dataverse_failure_skips_check_with_reason(tmp_path):
    # Regression test for John's "swallow exception" finding.
    _write(tmp_path, "real-topic.mcs.yml",
           'modelDescription: short\n')
    out = _check_topic_descriptions(tmp_path, "test", runner=_StubRunner())
    skipped = [r for r in out if r.status == "SKIPPED"]
    assert skipped, "missing Dataverse connection must produce a SKIPPED result, not silently fail open"
    assert "Dataverse" in skipped[0].result or "config" in skipped[0].result
```

Wire it into CI by adding `pytest` + `pyyaml` (already there) to `requirements-dev.txt` and a `pytest scripts/flightcheck/tests/` job in `.github/workflows/ci.yml`. If you'd rather keep tests out of `scripts/`, move to `tests/flightcheck/` at the repo root and adjust the import.

scripts/flightcheck/checks/local_files.py
saengland: @GrahamMcMynn - the swallow-and-return-set() pattern is the root cause of the false-failure mode John flagged in his summary. Make the function return a result instead of just the set so the caller can distinguish "queried successfully, no disabled topics" from "couldn't query at all":

```python
@dataclass
class _DisabledTopicsLookup:
    topics: set[str]            # filename stems of disabled topics
    queried: bool               # True if we successfully reached Dataverse
    skip_reason: str = ""       # populated when queried is False


def _get_disabled_topics(runner, agent_name: str) -> _DisabledTopicsLookup:
    if not runner or not getattr(runner, 'dv_token', None) or not getattr(runner, 'env_url', None):
        return _DisabledTopicsLookup(set(), queried=False,
            skip_reason="no Dataverse connection on runner")

    try:
        with open("my/config.json") as f:
            config = json.load(f)
    except FileNotFoundError:
        return _DisabledTopicsLookup(set(), queried=False,
            skip_reason="my/config.json not found - run /setup first")
        return _DisabledTopicsLookup(set(), queried=False,
            skip_reason=f"my/config.json is not valid JSON: {e}")

    bot_id = next(
        (a.get("botId") for a in config.get("agents", []) if a.get("slug") == agent_name),
        None,
    )
    if not bot_id:
        return _DisabledTopicsLookup(set(), queried=False,
            skip_reason=f"no botId for agent slug '{agent_name}' in my/config.json")

    try:
        from auth import query_all
        components = query_all(
            runner.env_url, runner.dv_token, "botcomponents",
            "name,schemaname,statecode",
            filter_expr=f"_parentbotid_value eq '{bot_id}' and componenttype eq 9 and statecode eq 1",
        )
    except Exception as e:
        return _DisabledTopicsLookup(set(), queried=False,
            skip_reason=f"Dataverse query failed: {e}")

    disabled = {_friendly_filename(c.get("name", "")) for c in components if c.get("name")}
    return _DisabledTopicsLookup(disabled, queried=True)
```

Then in `_check_topic_descriptions`, surface a SKIPPED result (not a silent set()) when the lookup fails:

```python
disabled_lookup = _get_disabled_topics(runner, agent_name)
if not disabled_lookup.queried:
    results.append(CheckResult(
        checkpoint_id="CONFIG-014", category=f"Topics ({label})",
        priority=Priority.MEDIUM.value, status=Status.SKIPPED.value,
        description=f"{label}: Topic description quality",
        result=f"Cannot reliably check disabled topics: {disabled_lookup.skip_reason}",
        remediation="Re-run after fixing the upstream issue, or pass --skip-disabled-topic-check to ignore.",
    ))
    return results
disabled_topics = disabled_lookup.topics
```

This both kills the false-failure mode (we no longer report disabled topics as having short descriptions when we couldn't reach Dataverse) and surfaces the real error.

scripts/flightcheck/checks/local_files.py
saengland: @GrahamMcMynn - swap the patterns from substrings to exact-stem matches against the file basename. Then `on-error-followup.mcs.yml` no longer collides with the `on-error` system topic:

```python
# Exact filename stems for system topics that don't use AI-based routing.
# Match against the .mcs.yml stem (filename minus extension), not via
# substring search, so user topics like `on-error-followup.mcs.yml`
# aren't silently skipped.
_SYSTEM_TOPIC_STEMS = frozenset({
    "conversation-start",
    "on-error",
    "reset-conversation",
    "log-telemetry",
    "microsoft-self-help",
    "response-preparation",
})

# In the loop:
stem = tf.stem.removesuffix(".mcs").lower()  # Path('reset-conversation.mcs.yml').stem == 'reset-conversation.mcs'
if stem in _SYSTEM_TOPIC_STEMS:
    continue
```

(`removesuffix` is 3.9+, which the kit already requires per `requirements.txt`.)

If you want to be even more defensive (i.e. trust a topic-schema marker over a filename convention), the canonical signal is `botMode: System` or `componenttype` flagged at extract time. Filename match is the cheap-and-correct first cut though, and matches what `extract.py` already names them.

Bonus: drop `_SYSTEM_TOPIC_PATTERNS` entirely once the new set is in - the regex-list approach was a leftover from when stems weren't normalized.

scripts/flightcheck/checks/local_files.py
saengland: @GrahamMcMynn - the bare-word `\btodo\b` and `\bplaceholder\b` patterns are the false-positive risk. A task-tracker agent topic legitimately named "Create a Todo" or a help topic that says "this is a placeholder topic for the agent's name" both FAIL today. Tighten to bracketed template text or unmistakable unfilled-stub phrases:

```python
# Patterns that indicate a description was never actually filled in.
# Conservative on purpose: better to miss one stub than to fail-flag a
# legitimate description that happens to mention "todo" or "placeholder"
# as content (task-tracker agents, template-help agents).
_PLACEHOLDER_PATTERNS = [
    # Bracketed template text the maker forgot to replace
    r"\[add\s+keywords?",
    r"\[add\s+.*here\]",
    r"\[describe[^\]]*\]",
    r"\[placeholder[^\]]*\]",
    r"\[todo[^\]]*\]",
    r"\[fill\s+in[^\]]*\]",
    r"\[your\s+description\s+here\]",
    r"<\s*describe[^>]*>",
    r"<\s*placeholder[^>]*>",
    # Unmistakable stub phrases
    r"\bdescribe\s+this\s+topic\b",
    r"\badd\s+your\s+description\b",
    r"\bTODO:\s+(write|add|fill)",
    r"\bplaceholder\s+description\b",
]
```

Drop bare `\btodo\b` and `\bplaceholder\b` from the FAILED list. If you still want a softer signal for those, surface them as WARNING with the matched fragment so a maker can eyeball whether their topic is actually about todos or just hasn't been filled in:

```python
_SUSPICIOUS_BARE_WORDS = re.compile(r"\b(todo|placeholder)\b", re.IGNORECASE)
```

Then in the loop, `_PLACEHOLDER_PATTERNS` -> FAILED, `_SUSPICIOUS_BARE_WORDS` -> WARNING with the matched word and surrounding context.

scripts/flightcheck/checks/local_files.py
```diff
@@ -269,6 +272,199 @@ def _check_variables(agent_path: Path, label: str) -> list[CheckResult]:
...
+
+        # Extract modelDescription (multiline block scalar or inline)
+        # The pattern allows blank lines within the indented block
+        desc_match = re.search(
+            r'modelDescription:\s*\|[\-\+]?\s*\n((?:(?:[ \t]+.+|[ \t]*)\n?)+)', content
+        )
+        if desc_match:
+            desc_text = desc_match.group(1).strip()
+        else:
+            inline_match = re.search(r'modelDescription:\s*(?!\|)(.+)', content)
```
saengland, line 394: **HIGH** - When `modelDescription:` has no value on the line (empty field), the inline-fallback regex `r'modelDescription:\s*(?!\|)(.+)'` allows `\s*` to consume the trailing newline and then captures the NEXT YAML field's content as the description. Verified: `"modelDescription:\ntriggerQueries:\n"` returns `"triggerQueries:"`.

This produces misleading word counts in the `too_short` report and can silently false-NEGATIVE if the next field's value is long enough to clear the 20-word threshold. The check will appear to pass on topics that have no description at all.

Fix: constrain `\s*` to same-line whitespace and treat empty values as missing:

```python
inline_match = re.search(r"modelDescription:[ \t]*(?!\|)([^\n]*)", content)
if inline_match:
    desc_text = inline_match.group(1).strip()
    if not desc_text:
        too_short.append(f"{display_name} (0 words)")
        continue
```

Or after extraction, treat `desc_text` as missing if it is empty OR matches `r"^[A-Za-z_][A-Za-z0-9_]*\s*:\s*$"` (looks like a bare YAML key) and report it as 0-word/missing rather than feeding it to the word-count path.

scripts/flightcheck/checks/local_files.py
```diff
@@ -269,6 +272,199 @@ def _check_variables(agent_path: Path, label: str) -> list[CheckResult]:
...
+
+
+def _friendly_filename(name: str) -> str:
+    """Convert a Dataverse topic name to the local filename stem (matching extract.py logic)."""
+    raw = re.sub(r"^\[.*?\]\s*[-\u2013\u2014]\s*", "", name)
+    slug = raw.strip().replace("_", "-")
+    slug = re.sub(r"[^a-zA-Z0-9\s-]", "", slug)
+    slug = re.sub(r"[\s]+", "-", slug)
+    slug = re.sub(r"-+", "-", slug).strip("-").lower()
+    return slug
```
saengland, line 308: **MEDIUM** - `_friendly_filename` drops the `schemaname` fallback that `scripts/extract.py:friendly_filename` uses (`return slug or schemaname.rsplit(".", 1)[-1].lower()`). Any disabled Dataverse topic with an empty/whitespace `name` field will produce an empty slug here, never match an on-disk file, and therefore never be skipped - so disabled topics get flagged for placeholder/short descriptions even though the Dataverse query was successful.

The docstring claims this matches `extract.py` logic, but it only partially does.

Fix: take both fields and mirror `extract.py` exactly:

```python
def _friendly_filename(name: str, schemaname: str = "") -> str:
    raw = name if name else schemaname.rsplit(".", 1)[-1] if schemaname else ""
    raw = re.sub(r"^\[.*?\]\s*[-\u2013\u2014]\s*", "", raw)
    slug = raw.strip().replace("_", "-")
    slug = re.sub(r"[^a-zA-Z0-9\s-]", "", slug)
    slug = re.sub(r"[\s]+", "-", slug)
    slug = re.sub(r"-+", "-", slug).strip("-").lower()
    return slug or (schemaname.rsplit(".", 1)[-1].lower() if schemaname else "")
```

Then at line 345 pass schemaname through: `disabled.add(_friendly_filename(name, c.get("schemaname", "")))` and drop the `if name:` guard so nameless rows still get a slug from schemaname.

scripts/flightcheck/checks/local_files.py
```diff
@@ -269,6 +272,199 @@ def _check_variables(agent_path: Path, label: str) -> list[CheckResult]:
...
+
+def _get_disabled_topic_names(runner, agent_name: str) -> set[str]:
+    """Query Dataverse for disabled topics and return their local filenames (lowercased)."""
+    if not runner or not getattr(runner, 'dv_token', None) or not getattr(runner, 'env_url', None):
+        return set()
+
+    # Find botId from config
+    import json
+    try:
+        config = json.load(open("my/config.json"))
```
saengland, line 319: **MEDIUM** - Three issues in one line:

1. `json.load(open("my/config.json"))` leaks the file handle (no `with`).
2. Path is cwd-relative - breaks if FlightCheck is invoked from anywhere other than the repo root.
3. Re-reads config that `cli.py` has already loaded onto the runner as `runner.config`. The rest of the runner consistently uses `runner.config` / `runner.env_url` / `runner.dv_token`.

Fix: use the already-loaded config off the runner:

```python
config = getattr(runner, "config", None)
if not config:
    return set()
```

Then drop the `import json` and the `try/except (FileNotFoundError, json.JSONDecodeError)` block entirely. Also makes the function unit-testable without monkeypatching the filesystem.

scripts/flightcheck/checks/local_files.py
saengland, line 347: **MEDIUM** - Bare `except Exception: return set()` swallows every real failure of the Dataverse query (auth-expired 401, network timeout, malformed response, schema drift on `botcomponents`). When this silently returns an empty set, `_check_topic_descriptions` proceeds to flag every disabled topic as a placeholder/short-description failure with no signal in the FlightCheck report explaining why the disabled-topic skip didn't apply.

Users will see false-positive CONFIG-014 failures and have no clue the Dataverse step quietly broke. Worse, programming bugs (NameError, TypeError) also get swallowed.

Fix: either (a) let the exception propagate so the runner's outer try/except in `runner.py:run()` records a CHK-ERR result, OR (b) catch and emit a diagnostic by returning a sentinel and having the caller append a SKIPPED/WARNING CheckResult. At minimum, narrow the catch:

```python
except (requests.RequestException, KeyError, ValueError) as e:
    print(f"  WARN: CONFIG-014 disabled-topic lookup failed for {agent_name}: {e}")
    return set()
```

So programming bugs surface in the run.

scripts/flightcheck/checks/local_files.py
```diff
@@ -269,6 +272,199 @@ def _check_variables(agent_path: Path, label: str) -> list[CheckResult]:
...
+    return slug
+
+
+def _get_disabled_topic_names(runner, agent_name: str) -> set[str]:
+    """Query Dataverse for disabled topics and return their local filenames (lowercased)."""
+    if not runner or not getattr(runner, 'dv_token', None) or not getattr(runner, 'env_url', None):
+        return set()
+
+    # Find botId from config
+    import json
```
saengland, line 317: **LOW** - `import json` (line 317) and `from auth import query_all` (line 332) are placed inside the function body. The rest of the FlightCheck modules put all imports at the top (e.g. `local_files.py` lines 11-15, `cli.py`, `runner.py`). Inline imports here add a tiny per-call cost but more importantly hide the `auth` dependency from anyone reading the module header.

Fix: move both to the top of the file with the existing imports:

```python
import json
import os
import re
from pathlib import Path

from ..runner import CheckResult, Status, Priority

try:
    from auth import query_all  # scripts/auth.py - on sys.path via cli.py bootstrap
except ImportError:
    query_all = None
```

Then guard `_get_disabled_topic_names` with `if query_all is None: return set()`.

scripts/flightcheck/checks/local_files.py
```diff
@@ -269,6 +272,199 @@ def _check_variables(agent_path: Path, label: str) -> list[CheckResult]:
...
+            if inline_match:
+                desc_text = inline_match.group(1).strip()
+            else:
+                # No modelDescription — topic may use triggerQueries only
+                continue
+
+        checked += 1
+
+        # Get the display name shown in Copilot Studio
+        name_match = re.search(r'modelDisplayName:\s*(.+)', content)
```
saengland, line 404: **LOW** - `r"modelDisplayName:\s*(.+)"` has the same missing `(?!\|)` guard as the description regex. If a topic ever uses a block-scalar display name (`modelDisplayName: |\n  Foo Bar`), the captured `display_name` becomes `"|"` (then `"|"` after strip), producing a confusing user-facing message like `"| (3 words)"`.

Cosmetic only - the check still fires on the right topic - but worth aligning with the description regex.

Fix: tighten to single-line, non-block:

```python
r"modelDisplayName:[ \t]*(?!\|)([^\n]+)"
```

If no match, fall back to the filename-derived label (the existing else branch). Also incidentally handles `modelDisplayName:` with no value by falling through to the filename label.

saengland (changes requested):
Reviewed CONFIG-014. The premise (heuristic on description length + placeholder detection) is right and slots into the kit cleanly. Six findings, one is a real bug:

| # | Sev | Issue |
|---|---|---|
| 1 | HIGH | `modelDescription:` regex eats the next YAML field on empty value - silent false-NEGATIVE / misleading word counts |
| 2 | MEDIUM | `_friendly_filename` drops `extract.py`''s schemaname fallback - disabled topics with empty `name` field get false-flagged |
| 3 | MEDIUM | `json.load(open("my/config.json"))` leaks file handle, cwd-relative, duplicates already-loaded `runner.config` |
| 4 | MEDIUM | Bare `except Exception: return set()` swallows auth-expired / transport / programming bugs - silent false-positives |
| 5 | LOW | Inline imports (`import json`, `from auth import query_all`) instead of top-of-file |
| 6 | LOW | `modelDisplayName:` regex has same `(?!\|)` flaw - cosmetic |

#1 is the blocker - without it the check silently lies in production reports.

saengland:

Action needed on your side:

saengland:
Reopened as #59 against main. The previous review (this PR) is preserved here for context.

PR #52, opened 2026-05-07
PR #6: Connect + Onboarding skills

saengland:
Reopens #34. The original PR was auto-merged into `main2` on May 7 without reviewer approval (branch protection wasn't yet in place; auto-merge had nothing to wait for). The merge has been reverted on `main2` and this PR re-submits the same head commit (`d3c3e0b`) for proper review.

No code changes from #34 - same head SHA. CI will re-run.

PR #53, opened 2026-05-07
PR #7: Operational skills

saengland:
Reopens #35. The original PR was auto-merged into `main2` on May 7 without reviewer approval (branch protection wasn't yet in place; auto-merge had nothing to wait for). The merge has been reverted on `main2` and this PR re-submits the same head commit (`c1c240c`) for proper review.

No code changes from #35 - same head SHA. CI will re-run.


saengland:
@johnguy0 this is the reopen of #35 (PR #7). The original was reverted on `main2` (revert commit `f5fd630`); the auto-merge state on the original is stale.

For your agent's wait-for-prior rule: the `PR #6` -> `PR #7` -> ... -> `PR #10` series (#52 -> #53 -> #54 -> #55 -> #56) is the ordering. #52 is in re-review now; this PR comes next once #52 lands.

solutions/ess-maker-skills/src/skills/cleanup/SKILL.md
```diff
@@ -0,0 +1,159 @@
+# Cleanup Skill
+
+This skill scans a cloned ESS agent for compile errors and walks the user through fixing them.
+
+## Rules
+
+- Do NOT run terminal commands or scripts. Do NOT use Python, PowerShell, Node.js, or any other scripting language. Use your built-in file reading, writing, and editing tools only.
```
johnguy0: This rule conflicts with Step 2b (line 46) and Step 6 of this same skill, both of which tell the agent to run `python scripts/checkpoint.py` and `python scripts/push.py` in the terminal. Either remove the blanket no-terminal-commands rule or carve out an explicit allowlist for the checkpoint/push helpers. As written the agent will refuse to run the very steps it is being asked to perform.

solutions/ess-maker-skills/src/skills/topics/test/SKILL.md
```diff
@@ -0,0 +1,4 @@
+# Test Topic Skill
+
+<!-- TODO: How to test a topic in Copilot Studio test pane -->
+<!-- TODO: Common topic issues, trigger phrase conflicts, variable binding errors -->
```
johnguy0, line 14: This is a TODO stub. If `/test-topic` is wired into the menu, invoking it will land the agent on an empty skill. Same applies to `workflows/test/SKILL.md`. Either fill these in before merge or pull both stubs into a follow-up PR so the menu does not advertise commands that no-op.

solutions/ess-maker-skills/src/skills/topics/delete/SKILL.md
```diff
@@ -0,0 +1,131 @@
...
+
+```
+python scripts/push.py --yes
+```
+
+This authenticates to Dataverse, deletes the bot component record, and
+updates the local baseline.
+
+**If the push fails:**
+- Show the error output to the user.
```
johnguy0, line 104: `python scripts/checkpoint.py --revert` only works if `checkpoint.py` actually accepts a `--revert` flag. Worth confirming the flag exists and reverts to the most recent named checkpoint, otherwise this fallback path will fail right after a botched push, which is the moment it most needs to work.

johnguy0 (commented):
Took a pass through the operational skills set. Most of the SKILL.md files look solid, the cleanup/delete/update flows have the right shape (config read, checkpoint, dry-run, push, verify) and the rules call out the destructive-action guardrails clearly.

A few things to clean up before merge:

1. `cleanup/SKILL.md` contradicts itself, the Rules say no terminal commands but Step 2b/4 instruct the agent to run `python scripts/checkpoint.py` and `python scripts/push.py`. We need to either drop the no-terminal-commands rule or carve out an explicit allowlist for the checkpoint/push scripts. Right now the agent will hit conflicting instructions and stall.
2. `topics/test/SKILL.md` and `workflows/test/SKILL.md` are TODO stubs. If `/test-topic` or `/test-workflow` are wired into the menu, invoking either is going to land the agent on an empty skill. Either fill them in or pull them out of this PR and ship them in a follow-up.

Nothing else blocking. Inline comments below.

solutions/ess-maker-skills/src/skills/cleanup/SKILL.md
saengland: Fixed in `cf69dfe`. The rule on `cleanup/SKILL.md` line 7 now reads:

> Do NOT run terminal commands or scripts EXCEPT the kit's own helper scripts under `scripts/` (specifically `python scripts/checkpoint.py` and `python scripts/push.py`, which Steps 2b and 6 require).

So Step 2b's `python scripts/checkpoint.py "pre-scan-fixes"` and Step 6's `/push` no longer get refused.

solutions/ess-maker-skills/src/skills/topics/test/SKILL.md
saengland: Fixed in `cf69dfe`. Both `topics/test/SKILL.md` and `workflows/test/SKILL.md` now have an explicit "not yet implemented, here is what to do manually" message instead of being TODO-only stubs.

Topics test now tells the user to use the Copilot Studio test pane and walks through where to look for failed nodes. Workflows test tells the user to use Power Automate's Test -> Manually pane. Both files explicitly say "Then stop. Do NOT pretend to run a test or report results." so the agent doesn't fabricate results, and both keep the TODO comments for the future automated implementation.

Menu invocation now lands on a concrete user-facing instruction instead of an empty skill.

solutions/ess-maker-skills/src/skills/topics/delete/SKILL.md
saengland: Confirmed - the flag is real and works the way the SKILL describes:

`cmd_revert(agent_dir)` at `solutions/ess-maker-skills/scripts/checkpoint.py:131` (and listed in the usage docstring at line 12) does:

1. Auto-saves the current state to a fresh checkpoint first (so the revert is itself reversible)
2. Restores from the most recent numbered checkpoint
3. Prints `Reverted to checkpoint {target}.`

So `python scripts/checkpoint.py --revert` after a botched push restores the snapshot Step 3 took (`pre-delete-{TopicName}`) plus any subsequent ones up to the most recent. No code change needed; thread is just a confirmation.

johnguy0 (commented):
Took another pass at the round-2 fixes against `cf69dfe` (single commit). All three threads from the first round are addressed:

1. `cleanup/SKILL.md` Rules — the no-terminal-commands line now carves out `python scripts/checkpoint.py` and `python scripts/push.py` explicitly and points back at Steps 2b and 6, so the agent won't refuse the very steps the skill asks it to perform. ✅
2. `topics/test/SKILL.md` and `workflows/test/SKILL.md` — both now have a real "not yet implemented, here's the manual step" body (Copilot Studio Test pane / Power Automate Test → Manually) plus an explicit "Then stop. Do NOT pretend to run a test or report results." guard. Menu invocation lands on a concrete instruction instead of an empty stub, and the TODO comments stay so the future automated implementation is still flagged. ✅
3. `topics/delete/SKILL.md` `--revert` flag — confirmed against `solutions/ess-maker-skills/scripts/checkpoint.py`. `cmd_revert` exists at line 131, is documented in the usage docstring at line 12, and correctly auto-saves the current state to a fresh checkpoint before restoring `existing[-1]` (the last user-created checkpoint, computed before the auto-save). The fallback path in the SKILL works as written. ✅

No new feedback against the round-2 diff. Ready for approval from my side.

PR #57, opened 2026-05-07
Add CONFIG-013: Knowledge source crawl/index readiness check

saengland:
Reopens #37. The original was closed by GitHub when I retargeted base from main -> main2 (orphan-history collision; main2 has no common ancestor with main). Once main2 becomes default, @GrahamMcMynn please rebase this branch onto main2 and force-push, then I'll retarget. Existing review comments are visible on the closed #37 and will be re-posted here once the rebase lands. CI will re-run.


scripts/flightcheck/checks/local_files.py
```diff
@@ -288,3 +290,199 @@ def _check_template_configs(agent_path: Path, label: str) -> list[CheckResult]:
...
+            description=f"{label}: Knowledge source readiness",
+            result="No knowledge source files found",
+        ))
+        return results
+
+    # Require the Island Gateway API for live status
+    pva = getattr(runner, "pva", None) if runner else None
+    bot_id = None
+    if runner and hasattr(runner, "config"):
+        bot_id = runner.config.get("agent", {}).get("botId")
```
johnguy0, line 336: Blocker. Schema mismatch with the rest of the codebase. `_check_single_agent` is called per agent in a loop (line 62) for the multi-agent layout, and PR #59's CONFIG-014 reads `bot_id` correctly via `for agent in config.get('agents', []): if agent.get('slug') == agent_name`. This line uses the legacy single-agent fallback `config.get('agent', {}).get('botId')`, which returns `None` against the multi-agent schema, so `bot_id` stays None, the guard on line 338 short-circuits, and CONFIG-013 silently never runs. It will look like the check is enabled but it never queries the gateway.

Fix: pass `agent_name` through to `_check_knowledge_sources` (the call site on line 89 doesn't), then look up the matching entry in `config.get('agents', [])` by slug, with the single-agent `config.get('agent', {})` as a backward-compat fallback. The CONFIG-014 loop in PR #59 is the pattern to mirror.

scripts/flightcheck/checks/local_files.py
```diff
@@ -288,3 +290,199 @@ def _check_template_configs(agent_path: Path, label: str) -> list[CheckResult]:
...
+
+    if not pva or not hasattr(pva, "is_configured") or not pva.is_configured or not bot_id:
```
johnguy0, line 338: Minor. The `hasattr(pva, 'is_configured')` guard is defensive but unnecessary, `is_configured` is a property we own on `PVAClient` and is always present. Simplify to `if not pva or not pva.is_configured or not bot_id`. Reads more directly.

scripts/flightcheck/pva_client.py
```diff
@@ -0,0 +1,197 @@
...
+        # Extract the org name from our env_url for matching
+        # e.g., "https://orgb78b4a3b.crm.dynamics.com" → "orgb78b4a3b"
+        org_match = self.env_url.split("//")[1].split(".")[0] if "//" in self.env_url else ""
+
+        envs = resp.json().get("value", [])
+        for env in envs:
+            props = env.get("properties", {})
+            linked = props.get("linkedEnvironmentMetadata", {})
+            instance_url = linked.get("instanceUrl", "")
+            if org_match and org_match in instance_url:
```
johnguy0, line 142: Substring match is fragile. `'org123' in 'org1234contoso.crm.dynamics.com'` is true. With multiple environments in the same tenant where org names share a prefix, this can pick the wrong env. Pull the hostname out of `instance_url` (e.g., `urlparse(instance_url).hostname.split('.')[0]`) and compare for equality against `org_match`.

scripts/flightcheck/pva_client.py
```diff
@@ -0,0 +1,197 @@
...
+    import requests
+except ImportError:
+    print("ERROR: 'requests' package not found. Run: pip install requests")
+    sys.exit(1)
+
+
+CLIENT_ID = "51f81489-12ee-4a9e-aaae-a2591f45987d"
+
+# PVA app ID — used as the token audience for the Island Gateway API
+PVA_SCOPE = "96ff4394-9197-43aa-b393-6a41652e21f8/.default"
```
johnguy0, line 32: Add a comment naming this GUID. `96ff4394-9197-43aa-b393-6a41652e21f8` is the well-known Power Virtual Agents resource ID, but unannotated it looks like a magic string and the next person to read this won't know if it's safe to change. One-line comment is enough.

johnguy0 (commented):
Reviewed CONFIG-013 (knowledge source readiness) and the new Island Gateway client. The architecture is right, lifecycle `state` vs crawl `status` distinction is exactly the kind of false-pass we want to avoid, and the WARNING-vs-PASSED handling for missing/unknown statuses is the right call.

One real blocker around how the config is read, plus a couple of smaller things in the gateway client. Inline below.

Copilot-instructions for the no-fabricated-URLs rule reads well, no concerns there.

saengland:
Reposted as #63 onto new `main` (post-cutover). Path translation handled via split-patch. Authorship preserved via Co-authored-by. Will close this once #63 lands.

saengland:
Closing as superseded by #63 (reposted onto new `main` after the main2 -> main cutover). Authorship preserved on the replacement via Co-authored-by trailer. The branch behind this PR is left untouched - delete at your discretion.

PR #59, opened 2026-05-07
FlightCheck: Add CONFIG-014 topic description quality check

saengland:
Reopens #40. The original was closed by GitHub when I retargeted base from main -> main2 (orphan-history collision; main2 has no common ancestor with main). Once main2 becomes default, @GrahamMcMynn please rebase this branch onto main2 and force-push, then I'll retarget. Existing review comments are visible on the closed #40 and will be re-posted here once the rebase lands. CI will re-run.

### DIFF
--- a/scripts/flightcheck/checks/local_files.py
+++ b/scripts/flightcheck/checks/local_files.py
@@ -59,12 +59,12 @@ def run_local_file_checks(runner) -> list[CheckResult]:
     # Run checks for each agent
     for agent_path in sorted(agent_folders):
         agent_name = agent_path.name
-        results.extend(_check_single_agent(agent_path, agent_name))
+        results.extend(_check_single_agent(agent_path, agent_name, runner))
 
     return results
 
 
-def _check_single_agent(agent_path: Path, agent_name: str) -> list[CheckResult]:
+def _check_single_agent(agent_path: Path, agent_name: str, runner=None) -> list[CheckResult]:
     """Run all local file checks for a single agent."""
     results: list[CheckResult] = []
 
@@ -82,6 +82,9 @@ def _check_single_agent(agent_path: Path, agent_name: str) -> list[CheckResult]:
 
     # ---- Variables ----
     results.extend(_check_variables(agent_path, label))
+
+    # ---- Topic description quality ----
+    results.extend(_check_topic_descriptions(agent_path, label, runner, agent_name))
 
     # ---- Template configs ----
     results.extend(_check_template_configs(agent_path, label))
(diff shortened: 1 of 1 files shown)


scripts/flightcheck/checks/local_files.py
```diff
@@ -269,6 +272,199 @@ def _check_variables(agent_path: Path, label: str) -> list[CheckResult]:
...
+            filter_expr=f"_parentbotid_value eq '{bot_id}' and componenttype eq 9 and statecode eq 1",
+        )
+        # Convert Dataverse names to local filenames for matching
+        disabled = set()
+        for c in components:
+            name = c.get("name", "")
+            if name:
+                disabled.add(_friendly_filename(name))
+        return disabled
+    except Exception:
```
johnguy0, line 347: Bare `except Exception: return set()` will silently disable the disabled-topics filter on any Dataverse error (auth, network, schema change, query syntax, you name it). When that happens, CONFIG-014 starts flagging legitimately disabled topics for short descriptions, which is exactly the noisy false-positive case we're trying to avoid.

At minimum, log the exception so we can see it when it happens. Better: catch the specific exceptions we expect (`requests.RequestException`, `KeyError`, etc.) and let unexpected ones propagate.

scripts/flightcheck/checks/local_files.py
```diff
@@ -269,6 +272,199 @@ def _check_variables(agent_path: Path, label: str) -> list[CheckResult]:
...
+    r"response-preparation",
+]
+
+# Placeholder phrases that indicate the description was never filled in
+_PLACEHOLDER_PATTERNS = [
+    r"\[add\s+keywords",
+    r"\[add\s+.*here\]",
+    r"\[describe",
+    r"\[placeholder",
+    r"\btodo\b",
```
johnguy0, line 294: `\btodo\b` is too aggressive. Any topic whose description legitimately mentions "TODO list" or "todo items" (think a personal-tasks topic) gets flagged as having placeholder text. Tighten to anchor on common placeholder phrasings, e.g. `\btodo:\b` or `\btodo\s*-` or `^TODO\b`. Same goes for `\bplaceholder\b` if any topic happens to talk about placeholders as a concept.

scripts/flightcheck/checks/local_files.py
```diff
@@ -269,6 +272,199 @@ def _check_variables(agent_path: Path, label: str) -> list[CheckResult]:
...
+    checked = 0
+
+    # Query Dataverse for disabled topics to skip them
+    disabled_topics = _get_disabled_topic_names(runner, agent_name)
+
+    for tf in sorted(topics_dir.glob("*.mcs.yml")):
+        filename = tf.name.lower()
+
+        # Skip system topics that don't use AI-based routing
+        if any(re.search(pat, filename) for pat in _SYSTEM_TOPIC_PATTERNS):
```
johnguy0, line 376: `re.search` is unanchored, so `_SYSTEM_TOPIC_PATTERNS = ["on-error", ...]` will match any filename that contains "on-error" anywhere, not just system topics named that. A topic file like `handle-on-error-cases.mcs.yml` would get skipped silently. Anchor the pattern (`^conversation-start\.mcs\.yml$` style) or compare the stem against an exact set.

scripts/flightcheck/checks/local_files.py
```diff
@@ -269,6 +272,199 @@ def _check_variables(agent_path: Path, label: str) -> list[CheckResult]:
...
+        stem = tf.stem.replace(".mcs", "").lower()
+        if stem in disabled_topics:
+            continue
+
+        content = tf.read_text(encoding="utf-8", errors="replace")
+
+        # Extract modelDescription (multiline block scalar or inline)
+        # The pattern allows blank lines within the indented block
+        desc_match = re.search(
+            r'modelDescription:\s*\|[\-\+]?\s*\n((?:(?:[ \t]+.+|[ \t]*)\n?)+)', content
```
johnguy0, line 389: Parsing YAML block scalars with regex is going to bite us eventually, the regex captures across keys if the indentation is off and misses valid forms like `>`-style scalars or quoted strings spanning lines. Since `pyyaml` is already a transitive dep of the project, using `yaml.safe_load(content).get('beginDialog', {}).get('intent', {}).get('modelDescription')` (or wherever it actually lives in the schema) would be more robust. Fine to land as-is and follow up if you'd rather scope the change down.

johnguy0 (commented):
CONFIG-014 is a useful addition, the placeholder/short-description heuristics catch real authoring problems. The split between `FAILED` for placeholders and `WARNING` for short descriptions is the right severity call.

A few smaller things that aren't strictly blocking but worth tightening, especially the silent exception swallowing and the TODO-anywhere-in-the-string false positive. Inline below.

saengland:
Reposted as #62 onto new `main` (post-cutover). Authorship preserved via Co-authored-by. Will close this once #62 lands.

saengland:
Closing as superseded by #62 (reposted onto new `main` after the main2 -> main cutover). Authorship preserved on the replacement via Co-authored-by trailer. The branch behind this PR is left untouched - delete at your discretion.

PR #46, opened 2026-05-06
Sync missing Workday samples (extended topics + scenario screenshots) from upstream

srideshpande:
## Summary

Synced missing Workday samples from the upstream `CopilotStudioSamples/EmployeeSelfServiceAgent` repo into `src/samples/workday/`. Adds 17 files that were absent from the kit; leaves all existing files (identical or kit-customized) untouched.

## What changed

Added **17 files** under `src/samples/workday/`:

### New folder: `workday/extended/` (9 files)
Extended Workday topic samples (Inbox Tasks, Pay Slips, Request Feedback, Transfer Employee).

- `workday/extended/README.md`
- `workday/extended/GetInboxTasks.yaml`
- `workday/extended/GetPaySlips.yaml`
- `workday/extended/RequestFeedback.yaml`
- `workday/extended/TransferEmployee.yaml`
- `workday/extended/assets/GetInboxTasks.png`
- `workday/extended/assets/GetPaySlips.png`
- `workday/extended/assets/RequestFeedback.png`
- `workday/extended/assets/TransferEmployee.png`

### Scenario screenshots added to existing `workday/employee/` folders (8 files)
- `EmployeeAddDependents/add_dependents.png`
- `EmployeeGetVacationBalance/get_vacation_balance.png`
- `EmployeeUpdateResidentialAddress/update_address.png`
- `WorkdayEmployeeRequestTimeOff/request_time_off.png`
- `WorkdayEmployeeRequestTimeOff/requestTimeOffAdapativeCard.png`
- `WorkdayEmployeeRequestTimeOff/requestTimeOffSubmitted.png`
- `WorkdayManageEmergencyContact/update_emergency_contact.png`
- `WorkdayManageEmergencyContact/update_emergency_contact_2.png`

## What was NOT changed (and why)

A full diff of the upstream `Workday/` tree vs. `src/samples/workday/` found 65 Workday files. The 17 above were the only ones not already present.

| Status | Workday files | Action |
|---|---|---|
| New (missing at destination) | 17 | **Copied** |
| Already identical (byte-for-byte) | 35 | Skipped — already present |
| Exist but content differs | 13 | **Left untouched** — kit-customized versions preserved |

[...]

PR #60, opened 2026-05-07
Sync missing Workday samples (extended topics + scenario screenshots) from upstream

saengland:
Reopens #46. The original was closed by GitHub when I retargeted base from main -> main2 (orphan-history collision; main2 has no common ancestor with main). Once main2 becomes default, @srideshpande please rebase this branch onto main2 and force-push, then I'll retarget. Existing review comments are visible on the closed #46 and will be re-posted here once the rebase lands. CI will re-run.


src/samples/workday/extended/RequestFeedback.yaml
```diff
@@ -0,0 +1,393 @@
...
+      variable: Topic.responderName
+      value: =If(IsBlank(Topic.selectedResponderId), Topic.responderName, LookUp(Topic.parsedResponder.data, id = Topic.selectedResponderId).descriptor)
+
+    # Map feedbackSharing choice to Workday fields:
+    # "shareWithMe"      - feedbackConfidential=false, showFeedbackProviderName=true
+    # "shareWithManager" - feedbackConfidential=true,  showFeedbackProviderName=true
+    - kind: SetVariable
+      id: set_confidential
+      variable: Topic.feedbackConfidential
+      value: =If(Topic.feedbackSharing = "shareWithManager", "true", "false")
```
johnguy0, line 292: This mapping looks inverted to me, and at minimum the comments above contradict the UI. Workday's `feedbackConfidential=true` means the feedback request stays private to the requester (manager doesn't see it). The comment block on lines 287-288 maps `shareWithManager` to `feedbackConfidential=true`, but the visibility label on line 361 then renders `"Shared with you and your manager"` for that same choice. Those two facts can't both be right.

Can you double-check against the Workday `RequestFeedback` API doc and either flip the boolean (`shareWithManager` → `false` if confidential=true means "private") or rewrite the UI label so it matches what's actually being submitted? Whichever way it goes, the user is currently being told the opposite of what we send.

johnguy0 (commented):
Took a pass through the four extended Workday samples (inbox tasks, pay slips, request feedback, transfer employee). The shape is consistent with the existing employee samples, the `modelDescription` blocks have the display rules and "answer only for the requesting user" guardrails, and the adaptive-card layouts are reasonable.

One real bug in `RequestFeedback.yaml` around the `feedbackConfidential` mapping that I want resolved before merge. Inline below.

The rest is fine.

saengland:
Reposted as #64 onto new `main` (post-cutover). Path translation done with binary-safe git show extraction. Layout question for you in the new PR description (whether to keep ExtendedScenarios/ as a sibling folder or collapse into EmployeeScenarios/). Will close this once #64 lands.

saengland:
Closing as superseded by #64 (reposted onto new `main` after the main2 -> main cutover). Authorship preserved on the replacement via Co-authored-by trailer. The branch behind this PR is left untouched - delete at your discretion.

PR #54, opened 2026-05-07
PR #8: Copilot prompts + orchestration

saengland:
Reopens #32. The original PR was auto-merged into `main2` on May 7 without reviewer approval (branch protection wasn't yet in place; auto-merge had nothing to wait for). The merge has been reverted on `main2` and this PR re-submits the same head commit (`ed62264`) for proper review.

No code changes from #32 - same head SHA. CI will re-run.

PR #62, opened 2026-05-08
FlightCheck: Add CONFIG-014 topic description quality check [reposted from #59]

saengland:
Reposts #59 (Graham's CONFIG-014 topic description quality check) onto the new main after today's cutover.

## What changed vs original
- Path translation only: scripts/flightcheck/checks/local_files.py -> solutions/ess-maker-skills/scripts/flightcheck/checks/local_files.py
- Used git am --3way to handle the divergence between old main's local_files.py (237 lines) and new main's (236 lines). The 3-way merge resolved cleanly, post-apply file is 401 lines with CONFIG-014 added.
- Authorship preserved via Co-authored-by trailer.

John flagged three items in his round-1 review on #59:
- `\btodo\b` is too aggressive (will catch legitimate "TODO list" topics). Recommend tightening to a more specific placeholder pattern.
- `_SYSTEM_TOPIC_PATTERNS` regex is unanchored, will match any filename containing the substring.
- YAML block-scalar regex parsing will break on `>`-style scalars and quoted forms.

Recommend addressing in fixup commits on this branch before merge.

Closes #59.

Co-authored-by: Graham McMynn <grahamc@microsoft.com>

### DIFF
--- a/solutions/ess-maker-skills/scripts/flightcheck/checks/local_files.py
+++ b/solutions/ess-maker-skills/scripts/flightcheck/checks/local_files.py
@@ -280,31 +280,41 @@ _MIN_DESCRIPTION_WORDS = 20
 # Match against the exact filename stem (without ".mcs.yml") so we don't
 # accidentally skip topics whose filename merely contains one of these words
 # (e.g. "handle-checkout-error.mcs.yml" must not match "on-error").
+# Both `log-telemetry` and `log-telemetry-event` are included since different
+# ESS template versions use either name.
 _SYSTEM_TOPIC_STEMS = {
     "conversation-start",
     "on-error",
     "reset-conversation",
+    "log-telemetry",
     "log-telemetry-event",
     "microsoft-self-help",
     "response-preparation",
 }
 
-# Placeholder marker patterns that indicate the description was never filled in.
-# Use specific marker forms (TODO:, [TODO], <placeholder>, etc.) — a simple
-# \btodo\b would false-flag legitimate topics like "Today's tasks" or
-# descriptions that mention the word "todo" in normal sentences.
-_PLACEHOLDER_PATTERNS = [
+# Placeholder patterns are split into two groups because the matching loops
+# below use different case-sensitivity rules:
+#
+# _PLACEHOLDER_PATTERNS_INSENSITIVE: bracketed/markup forms and full phrases
+#   that are unambiguous markers regardless of how they're cased.
+# _PLACEHOLDER_PATTERNS_SENSITIVE: all-caps marker forms that authors actually
+#   leave behind (TODO:, TBD, PLACEHOLDER). These MUST stay case-sensitive —
+#   matching them under re.IGNORECASE would false-flag legitimate sentences
+#   like "this topic acts as a placeholder until ..." or "today's todo list".
+_PLACEHOLDER_PATTERNS_INSENSITIVE = [
     r"\[add\s+keywords",
     r"\[add\s+.*here\]",
     r"\[describe",
     r"\[placeholder",
+    r"<placeholder>",
+    r"describe\s+this\s+topic",
+    r"add\s+your\s+description",
+]
+_PLACEHOLDER_PATTERNS_SENSITIVE = [
     r"\bTODO:",
     r"\[TODO\]",
     r"\bTBD\b",
-    r"<placeholder>",
     r"\bPLACEHOLDER\b",
-    r"describe\s+this\s+topic",
-    r"add\s+your\s+description",
 ]
 
 
@@ -323,13 +333,11 @@ def _get_disabled_topic_names(runner, agent_name: str) -> set[str]:
     if not runner or not getattr(runner, 'dv_token', None) or not getattr(runner, 'env_url', None):
         return set()
 
-    # Find botId from config
-    import json
-    try:
-        config = json.load(open("my/config.json"))
-    except (FileNotFoundError, json.JSONDecodeError):
-        return set()
-
+    # Find botId from config. Prefer the already-parsed runner.config attached
+    # by cli.py — re-reading from disk hardcodes the config path and breaks
+    # whenever the kit's layout changes (regression after the folder reorg
+    # moved my/config.json to .local/config.json).
+    config = getattr(runner, "config", None) or {}
     bot_id = None
     for agent in config.get("agents", []):
         if agent.get("slug") == agent_name:
@@ -425,8 +433,14 @@ def _check_topic_descriptions(agent_path: Path, label: str, runner=None, agent_n
             readable = stem.replace("-", " ").title()
             display_name = f"{readable} (file: {tf.name})"
 
-        # Check for placeholder text
-        if any(re.search(pat, desc_text, re.IGNORECASE) for pat in _PLACEHOLDER_PATTERNS):
+        # Check for placeholder text. Two passes because the marker patterns
+        # (TODO:, TBD, PLACEHOLDER) MUST stay case-sensitive — matching them
+        # under IGNORECASE would re-introduce the false-flag bug from round 1.
+        is_placeholder = (
+            any(re.search(pat, desc_text, re.IGNORECASE) for pat in _PLACEHOLDER_PATTERNS_INSENSITIVE)
+            or any(re.search(pat, desc_text) for pat in _PLACEHOLDER_PATTERNS_SENSITIVE)
+        )
+        if is_placeholder:
             has_placeholder.append(display_name)
             continue
 
--- a/solutions/ess-maker-skills/scripts/flightcheck/checks/local_files.py
+++ b/solutions/ess-maker-skills/scripts/flightcheck/checks/local_files.py
@@ -58,12 +58,12 @@ def run_local_file_checks(runner) -> list[CheckResult]:
     # Run checks for each agent
     for agent_path in sorted(agent_folders):
         agent_name = agent_path.name
-        results.extend(_check_single_agent(agent_path, agent_name))
+        results.extend(_check_single_agent(agent_path, agent_name, runner))
 
     return results
 
 
-def _check_single_agent(agent_path: Path, agent_name: str) -> list[CheckResult]:
+def _check_single_agent(agent_path: Path, agent_name: str, runner=None) -> list[CheckResult]:
     """Run all local file checks for a single agent."""
     results: list[CheckResult] = []
 
@@ -81,6 +81,9 @@ def _check_single_agent(agent_path: Path, agent_name: str) -> list[CheckResult]:
 
     # ---- Variables ----
     results.extend(_check_variables(agent_path, label))
+
+    # ---- Topic description quality ----
+    results.extend(_check_topic_descriptions(agent_path, label, runner, agent_name))
 
     # ---- Template configs ----
     results.extend(_check_template_configs(agent_path, label))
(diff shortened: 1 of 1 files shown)


solutions/ess-maker-skills/scripts/flightcheck/checks/local_files.py
```diff
@@ -268,6 +271,199 @@ def _check_variables(agent_path: Path, label: str) -> list[CheckResult]:
...
+    r"response-preparation",
+]
+
+# Placeholder phrases that indicate the description was never filled in
+_PLACEHOLDER_PATTERNS = [
+    r"\[add\s+keywords",
+    r"\[add\s+.*here\]",
+    r"\[describe",
+    r"\[placeholder",
+    r"\btodo\b",
```
johnguy0: `\btodo\b` is too aggressive. A real topic called "TODO list", "Today's tasks", or any description that mentions "todo" in a sentence ("this topic handles user todo entries") will get failed as a placeholder. Tighten to a marker pattern that only matches authoring placeholders: `r"\bTODO:"`, `r"\[TODO\]"`, or `r"\bTBD\b"` (the all-caps form authors actually leave behind). Same applies to `\bplaceholder\b` on line 294, match `<placeholder>` or `PLACEHOLDER` instead.

solutions/ess-maker-skills/scripts/flightcheck/checks/local_files.py
```diff
@@ -268,6 +271,199 @@ def _check_variables(agent_path: Path, label: str) -> list[CheckResult]:
...
+    checked = 0
+
+    # Query Dataverse for disabled topics to skip them
+    disabled_topics = _get_disabled_topic_names(runner, agent_name)
+
+    for tf in sorted(topics_dir.glob("*.mcs.yml")):
+        filename = tf.name.lower()
+
+        # Skip system topics that don't use AI-based routing
+        if any(re.search(pat, filename) for pat in _SYSTEM_TOPIC_PATTERNS):
```
johnguy0: `re.search(pat, filename)` is unanchored, so the `_SYSTEM_TOPIC_PATTERNS` entries match anywhere in the filename. `on-error` will skip a legitimate topic file like `handle-checkout-error.mcs.yml`, `reset-conversation` will skip `reset-conversation-history.mcs.yml`, etc. Either anchor each pattern (`^conversation-start\.` etc.) or just compare `tf.stem.replace(".mcs", "")` against a set of exact filename stems. The set form is cheaper and reads better.

solutions/ess-maker-skills/scripts/flightcheck/checks/local_files.py
```diff
@@ -268,6 +271,199 @@ def _check_variables(agent_path: Path, label: str) -> list[CheckResult]:
...
+        if stem in disabled_topics:
+            continue
+
+        content = tf.read_text(encoding="utf-8", errors="replace")
+
+        # Extract modelDescription (multiline block scalar or inline)
+        # The pattern allows blank lines within the indented block
+        desc_match = re.search(
+            r'modelDescription:\s*\|[\-\+]?\s*\n((?:(?:[ \t]+.+|[ \t]*)\n?)+)', content
+        )
```
johnguy0: Regex-parsing YAML works for the literal-block (`|`) and inline forms you've covered, but it will silently miss real topics that use any other valid YAML scalar form: folded block (`modelDescription: >`), single-quoted (`modelDescription: 'foo bar'`), double-quoted, or anything with a comment trailing the key on the same line. Those topics will fall through to the `continue` on line 398 and not get checked at all, so a placeholder description in a folded-block topic would silently pass. Parse the YAML for real with `yaml.safe_load(content)` and walk `actions[]`/`triggers[]` to find the `modelDescription` field. `yaml` is already a dependency of the kit.

johnguy0 (commented):
Path-translation repost of #59 with a 3-way merge. Pulled the diff and re-checked CONFIG-014 against the new path. Three-way merge looks clean, file ends at 401 lines with the new check wired in.

Same three follow-ups I flagged on #59 are still in this branch as written. None of them are deal-breakers in isolation, but together they will produce noisy false positives on real authoring work, which is exactly what we don't want from a quality check. Inline below.

Happy to merge once those are addressed, or split them into a fixup commit on this branch.

saengland:
Caught a stale `my/` reference in this branch. The convention John established in his PR #2 review (and reinforced 8+ times since, most recently PR #54) is `my/` -> `.local/` for kit-managed bookkeeping like `config.json`. The scripts in this repo already use the new path everywhere (`auth.py:283`, `setup.py:514`, `checkpoint.py:28`, `flightcheck/cli.py:83`).

In `scripts/flightcheck/checks/local_files.py`:

```python
config = json.load(open("my/config.json"))
```

should be:

```python
config = json.load(open(".local/config.json"))
```

This is runtime code, so the check actually breaks if it runs against a properly-configured kit (the file lives at `.local/config.json`, not `my/config.json`). Cheap to fix in a fixup commit alongside John's three CONFIG-014 false-positive items.

solutions/ess-maker-skills/scripts/flightcheck/checks/local_files.py
GrahamMcMynn: Fixed in commit 4839397. Removed broad word-boundary matches (todo, placeholder) that would false-flag legitimate topics. The list now matches authoring marker forms only: TODO:, [TODO], TBD, <placeholder>, PLACEHOLDER. The bracketed and full-phrase patterns (e.g. [add keywords here], describe this topic, add your description) are kept since they have low false-positive risk.

solutions/ess-maker-skills/scripts/flightcheck/checks/local_files.py
GrahamMcMynn: Fixed in commit 4839397. Replaced the regex list _SYSTEM_TOPIC_PATTERNS with an exact-match set _SYSTEM_TOPIC_STEMS. The check is now `if stem in _SYSTEM_TOPIC_STEMS` against the literal filename without .mcs.yml - cheaper, safer, no false skips for files like handle-checkout-error.mcs.yml. Verified with a synthetic test case.

solutions/ess-maker-skills/scripts/flightcheck/checks/local_files.py
GrahamMcMynn: Fixed in commit 4839397. Switched to yaml.safe_load (PyYAML is already in scripts/requirements.txt at >=6.0.2). The check now correctly handles every valid YAML scalar form: literal block, folded block, single-quoted, double-quoted, plain. Verified end-to-end with a test that built a synthetic agent folder containing all four scalar forms - the previously-silent folded-block placeholder topic is now correctly FAILED. Also added explicit YAML parse error reporting so files with broken YAML surface as a WARNING with the filename instead of being silently skipped.

solutions/ess-maker-skills/scripts/flightcheck/checks/local_files.py
```diff
@@ -268,6 +273,224 @@ def _check_variables(agent_path: Path, label: str) -> list[CheckResult]:
...
+_PLACEHOLDER_PATTERNS = [
+    r"\[add\s+keywords",
+    r"\[add\s+.*here\]",
+    r"\[describe",
+    r"\[placeholder",
+    r"\bTODO:",
+    r"\[TODO\]",
+    r"\bTBD\b",
+    r"<placeholder>",
+    r"\bPLACEHOLDER\b",
```
johnguy0, line 317: The `re.search(..., re.IGNORECASE)` flag a few lines down (where these patterns are matched) makes `\bPLACEHOLDER\b` here functionally identical to the `\bplaceholder\b` pattern from round 1 that we agreed was too aggressive. A real description like "this topic acts as a placeholder until the AI routing description is finalized" would still false-flag.

Either drop `\bPLACEHOLDER\b` (the `<placeholder>` literal above already covers the common authoring marker), or split `_PLACEHOLDER_PATTERNS` into a case-sensitive group and a case-insensitive group so the all-caps form actually requires all-caps.

solutions/ess-maker-skills/scripts/flightcheck/checks/local_files.py
```diff
@@ -268,6 +273,224 @@ def _check_variables(agent_path: Path, label: str) -> list[CheckResult]:
...
+
+# System topics that don't use AI-based routing (no description needed).
+# Match against the exact filename stem (without ".mcs.yml") so we don't
+# accidentally skip topics whose filename merely contains one of these words
+# (e.g. "handle-checkout-error.mcs.yml" must not match "on-error").
+_SYSTEM_TOPIC_STEMS = {
+    "conversation-start",
+    "on-error",
+    "reset-conversation",
+    "log-telemetry-event",
```
johnguy0, line 290: Quick sanity check on this rename. Round 1 had `r"log-telemetry"` as a regex, which would skip any filename containing `log-telemetry` (so both `log-telemetry.mcs.yml` and `log-telemetry-event.mcs.yml`). The new exact stem `log-telemetry-event` only skips one of those. If any agent in the repo has a `log-telemetry.mcs.yml` (without the `-event` suffix), it'll now get checked instead of skipped. Was that intentional, or should this entry be `log-telemetry`? If both filenames are in use across agents, add both to the set.

johnguy0 (commented):
Re-reviewing the round 2 changes. The yaml.safe_load swap and the exact-match stem set both look right, and the parse-error reporting is a nice add. Two things on the placeholder/system-topic changes I want to flag before this is approval-ready, both inline.

solutions/ess-maker-skills/scripts/flightcheck/checks/local_files.py
GrahamMcMynn: Good catch - same problem applied to TODO:, [TODO], and TBD as well, all four ALL-CAPS markers were neutralized by IGNORECASE. Fixed in commit 26427fd by splitting _PLACEHOLDER_PATTERNS into two groups: _PLACEHOLDER_PATTERNS_INSENSITIVE for bracketed/markup forms ([add keywords], <placeholder>, full phrases) and _PLACEHOLDER_PATTERNS_SENSITIVE for ALL-CAPS markers (TODO:, [TODO], TBD, PLACEHOLDER) which are now matched without IGNORECASE. Verified with synthetic tests covering legitimate descriptions like 'this topic acts as a placeholder until...', 'today todo list', and 'meeting time is tbd' - no false flags - while real markers like 'TODO: write the actual description' still get caught.

solutions/ess-maker-skills/scripts/flightcheck/checks/local_files.py
GrahamMcMynn: Good catch - confirmed in the test tenant only log-telemetry-event is present, but you're right that other agents may use the shorter log-telemetry name. Fixed in commit 26427fd by adding both stems to _SYSTEM_TOPIC_STEMS so neither form gets accidentally validated. Verified with synthetic tests creating both filenames - both correctly skipped.

johnguy0 (commented):
Round 3 looks good. Both round-2 items are addressed cleanly:

The PLACEHOLDER split into `_PLACEHOLDER_PATTERNS_INSENSITIVE` and `_PLACEHOLDER_PATTERNS_SENSITIVE` is exactly the right shape: the bracketed and full-phrase forms keep `re.IGNORECASE` (where false positives are not a real risk) and the all-caps markers (`TODO:`, `[TODO]`, `TBD`, `PLACEHOLDER`) drop it, so a real description like "this topic acts as a placeholder until ..." no longer false-flags. The inline comment explaining why the case-sensitive group must stay case-sensitive is the right thing to have there, makes the round-1 regression risk obvious to the next reader.

Adding both `log-telemetry` and `log-telemetry-event` to `_SYSTEM_TOPIC_STEMS` covers the cross-template case I was worried about, with no false skips since these are exact stem matches.

The bonus CONFIG-014 fix (switching `_get_disabled_topic_names` from re-reading `my/config.json` to using `runner.config`) is a genuine improvement on top of the round-2 ask. `cli.py` does `runner.config = config` before invoking the checks, so the attribute is reliably populated, and the `getattr(runner, "config", None) or {}` guard is a fine safety net. Decoupling the check from a hardcoded config path also kills the silently-caught exception that was masking the disabled-topic skip after the folder reorg.

Good to merge from my side.

PR #63, opened 2026-05-08
Add CONFIG-013: Knowledge source crawl/index readiness check [reposted from #57]

saengland:
Reposts #57 (Graham's CONFIG-013 + PVAClient) onto the new main after today's cutover.

## What changed vs original

PR #57 touched files at both root (CONTRIBUTING.md) and under what was scripts/ (now solutions/ess-maker-skills/scripts/). Used a split-patch approach to translate paths correctly:
- CONTRIBUTING.md stays at root.
- .github/copilot-instructions.md addition -> solutions/ess-maker-skills/.github/copilot-instructions.md
- scripts/flightcheck/{checks/local_files.py, cli.py, pva_client.py} -> solutions/ess-maker-skills/scripts/flightcheck/...
- Both root patch (--apply --3way) and solution patch (--apply --3way --directory=solutions/ess-maker-skills/) applied cleanly.
- Authorship preserved via Co-authored-by trailer on the squash commit.

## Fixup commit added on this branch

PR #57's CONTRIBUTING.md addition cross-referenced .github/copilot-instructions.md#no-fabricated-urls, which was correct pre-cutover but is now broken (file lives at solutions/ess-maker-skills/.github/copilot-instructions.md). Added a fixup commit updating the link target.

John flagged three items in his round-1 review on #57:
- pva_client.py:32 - GUID 96ff4394-9197-43aa-b393-6a41652e21f8 needs a naming comment (well-known PVA resource ID).
- pva_client.py:142 - substring match on org name is fragile (org123 will match org1234contoso).
- local_files.py:338 - hasattr(pva, 'is_configured') guard is defensive but unnecessary.

Recommend addressing in fixup commits on this branch before merge.

Closes #57.

Co-authored-by: Graham McMynn <grahamc@microsoft.com>

### DIFF
--- a/solutions/ess-maker-skills/scripts/flightcheck/checks/local_files.py
+++ b/solutions/ess-maker-skills/scripts/flightcheck/checks/local_files.py
@@ -385,7 +385,7 @@ def _check_knowledge_sources_via_gateway(pva, bot_id: str, knowledge_files: list
             remediation=(
                 "Publish the agent in Copilot Studio to provision missing sources, "
                 "or remove the local file if it should not exist. Identify mismatched "
-                "sources by comparing filenames in my/agents/{slug}/knowledge/ against "
+                "sources by comparing filenames in workspace/agents/{slug}/knowledge/ against "
                 "Copilot Studio → Knowledge."
             ),
         ))
--- a/solutions/ess-maker-skills/scripts/flightcheck/checks/local_files.py
+++ b/solutions/ess-maker-skills/scripts/flightcheck/checks/local_files.py
@@ -335,7 +335,7 @@ def _check_knowledge_sources(agent_path: Path, label: str, runner=None) -> list[
     if runner and hasattr(runner, "config"):
         bot_id = runner.config.get("agent", {}).get("botId")
 
-    if not pva or not hasattr(pva, "is_configured") or not pva.is_configured or not bot_id:
+    if not pva or not pva.is_configured or not bot_id:
         results.append(CheckResult(
             checkpoint_id="CONFIG-013", category=f"Knowledge Sources ({label})",
             priority=Priority.HIGH.value, status=Status.SKIPPED.value,
--- a/solutions/ess-maker-skills/scripts/flightcheck/checks/local_files.py
+++ b/solutions/ess-maker-skills/scripts/flightcheck/checks/local_files.py
@@ -58,12 +58,12 @@ def run_local_file_checks(runner) -> list[CheckResult]:
     # Run checks for each agent
     for agent_path in sorted(agent_folders):
         agent_name = agent_path.name
-        results.extend(_check_single_agent(agent_path, agent_name))
+        results.extend(_check_single_agent(agent_path, agent_name, runner))
 
     return results
 
 
-def _check_single_agent(agent_path: Path, agent_name: str) -> list[CheckResult]:
+def _check_single_agent(agent_path: Path, agent_name: str, runner=None) -> list[CheckResult]:
     """Run all local file checks for a single agent."""
     results: list[CheckResult] = []
 
@@ -84,6 +84,9 @@ def _check_single_agent(agent_path: Path, agent_name: str) -> list[CheckResult]:
 
     # ---- Template configs ----
     results.extend(_check_template_configs(agent_path, label))
+
+    # ---- Knowledge source readiness ----
+    results.extend(_check_knowledge_sources(agent_path, label, runner))
 
     return results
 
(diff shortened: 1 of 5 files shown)


solutions/ess-maker-skills/scripts/flightcheck/pva_client.py
```diff
@@ -0,0 +1,197 @@
...
+    import requests
+except ImportError:
+    print("ERROR: 'requests' package not found. Run: pip install requests")
+    sys.exit(1)
+
+
+CLIENT_ID = "51f81489-12ee-4a9e-aaae-a2591f45987d"
+
+# PVA app ID — used as the token audience for the Island Gateway API
+PVA_SCOPE = "96ff4394-9197-43aa-b393-6a41652e21f8/.default"
```
johnguy0, line 34: This GUID is the well-known Power Virtual Agents first-party app ID, used as the audience for the Island Gateway. Add a one-line comment naming it explicitly (something like `# 96ff4394-... is the well-known PVA app ID, used as the audience for the Island Gateway API`) so the next person doesn't have to chase the GUID through Microsoft docs to confirm it's not a typo or a copy-pasted secret.

solutions/ess-maker-skills/scripts/flightcheck/pva_client.py
```diff
@@ -0,0 +1,197 @@
...
+        # Extract the org name from our env_url for matching
+        # e.g., "https://orgb78b4a3b.crm.dynamics.com" → "orgb78b4a3b"
+        org_match = self.env_url.split("//")[1].split(".")[0] if "//" in self.env_url else ""
+
+        envs = resp.json().get("value", [])
+        for env in envs:
+            props = env.get("properties", {})
+            linked = props.get("linkedEnvironmentMetadata", {})
+            instance_url = linked.get("instanceUrl", "")
+            if org_match and org_match in instance_url:
```
johnguy0: Substring match on org name is fragile. If `org_match` is `org123`, this also matches an environment whose Dataverse `instanceUrl` contains `org1234contoso`, and you bind to the wrong env. Tighten to a hostname compare against `urlparse(self.env_url).hostname`, or at minimum require the match to start with `org_match + "."`.

solutions/ess-maker-skills/scripts/flightcheck/checks/local_files.py
```diff
@@ -287,3 +290,199 @@ def _check_template_configs(agent_path: Path, label: str) -> list[CheckResult]:
...
+        ))
+        return results
+
+    # Require the Island Gateway API for live status
+    pva = getattr(runner, "pva", None) if runner else None
+    bot_id = None
+    if runner and hasattr(runner, "config"):
+        bot_id = runner.config.get("agent", {}).get("botId")
+
+    if not pva or not hasattr(pva, "is_configured") or not pva.is_configured or not bot_id:
```
johnguy0: The `hasattr(pva, "is_configured")` guard isn't earning its keep. `PVAClient` always defines the property, so the only path where this is false is when `pva` is `None`, which is already covered by the `not pva` clause on the same line. Drop the `hasattr` check; the line gets shorter and the intent gets clearer.

johnguy0 (commented):
Repost looks clean against the old #57 patch, and the fixup for the moved CONTRIBUTING.md link target was the right call. The three round-1 items I flagged on #57 are still in the carried-over code, calling them out again here so they don't get lost on the way to merge. The CONTRIBUTING.md "no fabricated URLs" rule and the copilot-instructions section are good additions on their own.

Not quite ready for approval yet. Once these three are addressed in fixup commits this should be a quick re-review.

saengland:
Caught a stale `my/` reference in this branch. The convention John established in his PR #2 review (and reinforced 8+ times since, most recently PR #54) is `my/agents/{slug}/` -> `workspace/agents/{slug}/` for user-edited agent content. The scripts in this repo already use the new path everywhere (`fetch_and_setup.py:206`, `flightcheck/checks/local_files.py:35`, `setup.py:22`).

In `scripts/flightcheck/checks/local_files.py`:

```python
"sources by comparing filenames in my/agents/{slug}/knowledge/ against "
```

should be:

```python
"sources by comparing filenames in workspace/agents/{slug}/knowledge/ against "
```

This is user-facing remediation/result text, so anyone hitting this CONFIG-013 result gets pointed at a path that doesn't exist on disk. Cheap to fix in a fixup commit alongside John's three carry-over items from #57.

solutions/ess-maker-skills/scripts/flightcheck/pva_client.py
GrahamMcMynn: Done in ca09aea. Spelled out the GUID inline as the well-known Power Virtual Agents first-party app ID.

solutions/ess-maker-skills/scripts/flightcheck/pva_client.py
GrahamMcMynn: Done in f498ec8. Now parses the hostname out of `instanceUrl` with `urlparse` and requires it to start with `org_match + "."`, so `org123` no longer matches `org1234contoso`.

solutions/ess-maker-skills/scripts/flightcheck/checks/local_files.py
GrahamMcMynn: Done in a9be816. Dropped the redundant `hasattr` check; `not pva` already covers the only path where the attribute access could fail.

GrahamMcMynn:
Stale `my/agents/{slug}/knowledge/` reference fixed in 976f069  now uses `workspace/agents/{slug}/knowledge/` to match the convention in `fetch_and_setup.py`, `setup.py`, and the rest of `local_files.py`.

johnguy0 (commented):
All three round-1 follow-ups are now in: GUID named explicitly (ca09aea), hostname prefix match with urlparse (f498ec8), and the redundant hasattr guard dropped (a9be816). Spot-checked the new prefix logic with org_match='org123' against an instanceUrl of 'https://org1234contoso.crm.dynamics.com'; correctly does not match now. Also caught the my/agents/{slug}/ → workspace/agents/{slug}/ remediation string in 976f0696, which is the right call: the rest of local_files.py (lines 31/34/35/41/53) and fetch_and_setup.py / setup.py all use workspace/agents/, so the old remediation was pointing users at a path that doesn't exist on disk. Good catch on doing that proactively. Ready for approval from my side.

GrahamMcMynn (approved):

johnguy0 (commented):
Took another look after the merge from main came through. The new diff against the last review is just PR #65 (drop main2 from workflow triggers + doc cleanup) and PR #66 (sync_docs.py and sync_samples.py under solutions/ess-maker-skills/scripts/) folding in, both of which I already reviewed and signed off on separately. Nothing CONFIG-013-specific changed in this push. All three round-1 follow-ups (GUID name, hostname prefix match, dropped hasattr guard) are still resolved. Ready for approval from my side.

PR #64, opened 2026-05-08
Sync missing Workday samples (extended topics + scenario screenshots) [reposted from #60]

saengland:
Reposts #60 (Sri's Workday samples sync from upstream) onto the new main after today's cutover.

## What changed vs original

PR #60 added files under src/samples/workday/{employee,extended}/ which on new main lives under solutions/ess-maker-skills/src/examples/ess-samples/Workday/. Path mapping:
- src/samples/workday/employee/<Scenario>/ -> solutions/ess-maker-skills/src/examples/ess-samples/Workday/EmployeeScenarios/<Scenario>/
- src/samples/workday/extended/* -> solutions/ess-maker-skills/src/examples/ess-samples/Workday/ExtendedScenarios/*

NOTE: I introduced a new ExtendedScenarios/ folder under Workday/ rather than collapsing the 4 extended topics into peers of EmployeeScenarios/<Scenario>/. Sri, please confirm this layout is what you want; happy to flatten if you prefer.

All 17 files extracted with binary-safe `git show` to handle the 12 PNG screenshots. Authorship preserved via Co-authored-by trailer.

John flagged on #60: `RequestFeedback.yaml:292` mapping looks inverted vs the UI comments. Workday's `feedbackConfidential=true` means the feedback request stays private to the requester; current mapping appears to set the opposite. Recommend fix on this branch before merge.

Closes #60.

Co-authored-by: Srikanth Deshpande <srdeshpande@microsoft.com>


solutions/ess-maker-skills/src/examples/ess-samples/Workday/ExtendedScenarios/RequestFeedback.yaml
```diff
@@ -0,0 +1,393 @@
...
+      variable: Topic.responderName
+      value: =If(IsBlank(Topic.selectedResponderId), Topic.responderName, LookUp(Topic.parsedResponder.data, id = Topic.selectedResponderId).descriptor)
+
+    # Map feedbackSharing choice to Workday fields:
+    # "shareWithMe"      - feedbackConfidential=false, showFeedbackProviderName=true
+    # "shareWithManager" - feedbackConfidential=true,  showFeedbackProviderName=true
+    - kind: SetVariable
+      id: set_confidential
+      variable: Topic.feedbackConfidential
+      value: =If(Topic.feedbackSharing = "shareWithManager", "true", "false")
```
johnguy0: This mapping is inverted. In Workday, `feedbackConfidential=true` means the request stays private to the requester (only they see the feedback), and `false` means the manager also sees it. The comment block right above (lines 286-288) describes the correct semantics, so the data is right; the `If()` expression just flips them. Should be `=If(Topic.feedbackSharing = "shareWithMe", "true", "false")`. As written, picking "shareWithMe" today actually shares the feedback with the manager, which is the worst possible flavor of bug for a privacy-shaped field.

johnguy0 (commented):
Path translation looks right and the binary-safe extraction with authorship preserved is exactly the right shape for this repost, thanks for being careful with it.

The RequestFeedback.yaml blocker I flagged on #60 is still present in this branch (line 292), calling it out below so it doesn't get lost. On the layout question in the PR description: I'm fine with `ExtendedScenarios/` as a peer to `EmployeeScenarios/` rather than flattening, especially since the README in this folder treats it as a clearly separated set. No need to flatten.

Not quite ready for approval yet, the feedbackConfidential mapping needs to flip before this goes in.

saengland:
Pushed c2987ac addressing the only blocker on this PR.

**Fixed:** The `If()` on line 292 now reads `=If(Topic.feedbackSharing = "shareWithMe", "true", "false")` per your prescription. Selecting "shareWithMe" now correctly sets `feedbackConfidential=true` (private to the requester); selecting "shareWithManager" sets it to `false` (manager sees it).

**Also fixed:** The comment block at lines 286-288 had the same inversion as the code:

```yaml
# (before)
# "shareWithMe"      - feedbackConfidential=false, showFeedbackProviderName=true
# "shareWithManager" - feedbackConfidential=true,  showFeedbackProviderName=true
```

I read your review as "the comment describes the correct semantics, the `If()` flips them" -- but the comment text actually had the same flipped mapping as the broken code. Fixing only the code would have left the file internally inconsistent (code says one thing, comment says the opposite). Rewrote the comment to match your stated Workday semantics with an explicit "true=private, false=manager-sees" intro line so the next reader doesn't have to re-derive it:

```yaml
# (after)
# In Workday, feedbackConfidential=true means the request stays private to the
# requester; feedbackConfidential=false means the manager also sees the feedback.
# "shareWithMe"      - feedbackConfidential=true,  showFeedbackProviderName=true
# "shareWithManager" - feedbackConfidential=false, showFeedbackProviderName=true
```

Happy to revert the comment block if I misread your intent on what was correct -- but the values now match the `If()` and your description.

Ready for re-review.

solutions/ess-maker-skills/src/examples/ess-samples/Workday/ExtendedScenarios/RequestFeedback.yaml
johnguy0: Fix matches what I asked for. Comment block is now explicit about Workday semantics, which is the right call so the next reader does not have to re-derive the mapping. Thanks.

johnguy0 (commented):
Re-reviewed against the new commit. The feedbackConfidential flip is exactly the change I asked for, and the comment block is now explicit about Workday's semantics so the next reader does not have to re-derive it. Scope is appropriately tight, only the one mapping line and the comment block changed.

Good to merge from my side.

PR #78, opened 2026-05-13
tests: FlightCheck integration-test infrastructure + 9 captured cassettes

GrahamMcMynn:
## Summary

Builds an integration-test infrastructure for the FlightCheck readiness suite. Each FlightCheck check that calls an external API should have a captured cassette of real responses; tests then replay the cassette so the check's logic is validated against actual API shapes rather than hand-rolled mocks.

This PR establishes that foundation across eight commits:

| # | Commit | What it adds |
|---|---|---|
| 1 | `38ce436` | pytest scaffold + shared fixtures + markers (`vcr`, `live`) |
| 2 | `df2262e` | First VCR.py recording wrappers + redaction primitives |
| 3 | `d2c51ce` | Dataverse mock + 19 tests (surfaces 1 latent regex bug in `auth.discover_tenant`) |
| 4 | `ce073c5` | Pivot to end-to-end FlightCheck tests; first three checkpoints |
| 5 | `0f19870` | WD-CONN-001 E2E tests + PP Admin mock (surfaces 1 latent bug in `_get_all` 404 handling) |
| 6 | `09e7755` | Shared redaction infrastructure for the breadth of APIs |
| 7 | `083a169` | **9 FlightCheck cassettes** + their recorders + the `INDEX.md` registry + `AGENTS.md` guardrails |
| 8 | `df91fc4` | Mock placeholder modules for Graph / ServiceNow / Workday; bug-pin test for PP Admin client; cleanup of pre-pivot MCP test scaffolding |

## Cassette inventory (commit 7)

| Cassette | Size | Surface captured |
|---|---|---|
| `dataverse_whoami.yaml` | 1.6 KB | WhoAmI()  the env-id derivation chain's first call |
| `flightcheck_graph.yaml` | 57.7 KB | Graph: organization, users, directoryRoles, conditionalAccess policies |
| `flightcheck_pp_admin.yaml` | 745 KB | BAP environments, PowerApps flows, connections (+ intentional 404 for Dataverse-only env) |
| `flightcheck_workday.yaml` | 941 KB | Workday SOAP runtime data path (Get_Workers) |
| `workday_rest_admin.yaml` | 17 KB | Workday OAuth REST admin (refresh_token grant) |
| `workday_config.yaml` | 16 KB | Workday admin SOAP probes (Identity_Management) |
| `workday_wql_admin.yaml` | 1066 KB | WQL admin (6 data sources, full discovery flow) |
| `island_gateway_botcomponents.yaml` | 119 KB | PVA Island Gateway: bot list + full bot-component snapshot |
| `flightcheck_servicenow.yaml` | 20 KB | ServiceNow Knowledge connector setup validation (10 calls per MS Learn doc) |

All cassettes are redacted via the inline `_common.py` pipeline (GUIDs, tokens, emails, tenant identifiers, Workday WIDs, Workday PII elements/attributes) and hand-scrubbed for strings the global scrubber doesn't catch (customer-named bots, ServiceNow `user_name` fields).

## Documentation

- **`tests/fixtures/cassettes/INDEX.md`** (new)  the registry of confirmed-real endpoints. Future agents check this BEFORE writing any check that calls a network API.
- **`tests/AGENTS.md`** (new)  cardinal rule "no integration tests against APIs not confirmed real with a cassette". Three-place check workflow. Template message for requesting new captures. Includes the **WQL chicken-and-egg blocker note**  see below.

## Known blocker: Workday WQL bearer-only auth

The WQL cassette is committed as **discovery evidence**, not as the foundation for a runtime check. Workday's REST/WQL endpoints accept only OAuth 2.0 Bearer tokens. To validate "the customer's ESS API Client is registered correctly via WQL," FlightCheck would need its own OAuth API Client registered in the same Workday tenant  nearly the same workflow as the ESS Workday integration setup the check would validate. Net: WQL-backed checks push the setup-pain problem one level deeper rather than solve it. Four workarounds analyzed (authorization_code+PKCE, piggyback on ESS-shipped client, SOAP-with-WS-Security, drop the check entirely); none currently viable without Microsoft-side changes or further investigation. Decision documented in three places (`INDEX.md`, recorder docstring, `AGENTS.md`).

## Latent bugs surfaced

Three bugs in production code pinned by regression tests:
1. `auth.discover_tenant` regex over-captures (commit 3)
2. `pp_admin_client._get_all` crashes on HTTP 404 (commit 5, pin test in commit 8)
3. One more in `test_workday_connections.py` (see TODO note)

Each pinned bug has a `TODO: when fixed, flip this assertion to ...` so the test is the canary.

## What's not in this PR

- HRSD/ITSM ServiceNow cassette (different integration from the Knowledge connector; separate recording session)
- Any new FlightCheck check itself  this PR is the *foundation* the next agent uses
- Workday WQL-backed checks (blocked, see above)
- Microsoft-side validation for the ServiceNow Knowledge connector configuration in Microsoft Admin Center / Search & intelligence (would need its own Graph extension)

## Testing

```powershell
pip install -e .[test]
pytest tests/
```

Should be green. Live mode (`--run-live`) is opt-in for tests marked `@pytest.mark.live` and is gated by `require_validated_mock` in `conftest.py`.


tests/flightcheck/checks/test_workday_env_vars.py
```diff
@@ -0,0 +1,343 @@
...
+
+
+# Make the production source importable. (The pytest pythonpath in
+# pyproject.toml already does this, but spell it out so the test file
+# is also runnable in isolation via `python -m tests.flightcheck.checks.test_workday_env_vars`.)
+SCRIPTS_PATH = (
+    pytest.Path(__file__).resolve().parents[3]
+    if hasattr(pytest, "Path")
+    else None
+)
```
GrahamMcMynn: This block is dead code. `pytest` (the module) has no `Path` attribute, so `hasattr(pytest, "Path")` is always False  the conditional always picks the `else None` branch and the `pytest.Path(__file__).resolve().parents[3]` call never runs. The resulting `SCRIPTS_PATH` is also unused anywhere in the file.

The comment above it claims this makes the file runnable in isolation (`python -m tests.flightcheck.checks.test_workday_env_vars`), but nothing is actually added to `sys.path`  running it standalone will still fail with `ModuleNotFoundError: No module named 'flightcheck'` because nothing puts `solutions/ess-maker-skills/scripts` on the path.

If the intent was to actually support standalone execution, you probably wanted something like:

```python
import sys
from pathlib import Path
SCRIPTS_PATH = Path(__file__).resolve().parents[3] / "solutions" / "ess-maker-skills" / "scripts"
if str(SCRIPTS_PATH) not in sys.path:
    sys.path.insert(0, str(SCRIPTS_PATH))
```

Otherwise just delete lines 4653; the `pyproject.toml` `pythonpath` setting already covers the `pytest` invocation path.

tests/mocks/pp_admin.py
```diff
@@ -0,0 +1,362 @@
...
+) -> dict[str, Any]:
+    """Build a single PowerApps flow record.
+
+    `state` accepts "Started", "Stopped", "Suspended". The check in
+    flightcheck/checks/workday.py:_check_flow_status treats
+    {"started", "on", "enabled"} (case-insensitive) as enabled.
+    """
+    return {
+        "name": flow_id or "00000000-0000-0000-0000-000000007101",
+        "id": f"/providers/Microsoft.ProcessSimple/environments/{{env}}/flows/{flow_id}",
```
GrahamMcMynn: `flow()` falls back to a default GUID for `name` via `flow_id or "00000000-..."`, but the `id` template interpolates `flow_id` directly  so calling `pp.flow()` with no arguments yields:

```json
{
  "name": "00000000-0000-0000-0000-000000007101",
  "id":   "/providers/Microsoft.ProcessSimple/environments/{env}/flows/None",
  ...
}
```

The literal string `None` ends up in the URL path. Currently no test calls `pp.flow()`, so it doesn't manifest, but this helper is the template the next agent will reach for when writing the flow-status checks  they'll get a record whose `name` and `id` disagree, and any assertion that touches `id` will be silently misleading.

Suggested fix: resolve the fallback once.

```python
def flow(*, flow_id: str | None = None, ...):
    effective_id = flow_id or "00000000-0000-0000-0000-000000007101"
    return {
        "name": effective_id,
        "id": f"/providers/Microsoft.ProcessSimple/environments/{{env}}/flows/{effective_id}",
        ...
    }
```

(Also worth deciding whether the `{{env}}` literal placeholder should be a real `env_id` parameter  `list_flows()` takes `env_id` but `flow()` doesn't, so the record's `id` won't match the URL it's served under. Separate question, just flagging it.)

GrahamMcMynn (commented):
Code review by ESS Code Review Agent. Two low-severity issues found in the test infrastructure  see inline comments. Most of the diff (cassettes, redaction primitives, recorder scripts, mock builders, pinned-bug regression tests) looks solid; the AGENTS.md / INDEX.md guardrails and `require_validated_mock` rails are nicely defensive.

tests/captures/record_flightcheck_servicenow.py
```diff
@@ -0,0 +1,296 @@
...
+    headers = {"Accept": "application/json"}
+
+    if not instance.lower().startswith("https://"):
+        print(f"ERROR: SERVICENOW_INSTANCE_URL must use https:// (got {instance!r}).")
+        sys.exit(1)
+
+    print(f"  Instance: {instance}")
+    print(f"  Account:  {env['SERVICENOW_USERNAME']}")
+    if "SERVICENOW_OAUTH_REGISTRY_NAME" in env:
+        print(f"  OAuth registry name filter: {env['SERVICENOW_OAUTH_REGISTRY_NAME']!r}")
```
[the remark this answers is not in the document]
GrahamMcMynn: Fixed in 041cc4d.

Root cause was CodeQL's data-flow analysis tainting any object derived from a request whose kwargs include credentials (here, `auth=HTTPBasicAuth(username, password)`). The `print()` calls only ever emitted status codes / server response bodies / env var labels  none actually contained credentials  but CodeQL can't see through the taint.

Fix mirrors the pattern in production `solutions/.../checks/workday.py`:
1. Split `_check_env()` to return `(metadata, secrets)`  two separate dicts. Prints driven by `metadata` only.
2. In `_do()`, sanitize tainted response data (`status_code`, `body_text`, `body_json`) into clean local primitives BEFORE any print, then `del r` to drop the source. CodeQL drops the taint when the data flow is decoupled.

tests/captures/record_flightcheck_servicenow.py
```diff
@@ -0,0 +1,296 @@
...
+            ok = r.status_code == expect_status
+            marker = "OK " if ok else "!! "
+            print(f"  {marker}{label:55} -> {r.status_code} (expected {expect_status})")
+            if not ok and r.status_code in (400, 401, 403, 404):
+                # Surface the error message for diagnosis (helpful when
+                # admin role isn't elevated and oauth_entity returns 403).
+                try:
+                    err = r.json()
+                    msg = err.get("error", {}).get("message", "") if isinstance(err.get("error"), dict) else str(err)
+                    print(f"     body: {str(msg)[:200]}")
```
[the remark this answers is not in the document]
GrahamMcMynn: Fixed in 041cc4d.

Root cause was CodeQL's data-flow analysis tainting any object derived from a request whose kwargs include credentials (here, `auth=HTTPBasicAuth(username, password)`). The `print()` calls only ever emitted status codes / server response bodies / env var labels  none actually contained credentials  but CodeQL can't see through the taint.

Fix mirrors the pattern in production `solutions/.../checks/workday.py`:
1. Split `_check_env()` to return `(metadata, secrets)`  two separate dicts. Prints driven by `metadata` only.
2. In `_do()`, sanitize tainted response data (`status_code`, `body_text`, `body_json`) into clean local primitives BEFORE any print, then `del r` to drop the source. CodeQL drops the taint when the data flow is decoupled.

tests/captures/record_flightcheck_servicenow.py
```diff
@@ -0,0 +1,296 @@
...
+                except ValueError:
+                    print(f"     body[:200]: {r.text[:200]}")
```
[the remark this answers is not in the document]
GrahamMcMynn: Fixed in 041cc4d.

Root cause was CodeQL's data-flow analysis tainting any object derived from a request whose kwargs include credentials (here, `auth=HTTPBasicAuth(username, password)`). The `print()` calls only ever emitted status codes / server response bodies / env var labels  none actually contained credentials  but CodeQL can't see through the taint.

Fix mirrors the pattern in production `solutions/.../checks/workday.py`:
1. Split `_check_env()` to return `(metadata, secrets)`  two separate dicts. Prints driven by `metadata` only.
2. In `_do()`, sanitize tainted response data (`status_code`, `body_text`, `body_json`) into clean local primitives BEFORE any print, then `del r` to drop the source. CodeQL drops the taint when the data flow is decoupled.

tests/captures/record_workday_rest_admin.py
```diff
@@ -0,0 +1,518 @@
...
+                },
+            },
+        ),
+    ]
+
+    for label, kwargs in attempts:
+        try:
+            r = requests.post(token_url, timeout=30, **kwargs)
+        except requests.RequestException as exc:
+            print(f"  Try [{label}]: unreachable ({exc!s})")
```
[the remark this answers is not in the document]
GrahamMcMynn: Fixed in 041cc4d.

Same CodeQL false-positive class as the servicenow recorder, plus one real partial leak.

Refactor: factored a single `_post_token_attempts(token_url, attempts)` helper that consumes the credential-tainted Response object inside itself and yields only sanitized primitives  `(label, status_code, payload, body_text)`. All three `_acquire_token_*` functions now use it and never touch a raw Response object. Exception strings replaced with `type(exc).__name__` so there's no chance of leaking a request URL via `{exc!s}`.

Real partial leak (line 433): the startup banner used to print `starts with {refresh_token[:8]}...`  8 chars of the actual refresh token. Replaced with length-only display: `configured ({N} chars)`. Same diagnostic intent (confirms a token is set), zero token bytes.

tests/captures/record_workday_rest_admin.py
```diff
@@ -0,0 +1,518 @@
...
+            continue
+
+        print(f"  Try [{label}]: HTTP {r.status_code}")
```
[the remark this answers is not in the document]
GrahamMcMynn: Fixed in 041cc4d.

Same CodeQL false-positive class as the servicenow recorder, plus one real partial leak.

Refactor: factored a single `_post_token_attempts(token_url, attempts)` helper that consumes the credential-tainted Response object inside itself and yields only sanitized primitives  `(label, status_code, payload, body_text)`. All three `_acquire_token_*` functions now use it and never touch a raw Response object. Exception strings replaced with `type(exc).__name__` so there's no chance of leaking a request URL via `{exc!s}`.

Real partial leak (line 433): the startup banner used to print `starts with {refresh_token[:8]}...`  8 chars of the actual refresh token. Replaced with length-only display: `configured ({N} chars)`. Same diagnostic intent (confirms a token is set), zero token bytes.

tests/captures/record_workday_rest_admin.py
```diff
@@ -0,0 +1,518 @@
...
+                },
+            },
+        ),
+    ]
+
+    for label, kwargs in attempts:
+        try:
+            r = requests.post(token_url, timeout=30, **kwargs)
+        except requests.RequestException as exc:
+            print(f"  Try [{label}]: unreachable ({exc!s})")
```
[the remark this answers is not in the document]
GrahamMcMynn: Fixed in 041cc4d.

Same CodeQL false-positive class as the servicenow recorder, plus one real partial leak.

Refactor: factored a single `_post_token_attempts(token_url, attempts)` helper that consumes the credential-tainted Response object inside itself and yields only sanitized primitives  `(label, status_code, payload, body_text)`. All three `_acquire_token_*` functions now use it and never touch a raw Response object. Exception strings replaced with `type(exc).__name__` so there's no chance of leaking a request URL via `{exc!s}`.

Real partial leak (line 433): the startup banner used to print `starts with {refresh_token[:8]}...`  8 chars of the actual refresh token. Replaced with length-only display: `configured ({N} chars)`. Same diagnostic intent (confirms a token is set), zero token bytes.

tests/captures/record_workday_rest_admin.py
```diff
@@ -0,0 +1,518 @@
...
+            continue
+
+        print(f"  Try [{label}]: HTTP {r.status_code}")
```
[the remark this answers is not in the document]
GrahamMcMynn: Fixed in 041cc4d.

Same CodeQL false-positive class as the servicenow recorder, plus one real partial leak.

Refactor: factored a single `_post_token_attempts(token_url, attempts)` helper that consumes the credential-tainted Response object inside itself and yields only sanitized primitives  `(label, status_code, payload, body_text)`. All three `_acquire_token_*` functions now use it and never touch a raw Response object. Exception strings replaced with `type(exc).__name__` so there's no chance of leaking a request URL via `{exc!s}`.

Real partial leak (line 433): the startup banner used to print `starts with {refresh_token[:8]}...`  8 chars of the actual refresh token. Replaced with length-only display: `configured ({N} chars)`. Same diagnostic intent (confirms a token is set), zero token bytes.

tests/captures/record_workday_rest_admin.py
```diff
@@ -0,0 +1,518 @@
...
+                },
+            },
+        ),
+    ]
+
+    for label, kwargs in attempts:
+        try:
+            r = requests.post(token_url, timeout=30, **kwargs)
+        except requests.RequestException as exc:
+            print(f"  Try [{label}]: unreachable ({exc!s})")
```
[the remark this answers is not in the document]
GrahamMcMynn: Fixed in 041cc4d.

Same CodeQL false-positive class as the servicenow recorder, plus one real partial leak.

Refactor: factored a single `_post_token_attempts(token_url, attempts)` helper that consumes the credential-tainted Response object inside itself and yields only sanitized primitives  `(label, status_code, payload, body_text)`. All three `_acquire_token_*` functions now use it and never touch a raw Response object. Exception strings replaced with `type(exc).__name__` so there's no chance of leaking a request URL via `{exc!s}`.

Real partial leak (line 433): the startup banner used to print `starts with {refresh_token[:8]}...`  8 chars of the actual refresh token. Replaced with length-only display: `configured ({N} chars)`. Same diagnostic intent (confirms a token is set), zero token bytes.

tests/captures/record_workday_rest_admin.py
```diff
@@ -0,0 +1,518 @@
...
+            continue
+
+        print(f"  Try [{label}]: HTTP {r.status_code}")
```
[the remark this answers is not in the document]
GrahamMcMynn: Fixed in 041cc4d.

Same CodeQL false-positive class as the servicenow recorder, plus one real partial leak.

Refactor: factored a single `_post_token_attempts(token_url, attempts)` helper that consumes the credential-tainted Response object inside itself and yields only sanitized primitives  `(label, status_code, payload, body_text)`. All three `_acquire_token_*` functions now use it and never touch a raw Response object. Exception strings replaced with `type(exc).__name__` so there's no chance of leaking a request URL via `{exc!s}`.

Real partial leak (line 433): the startup banner used to print `starts with {refresh_token[:8]}...`  8 chars of the actual refresh token. Replaced with length-only display: `configured ({N} chars)`. Same diagnostic intent (confirms a token is set), zero token bytes.

tests/captures/record_workday_rest_admin.py
```diff
@@ -0,0 +1,518 @@
...
+    if soap_host != raw_host:
+        print(f"  Note: rewrote web host {raw_host} -> services host {soap_host}")
+    tenant = env["WORKDAY_TENANT_NAME"]
+    token_url = f"{soap_host}/ccx/oauth2/{tenant}/token"
+    rest_base = f"{soap_host}/ccx/api"
+
+    print(f"  Token endpoint: {token_url}")
+    print(f"  REST base URL:  {rest_base}/<module>/v<n>/{tenant}/<resource>")
+    if refresh_token:
+        print(f"  Refresh token:  configured (starts with {refresh_token[:8]}...)")
```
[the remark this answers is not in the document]
GrahamMcMynn: Fixed in 041cc4d.

Same CodeQL false-positive class as the servicenow recorder, plus one real partial leak.

Refactor: factored a single `_post_token_attempts(token_url, attempts)` helper that consumes the credential-tainted Response object inside itself and yields only sanitized primitives  `(label, status_code, payload, body_text)`. All three `_acquire_token_*` functions now use it and never touch a raw Response object. Exception strings replaced with `type(exc).__name__` so there's no chance of leaking a request URL via `{exc!s}`.

Real partial leak (line 433): the startup banner used to print `starts with {refresh_token[:8]}...`  8 chars of the actual refresh token. Replaced with length-only display: `configured ({N} chars)`. Same diagnostic intent (confirms a token is set), zero token bytes.

tests/test_captures_redaction.py
```diff
@@ -0,0 +1,431 @@
...
+        text = "https://orgcae533e6.crm.dynamics.com/api/data"
+        out = _redact_text(text)
+        assert "orgcae533e6" not in out
+        assert "orgmocktenant" in out
+
+    def test_replaces_servicenow_dev_instance_hostname(self) -> None:
+        text = '"instance":"https://dev184242.service-now.com/api"'
+        out = _redact_text(text)
+        assert "dev184242" not in out.lower()
+        assert "devmocktenant.service-now.com" in out
```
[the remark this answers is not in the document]
GrahamMcMynn: Fixed in 041cc4d.

These were test assertions, not security checks  `out` is the output of `_redact_text(known_test_input)`, and we're verifying the redactor produced the expected output. Zero security risk in the original code, but CodeQL doesn't distinguish.

Replaced `assert "<host>" in out` with `assert re.search(r"https?://<host>\b", out), ...`  requires the hostname to appear in URL position (preceded by `://`), satisfies CodeQL, and gains a more useful failure message.

tests/test_captures_redaction.py
```diff
@@ -0,0 +1,431 @@
...
+        text = '"instance":"Dev184242"'
+        out = _redact_text(text)
+        assert "Dev184242" not in out
+        assert "DevMockInstance" in out
+
+    def test_replaces_successfactors_tenant_url(self) -> None:
+        text = "https://apisalesdemo8.successfactors.com/odata/v2"
+        out = _redact_text(text)
+        assert "apisalesdemo8" not in out.lower()
+        assert "apisalesdemomock.successfactors.com" in out
```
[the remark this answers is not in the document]
GrahamMcMynn: Fixed in 041cc4d.

These were test assertions, not security checks  `out` is the output of `_redact_text(known_test_input)`, and we're verifying the redactor produced the expected output. Zero security risk in the original code, but CodeQL doesn't distinguish.

Replaced `assert "<host>" in out` with `assert re.search(r"https?://<host>\b", out), ...`  requires the hostname to appear in URL position (preceded by `://`), satisfies CodeQL, and gains a more useful failure message.

tests/flightcheck/checks/test_workday_env_vars.py
GrahamMcMynn: Good catch. Took your second suggestion (delete the block) since the standalone-runnability story was never going to work  the file imports `from tests.conftest import ...` which requires the repo root on sys.path AND requires pytest fixtures.

Fixed in 041cc4d: deleted lines 46-53 plus the now-unused `import sys`. `pytest pythonpath` in pyproject.toml already covers the test invocation.

tests/mocks/pp_admin.py
GrahamMcMynn: Fixed in 041cc4d. Took both parts:

Primary: resolved the `flow_id or ...` fallback once into `effective_id`, used for both `name` and `id`. Calling `pp.flow()` with no args now produces a record where `name` and the trailing path segment of `id` agree.

Secondary (the flagged-as-separate concern): added an `env_id` parameter defaulting to `MOCK_ENV_ID`  the same default `list_flows()` uses. Records built by `flow()` are now self-consistent with what `list_flows()` would serve them under; callers can pass the same env_id to both.

Smoke-tested:
```
pp.flow()                          -> id .../environments/Default-...0001111/flows/...0007101
pp.flow(flow_id='abc-123', env_id='env-xyz')  -> id .../environments/env-xyz/flows/abc-123
```

tests/captures/record_workday_rest_admin.py
```diff
@@ -0,0 +1,544 @@
...
+            token_url,
+            env["WORKDAY_OAUTH_CLIENT_ID"],
+            env["WORKDAY_OAUTH_CLIENT_SECRET"],
+            private_key_path,
+            refresh_token,
+        )
+        if not token:
+            print("  ABORT: no access token; not making admin API calls.")
+            return
+        print(f"  Step 1: OK (token starts with {token[:8]}...)")
```
GrahamMcMynn: The fix in this commit addressed the partial-credential-leak pattern for the **refresh** token (lines 455-459  switched from ``refresh_token[:8]`` to ``len(refresh_token)``), but the symmetric leak of the **access** token at line 489 is still present:

```python
print(f"  Step 1: OK (token starts with {token[:8]}...)")
```

The access token returned by ``_acquire_token`` is also a bearer credential (a Workday JWT) that grants API access until expiry. Printing the first 8 chars exposes the JWT header (which decodes to algorithm + key id and is identifying) to:
- terminal scrollback / shell history files
- any CI/recording log capture (e.g., a ``script(1)`` or ``tee`` of the recording session)
- shoulder-surfing in screen-share scenarios while running the recorder

The cassette body itself is safe  ``SCRUB_JSON_KEYS["access_token"] = "REDACTED_OAUTH_ACCESS_TOKEN"`` scrubs it on write  but the console output is not.

**Suggested fix** (mirrors what you did for the refresh token):
```python
print(f"  Step 1: OK (received {len(token)}-char access token)")
```

Bonus: this also matches the same CodeQL data-flow taint pattern that drove the ``_post_token_attempts`` refactor  ``token`` is derived directly from the credential-tainted ``payload`` dict inside the ``_acquire_token_*`` helpers, so a future scan run might re-flag this line for the same reason. Belt-and-braces to fix it now.

GrahamMcMynn (commented):
Code review by ESS Code Review Agent. One security-relevant finding: partial access-token leak in the Workday REST recorder console output, symmetric to the refresh-token leak you fixed in this commit.

tests/captures/record_workday_rest_admin.py
GrahamMcMynn: Fixed in 6fd681c. You're right -- and the same pattern existed in TWO other recorders I missed in the previous round:

* `record_workday_rest_admin.py:489` (the one you flagged)
* `record_workday_wql.py:231`
* `record_flightcheck_island_gateway.py:226`

All three now use your suggested fix:

```python
print(f"  Step 1: OK (received {len(token)}-char access token)")
```

Same diagnostic intent, zero token bytes. Catches the JWT header leak you described to terminal scrollback / CI capture / shoulder-surfing in the screen-share scenario, plus pre-empts the CodeQL re-flag you noted.

tests/captures/record_flightcheck_servicenow.py
```diff
@@ -0,0 +1,333 @@
...
+        # The OAuth registry holds the client_id/client_secret used by the
+        # M365 Copilot Connector. The doc requires the callback URL be
+        # `https://gcs.office.com/v1.0/admin/oauth/callback` (commercial)
+        # or the GCC equivalent. Active=true.
+        oauth_path = (
+            "oauth_entity?sysparm_query=type=external_client"
+            "&sysparm_fields=name,client_id,redirect_url,active,refresh_lifetime,"
+            "access_token_lifetime,sys_id&sysparm_limit=20"
+        )
+        if env.get("SERVICENOW_OAUTH_REGISTRY_NAME"):
```
GrahamMcMynn: **Latent `NameError`: `env` is not defined in `main()`.**

`main()` only binds `metadata, secrets = _check_env()` (line 159). The reference here:

```python
if env.get("SERVICENOW_OAUTH_REGISTRY_NAME"):
    name_filter = quote(metadata["oauth_registry_name"], safe="")
```

raises `NameError` whenever `SERVICENOW_OAUTH_REGISTRY_NAME` is set, before any network call. The branch is normally untaken, so the bug is latent — but the documented "narrow the OAuth registry lookup to a specific name" code path is dead/crashing. Looks like a leftover from the `env` → `(metadata, secrets)` refactor (the unused module-level `OPTIONAL_ENV` at line 110 is another leftover from the same refactor).

**Suggested fix:** replace with `os.environ.get("SERVICENOW_OAUTH_REGISTRY_NAME")`, or check `"oauth_registry_name" in metadata`. Consider also removing `OPTIONAL_ENV` if it's truly unused.

GrahamMcMynn (commented):
Code review by ESS Code Review Agent. 4 issues identified: 1 latent NameError in the ServiceNow recorder, 1 cassette-recording correctness bug, 1 PII-redaction gap on nested Workday elements, 1 testing-contract mismatch in `require_validated_mock`.

tests/captures/_common.py
```diff
@@ -0,0 +1,672 @@
...
+        print("ERROR: vcrpy not installed. Run: pip install -r requirements-dev.txt")
+        sys.exit(1)
+
+    CASSETTE_DIR.mkdir(parents=True, exist_ok=True)
+    cassette_path = CASSETTE_DIR / f"{name}.yaml"
+
+    return vcr.VCR(
+        cassette_library_dir=str(CASSETTE_DIR),
+        path_transformer=vcr.VCR.ensure_suffix(".yaml"),
+        record_mode="new_episodes",
```
GrahamMcMynn: **`record_mode="new_episodes"` + no pre-delete = stale cassettes silently replayed on re-record.**

```python
return vcr.VCR(
    ...
    record_mode="new_episodes",
    match_on=("method", "scheme", "host", "port", "path", "query"),
    ...
).use_cassette(str(cassette_path))
```

With `new_episodes`, vcrpy only records requests that don't match an existing interaction — every matching request is served from the cassette. Combined with no pre-delete of `cassette_path`, re-running any `record_*.py` script after credentials rotate, after a redaction-rule change, or after a backend response changes will quietly serve the **old** recorded responses. The operator sees `HTTP 200`, assumes a fresh cassette was written, but the request never went out and the cassette is unchanged for those endpoints.

`match_on` also doesn't include body, so token-exchange POSTs to the same path/query always match the first recorded interaction. Cassettes can also grow monotonically for any URL whose query string varies on each run.

**Suggested fix:** for recorder wrappers use `record_mode="all"` (always re-record), or `cassette_path.unlink(missing_ok=True)` before yielding the VCR. If "append more endpoints to an existing recording" really is the intended workflow, document that and pick `match_on` keys that distinguish each scenario.

tests/captures/_common.py
```diff
@@ -0,0 +1,672 @@
...
+        text = pat.sub(dst, text)
+    # Workday-specific XML element scrubbing. Done after the generic
+    # regex pass so that element contents take precedence over any
+    # patterns the generic pass might have partially matched inside
+    # element bodies.
+    for element_name, replacement in WORKDAY_PII_ELEMENTS.items():
+        # Match <prefix:ElementName ...>contents</prefix:ElementName>
+        # with any namespace prefix and any attributes.
+        pattern = re.compile(
+            rf"(<[\w-]+:?{element_name}(?:\s[^>]*)?>)[^<]*(</[\w-]+:?{element_name}>)",
```
GrahamMcMynn: **Workday PII scrubber `[^<]*` content class fails open on any element with nested children.**

```python
rf"(<[\w-]+:?{element_name}(?:\s[^>]*)?>)[^<]*(</[\w-]+:?{element_name}>)"
```

The `[^<]*` content class refuses to match if the element body contains any `<`, so any element whose body wraps further child elements is **not** redacted. Workday SOAP routinely nests inside named PII elements, e.g.:

```xml
<wd:Address_Line_Data>
  <wd:Effective_Date>2024-01-01</wd:Effective_Date>
  123 Real Street
</wd:Address_Line_Data>
```

or `<wd:Phone_Number><wd:Formatted>(425) 555-0123</wd:Formatted></wd:Phone_Number>`. In those shapes the named-element rule fails open and the inner PII text reaches the committed cassette. Same issue affects the `([^<]*)` content group in `_WORKDAY_TYPED_ID_PATTERN` at line 316.

The redaction tests in `tests/test_captures_redaction.py` only cover leaf elements (e.g. `<wd:First_Name>Sarah</wd:First_Name>`) — none of them nest a child inside a PII element, so CI can't see this gap.

**Suggested fix:** use non-greedy DOTALL — `(.*?)` with `re.DOTALL` — and/or recurse into child elements. Add a regression test that asserts PII is gone from `<wd:Address_Line_Data><wd:Effective_Date>x</wd:Effective_Date>123 Real St</wd:Address_Line_Data>`.

tests/conftest.py
```diff
@@ -0,0 +1,282 @@
...
+        if "live" in item.keywords:
+            item.add_marker(skip_live)
+
+
+def require_validated_mock(mock_module: Any) -> None:
+    """Fail-fast guard for FlightCheck integration tests.
+
+    Call at the top of any test that depends on a mock module to ensure
+    you haven't accidentally written a test against a placeholder mock.
+    Raises pytest.UsageError (which fails collection, not just one test)
```
GrahamMcMynn: **`require_validated_mock` docstring contract doesn't match the implementation.**

The docstring says:

> *Raises pytest.UsageError (which fails collection, not just one test)*

but the implementation at lines 268 / 277 calls `pytest.fail(..., pytrace=False)`, which raises `_pytest.outcomes.Failed`, **not** `UsageError`. Several test modules invoke `require_validated_mock(...)` at module top level (e.g. `tests/flightcheck/checks/test_workday_connections.py`, `test_workday_env_vars.py`, `tests/flightcheck/test_pp_admin_client.py`, `tests/scripts/test_auth.py`). When `Failed` propagates out of import it surfaces as a **collection error on that single module**, not the global usage-error abort the docstring promises. A maintainer who later relies on the docstring contract — e.g. assumes a placeholder mock will stop the whole suite — will be surprised.

Corroborating evidence: `tests/test_mock_validation_policy.py` catches `pytest.fail.Exception`, which is the actual type raised.

**Suggested fix:** pick one story. Either change the implementation to `raise pytest.UsageError(...)` (matches the documented intent and global-abort behavior), or update the docstring to describe the actual `pytest.fail` semantics.

srideshpande (dismissed):

tests/captures/record_flightcheck_servicenow.py
GrahamMcMynn: Fixed in 287831e. Replaced `env.get(...)` with `\oauth_registry_name\ in metadata` (the dict that already carries it) and removed the now-orphaned `OPTIONAL_ENV` constant.

tests/captures/_common.py
GrahamMcMynn: Fixed in 287831e. Switched `build_cassette` to `record_mode=\all\` plus `cassette_path.unlink(missing_ok=True)` before yielding the VCR. Re-runs after credential rotation / redaction-rule change / backend response change can no longer silently serve the prior recording.

tests/captures/_common.py
GrahamMcMynn: Fixed in 287831e. Replaced `[^<]*` with non-greedy `.*?` (`re.DOTALL` was already on) on both the named-element scrubber AND `_WORKDAY_TYPED_ID_PATTERN`. Added 3 regression tests in `tests/test_captures_redaction.py` covering the nested-child shapes you cited (`Address_Line_Data` wrapping `Effective_Date` + leaf text, `Phone_Number` wrapping `Formatted`, typed-ID wrapping a child element).

tests/conftest.py
GrahamMcMynn: Fixed in 287831e. Updated the docstring to describe the actual `pytest.fail(..., pytrace=False)` semantics (per-module collection error, not global `UsageError` abort). Implementation left as-is since the existing `test_mock_validation_policy.py` tests already match the impl, and per-module collection-error is sufficient enforcement for the placeholder-mock guardrail.

srideshpande (dismissed):

srideshpande (approved):

PR #87, opened 2026-05-15
flightcheck: add WD-ENV-101  Workday ISU username vs Entra UPN format alignment

### LINKED ISSUE #82: FlightCheck: Add check for Workday ISU username alignment with Entra UPN format

**Frequency:** This misconfiguration has been observed in 1 customer incident(s) (long-running BCBSA Workday integration failure).

## What should be checked

The ESS Workday integration requires that the Integration System User (ISU) and end-user identifiers in Workday align with the Entra UPN format (or whichever claim ESS sends). When tenants federate via Okta or Ping, customers frequently leave ISU usernames in legacy formats (e.g., short employee IDs), which prevents the ESS request from being matched to a Workday user.

This is distinct from existing issue #67 (ISU-to-role mapping); this check is specifically about the *username string format*.

## Suggested implementation

- **Checkpoint ID:** WD-ENV-101
- **Category:** Workday
- **Priority:** HIGH
- **Validation logic:** Pull the configured ISU username from the customer's environment variable (`EmployeeContextRequestAccountName`). Compare its format against the tenant's Entra UPN claim format. Warn if shapes differ (e.g., contains `@`? matches tenant verified domains?). For a sample of expected ESS users, fetch their `Get_Workers` response and verify the `Username` field equals their Entra UPN.

## Context

GrahamMcMynn:
Implements the FlightCheck check requested in #82.

**Checkpoint:** `WD-ENV-101`
**Category:** Workday
**Priority:** HIGH
**Conditional?** No (runs whenever Workday flows are detected)

**API tier:**
  - Dataverse: `documented` (existing `tests/mocks/dataverse.py`, MS Learn `environmentvariabledefinition` / `environmentvariablevalue` reference)
  - Microsoft Graph v1.0: `validatable` (existing `tests/mocks/graph.py`, CSDL `EntityType Name="organization"` + `https://learn.microsoft.com/graph/api/organization-get`)

**Mock source:** Reuses existing builders only — no new mock module, no registry update.

## What this check does

Pulls `EmployeeContextRequestAccountName` from Dataverse and compares its shape against the tenant's verified Entra domains read from Microsoft Graph (`runner.graph.get_organization()` → `verifiedDomains[].name`). Federated tenants (Okta, Ping, ADFS) frequently leave the ISU username in a legacy short-employee-id format that doesn't match the UPN claim ESS sends on each request, which prevents Workday from matching the incoming request to a Worker — a common misconfiguration on tenants that provisioned the ISU before adopting UPN-shaped service-account naming.

Heuristics:
- No `@` in the ISU username → **Warning** (legacy short-id format — the most-cited misconfiguration root cause).
- `@` present but the domain part isn't in the tenant's verified domains → **Warning** (could be legitimate cross-tenant federation; surface for the operator to confirm).
- `<localpart>@<verified-domain>` → **Passed**.

## Tests

`tests/flightcheck/checks/test_workday_isu_username_format.py` covers GOOD (UPN-shaped + verified-domain match, plus the initial `*.onmicrosoft.com` domain), BAD (no-`@` legacy short-ID scenario, unverified-domain), defensive (empty verifiedDomains, empty org record, Graph raises), and SKIPPED paths (no Dataverse token, no Graph client, ISU env var unset — defers to WD-ENV-001).

11 new tests pass; the existing 18 flightcheck tests still pass unchanged. Full repo suite: 113 passed, 8 skipped.

## Verification context

The source issue did not include a `## Verification` section, so the implementation pattern was inferred from the most-similar existing check: `_check_env_vars` in the same module (`checks/workday.py`) for the Dataverse half, and `checks/prerequisites.py` for the `runner.graph` usage pattern. Test file mirrors `tests/flightcheck/checks/test_workday_env_vars.py`.

End-to-end validated against a real tenant: ran flightcheck against a Workday-equipped environment with the Dataverse env var both UPN-shaped (PASSED) and deliberately set to a legacy short-id value (WARNED with the expected remediation).

## What was deferred (and why)

The deeper `Get_Workers User_ID == Entra UPN` per-Worker comparison from the suggested implementation is intentionally **deferred** to a follow-up:

1. The Workday SOAP mock module (`tests/mocks/workday.py`) is `MOCK_STATUS = "placeholder"` and would need promotion to `validated` against `flightcheck_workday.yaml` before any test could exercise that surface (per the conftest enforcement).
2. "A sample of expected ESS users" requires a curated mapping from Workday Employee_IDs to Entra UPNs that no current FlightCheck config supplies (the existing `_check_workflows` only exercises a single `WORKDAY_TEST_EMPLOYEE_ID`).
3. The static format check landed here already addresses the legacy-ISU root cause (short-employee-id ISU format on a federated tenant) — a future `WD-WF-NNN` can wire the per-Worker comparison once the Workday mock is promoted and the sample-list input is formalised.

## Additional fixes piggy-backed on this PR

While validating WD-ENV-101 end-to-end against a real tenant, three latent bugs in the original May 7 Python-scripts commit surfaced — each one blocked the new check (or other checks) from running on most tenants. They are fixed in this PR because the new WD-ENV-101 check could not be validated without them.

1. **PP Admin env id derivation** (`pp_admin_client.py` + `cli.py`) — `derive_environment_id()` returned the Dataverse OrgId rather than the BAP env id (different GUIDs on most tenants), so every BAP-scoped check (ENV-001/002/003, EXT-001, flow listing, connection enumeration) 404'd. Now uses an instance-URL match against the BAP env list, mirroring `pva_client._discover_gateway()`.
2. **Flow listing endpoint** (`pp_admin_client.py`) — `get_flows()` was hitting `api.powerapps.com/.../v2/flows` (404, path doesn't exist there) instead of `api.flow.microsoft.com/.../v2/flows` with the required `service.flow.microsoft.com//.default` audience token. Adds `FLOW_BASE`, `FLOW_SCOPE`, dual-token `authenticate()`, and `use_flow_token=True` plumbing.
3. **License SKU matching** (`checks/prerequisites.py`) — PRE-001/002/003 used case-sensitive `in` checks against UPPER_CASE patterns while Graph returns mixed-case values (`Microsoft_365_Copilot`); also missed Teams bundles (`O365_BUSINESS_PREMIUM`, `ENTERPRISEPACK`, etc.). Added `_sku_matches()` helper + reference-list-grounded SKU tuples.

Test-side follow-up (`tests/mocks/pp_admin.py`, `tests/flightcheck/test_pp_admin_client.py`, `tests/fixtures/cassettes/INDEX.md`): aligned with the corrected Power Automate endpoint. The previous "bug 4" regression test pinning a 404 from `api.powerapps.com` as "expected for Dataverse-only environments" was a misdiagnosis — the 404 was a wrong-URL artefact — and has been removed.

A companion PR (#95) documents the corresponding **guidance updates** to `flightcheck/AGENTS.md` so future agents don't reintroduce the same classes of bug.

Fixes #82

### DIFF
--- a/solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py
+++ b/solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py
@@ -307,7 +307,9 @@ def _check_isu_username_format(runner) -> list[CheckResult]:
 
     # ── Step 1: read the ISU env var from Dataverse. We do this FIRST,
     # before consulting Graph, because the no-`@` legacy-format detection
-    # (the most-cited misconfiguration root cause — the BCBSA scenario)
+    # (the most-cited misconfiguration root cause — legacy short-ID ISU
+    # provisioning on federated tenants, common where the ISU was set
+    # up before the tenant adopted UPN-shaped service-account naming)
     # can be reported off the Dataverse value alone.
     try:
         sys.path.insert(0, os.path.join(os.path.dirname(__file__), "..", ".."))
--- a/solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py
+++ b/solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py
@@ -160,6 +160,9 @@ def run_workday_checks(runner) -> list[CheckResult]:
     # --- Environment Variables ---
     results.extend(_check_env_vars(runner))
 
+    # --- ISU username vs Entra UPN format alignment ---
+    results.extend(_check_isu_username_format(runner))
+
     # --- Connection References ---
     results.extend(_check_connections(runner))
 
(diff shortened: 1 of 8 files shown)


solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py
```diff
@@ -259,6 +262,203 @@ def _check_env_vars(runner) -> list[CheckResult]:
...
+    if not env_url or not dv_token:
+        results.append(CheckResult(
+            checkpoint_id="WD-ENV-101", category="Workday",
+            priority=Priority.HIGH.value, status=Status.SKIPPED.value,
+            description="ISU username vs Entra UPN format alignment",
+            result="Dataverse token not available — skipping ISU format check",
+        ))
+        return results
+
+    graph = getattr(runner, "graph", None)
```
GrahamMcMynn: This early Graph guard makes the check miss the highest-value failure mode. Detecting a legacy short-ID ISU (`'@' not in isu_value`) only needs the Dataverse env var, but when Graph auth is unavailable this returns `Skipped` before reading the ISU value. That means a customer with `EmployeeContextRequestAccountName=ISU12345` and a failed/skipped Graph sign-in gets no WD-ENV-101 warning for the BCBSA-style misconfiguration. Please read the ISU value first and emit the no-`@` warning before requiring `runner.graph`; only require Graph for the verified-domain comparison, and add a regression test for no-`@` with `graph=None`.

GrahamMcMynn (commented):

solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py
GrahamMcMynn: Fixed in be61b5d. Reordered the control flow so the ISU env var is read first; the no-@ legacy-format detection now WARNs off the Dataverse value alone (regardless of Graph availability), matching the reviewer's recommendation. Graph is only required for the verified-domain comparison on UPN-shaped values, and produces a SKIPPED with a partial-result message when unavailable.

Added regression test 	est_warns_on_legacy_isu_format_even_when_no_graph_client and renamed the existing graph-missing test to clarify it only covers the UPN-shaped path. All 29 flightcheck tests + 113 full-suite tests still pass.

solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py
GrahamMcMynn, line 1: The customer name BCBSA is in the PR.  Please remove all instances of customer names.

solutions/ess-maker-skills/scripts/flightcheck/cli.py
```diff
@@ -154,6 +146,19 @@ def main():
...
         print("  (Some checks will be skipped)")
+        pp_admin = None
+
+    # Derive the BAP environment ID. This MUST run after pp_admin is
+    # authenticated: the correct id comes from the BAP env list
+    # (matched on linkedEnvironmentMetadata.instanceUrl), not from the
+    # Dataverse WhoAmI OrganizationId (which is a different guid for
+    # almost every tenant — see derive_environment_id docstring).
+    print("Deriving Power Platform environment ID...")
+    env_id = derive_environment_id(env_url, dv_token, pp_admin=pp_admin)
```
srideshpande, line 166: Q: Why do we want to call `derive_environment_id` when pp_admin is none? should we first check if it is not none and only then call this method?

solutions/ess-maker-skills/scripts/flightcheck/cli.py
```diff
@@ -154,6 +146,19 @@ def main():
...
+    if env_id:
+        print(f"Environment ID: {env_id}")
+    else:
+        print("WARNING: Could not derive environment ID. Some checks may be limited.")
```
srideshpande: If we first check pp_admin value, we can print a clearer message, helps with debugging I suppose. what do you think?

If pp_admin is not none, then if the env_id is still null then should we error out? is that a possibility at all?

solutions/ess-maker-skills/scripts/flightcheck/cli.py
GrahamMcMynn: Great question, answering both at once.

`pp_admin is None` only means MSAL token acquisition for the PowerApps audience failed entirely (sign-in error, cancelled browser, network failure, MSAL internal error)  set on cli.py:149 in the `except` block of `pp_admin.authenticate()`. It does **not** mean `user lacks the Power Platform Administrator role`  that check happens later, per-API: BAP admin-scope endpoints return 401/403 if the signed-in user isn't a PP admin, and `_get()` traps those at pp_admin_client.py:171-172 and returns `{"_error": "insufficient_permissions"}`. `pp_admin` itself is a valid client either way.

`derive_environment_id` accepts `pp_admin=None` deliberately. Its documented Dataverse `WhoAmI/OrganizationId` fallback (wrong as a BAP id, but tolerated by checks that don't strictly need the BAP id) lets a substantial fraction of FlightCheck still run when Power Platform sign-in is broken:

- `PRE-*`  M365 Copilot / Copilot Studio / Teams license SKU checks (Graph, different token)
- `AUTH-*`  Graph health
- `WD-ENV-*`  Workday env var / ISU format (Dataverse only)  includes the new WD-ENV-101 from this PR
- `WD-WF-*`  Workday SOAP runtime tests against the live tenant (uses Workday creds, not Power Platform)
- `CONFIG-*`  local agent / topic / variable / knowledge source

That's roughly half of FlightCheck, and it's often the first stuff an operator needs to fix anyway. Erroring out at env_id derivation would block all of it behind a Power Platform sign-in problem the operator may not be able to fix in the moment.

Agreed the *message* was too vague though. Pushed 60bd983 which:

1. Adds an inline comment above the call documenting the design intent so future readers don't ask the same question.
2. Splits the diagnostic into three distinct outcomes so the operator knows what happened and which checks to expect:
   - `Environment ID: <id>` (BAP-resolved, all good)
   - `Environment ID: <id> (Dataverse OrganizationId fallback  Power Platform sign-in failed, so BAP-scoped checks (ENV-*, EXT-*, WD-CONN-*) will be skipped)`
   - `WARNING: Could not derive environment ID for <env_url>. BAP-scoped checks (ENV-*, EXT-*, WD-CONN-*) will be skipped; license, auth, Workday env-var, Workday SOAP, and local-file checks will still run.`

All 113 full-suite tests + 8 expected skips still pass.

solutions/ess-maker-skills/scripts/flightcheck/cli.py
GrahamMcMynn: Combined answer with the related thread above  short version: `pp_admin is None` means MSAL sign-in failed entirely (not `no PP Admin role`; that's caught later via per-API 401/403). We deliberately don't error out because PRE-* (license), AUTH-*, WD-ENV-*, WD-WF-* (Workday SOAP runtime), and CONFIG-* (local files) checks  roughly half of FlightCheck  still run without `pp_admin` and are often the first things an operator needs to fix.

60bd983 splits the diagnostic into three cases (BAP-resolved / WhoAmI fallback / both-failed) so the operator knows which checks to expect  addresses the `clearer message` ask. See the resolved thread on cli.py:157 for the full rationale and the design intent of `derive_environment_id` accepting `pp_admin=None`.

srideshpande (approved):

PR #88, opened 2026-05-15
flightcheck: add WD-CONN-101  Workday connection token / grant health (deep)

### LINKED ISSUE #81: FlightCheck: Add check for Workday SOAP connection OAuth token health and expiry

**Frequency:** This misconfiguration has been observed in 1 customer incident(s) (intermittent HTTP 400 to end users; visible to customer leadership).

## What should be checked

Today's Workday connection check (`WD-CONN-001+`) only verifies that the connection is in `Connected` state. It does not detect expired or revoked OAuth tokens until users start failing. Add a deeper check that validates the OAuth token is current, not nearing expiry, and that a recent test call against the Workday endpoint succeeds.

## Suggested implementation

- **Checkpoint ID:** WD-CONN-101
- **Category:** Workday
- **Priority:** HIGH
- **Validation logic:** Use Power Platform Admin API to enumerate connection references for the Workday SOAP API and inspect `statuses[].status` and `statuses[].error`. For each reference, attempt a low-cost SOAP call (e.g., `Get_Server_Timestamp` or a minimal `Get_Workers` with `count=1`) and report 401/403 explicitly. Warn if token last-refreshed timestamp is older than the customer's IdP refresh-token lifetime.

## Context

GrahamMcMynn:
Implements the FlightCheck check requested in #81.

**Checkpoint:** `WD-CONN-101`
**Category:** Workday
**Priority:** HIGH
**Conditional?** No (runs whenever Workday flows are detected, like the rest of `run_workday_checks`)

**API tier:** validated
**Mock source:**
  - Cassette: `tests/fixtures/cassettes/flightcheck_pp_admin.yaml`
  - Mock builder: `tests/mocks/pp_admin.py` (`MOCK_STATUS = "validated"`)
  - Same endpoint as WD-CONN-001 (`GET /providers/Microsoft.PowerApps/scopes/admin/environments/{env_id}/connections`)  already in the confirmed-cassette table in `tests/fixtures/cassettes/INDEX.md`, so no new cassette / no registry change required.

**What WD-CONN-101 adds beyond WD-CONN-001:**

WD-CONN-001 reports whether each Workday connection is in `Connected` state. WD-CONN-101 inspects the structured `statuses[0].error.{code,message}` block the PowerApps API returns on non-Connected statuses and surfaces:

* The connection display name and the connection owner's account (`properties.accountName`)
* The Entra error code verbatim (e.g. `AADSTS50173`, `AADSTS70008`, `AADSTS50058`, `AADSTS50076`)
* A pinned remediation hint per known error code, telling the operator *what* expired and *what to do* (re-auth in Power Platform, complete MFA, etc.)

The customer-facing failure mode this catches: a Workday SOAP connection that worked yesterday silently flips to 401/403 at runtime because the connection creator's Entra grant lapsed. Today WD-CONN-001 says "Error / re-authenticate"; WD-CONN-101 names the user, the grant code, and what specifically expired.

**Tests:** `tests/flightcheck/checks/test_workday_connection_token_health.py`  10 new tests, all passing locally.

* GOOD: all Workday connections Connected  Passed
* BAD: AADSTS50173 grant-expired  Failed with grant-expired hint
* BAD: AADSTS70008 refresh-token-inactivity  Failed with inactivity hint
* BAD: unrecognized AADSTS code  Failed with generic re-auth hint
* MIXED: one healthy + one expired  Failed, only the expired one in the remediation
* EDGE: no Workday connections  NotConfigured
* EDGE: only non-Workday connections (Office 365 etc.)  NotConfigured
* EDGE: 403 from BAP  pinned to current buggy NotConfigured behavior (shares root cause with WD-CONN-001's already-pinned `test_403_from_bap_is_misreported_as_not_configured` regression test; will flip together when `PPAdminClient._get_all` is fixed)
* EDGE: missing `env_id`  empty list, no crash, no request
* EDGE: missing `pp_admin` client (auth failed earlier)  empty list, no crash

Existing 35 flightcheck tests still pass (45 passed, 8 skipped  same skip count as baseline; the skips are tier-policy non-applicable in `test_mock_validation_policy.py`).

**Scope notes  parts of the issue that are intentionally NOT in this PR:**

The issue's "Suggested implementation" had three parts. Only part 1 fits the kit's actual architecture:

| # | Suggested | Status |
|---|---|---|
| 1 | Use PP Admin API to enumerate Workday connection refs and inspect `statuses[].status` and `statuses[].error` |  Implemented as WD-CONN-101 |
| 2 | Attempt a low-cost SOAP probe (`Get_Server_Timestamp` / minimal `Get_Workers`) and report 401/403 explicitly |  Already covered by the existing WD-WF-* checks, which call 17 real ESS workflows against the live Workday tenant when ISU credentials are supplied. WD-CONN-101 stays offline-only against the BAP cassette to avoid duplicating WD-WF-001 and to keep the check fast in CI. |
| 3 | Warn if "OAuth token last-refreshed timestamp" is older than the customer's IdP refresh-token lifetime |  Two blockers: (a) the runtime ESS Workday integration uses WS-Security UsernameToken (Basic auth via ISU username + password), so there is no Workday-side OAuth/refresh token to inspect; (b) the PowerApps connections API does not expose a `lastRefreshedTimestamp` field on connection records (verified against the validated cassette in `flightcheck_pp_admin.yaml`). Documented in the WD-CONN-101 docstring as a follow-up that would need a different API surface (e.g., Entra Sign-In Logs via Graph). |

Both deferrals are documented in the check's docstring + the source comments so a future agent or reviewer hitting the same issue won't re-litigate them.

Fixes #81

### DIFF
--- a/solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py
+++ b/solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py
@@ -607,7 +607,17 @@ TOKEN_HEALTH_ERROR_CODES = {
     "AADSTS50076": "MFA challenge required. Have the connection owner re-authenticate and complete MFA.",
     "Unauthorized": "Power Platform marked the credential unauthorized. Re-authenticate the connection.",
     "Unauthenticated": "Connection is not authenticated. Have the connection owner sign in to Power Platform.",
+    "ConfigurationNeeded": "Connection was created but never fully configured (required parameter missing). Either finish setup in Power Platform or delete the unbound connection.",
 }
+
+# Error-code values from TOKEN_HEALTH_ERROR_CODES that indicate the
+# connection was created but never configured (vs. lapsed auth on a
+# previously-working connection). Used by _classify_token_health_error
+# to set the severity/remediation class.
+_CONFIG_ERROR_CODES = frozenset({"ConfigurationNeeded"})
+# Error-code values that indicate a previously-working auth that has
+# lapsed and needs the owner to re-authenticate.
+_AUTH_ERROR_CODES = frozenset({"Unauthorized", "Unauthenticated"})
 
 # Matches Entra error-code identifiers like "AADSTS50173" or "AADSTS700082"
 # anywhere in a string. Anchored to a 5-7 digit AADSTS prefix to avoid
@@ -617,10 +627,12 @@ TOKEN_HEALTH_ERROR_CODES = {
 _AADSTS_CODE_RE = re.compile(r"\b(AADSTS\d{5,7})\b")
 
 
-def _classify_token_health_error(conn: dict) -> tuple[str | None, str | None, str]:
+def _classify_token_health_error(
+    conn: dict,
+) -> tuple[str | None, str | None, str, str]:
     """Inspect ``statuses[0].error`` on a connection record and return
-    ``(reported_code, reported_message, hint)`` for token-health
-    classification.
+    ``(reported_code, reported_message, hint, severity_class)`` for
+    token-health classification.
 
     The PowerApps connections API includes a structured ``error`` block
     on non-Connected statuses (target, code, message). The actual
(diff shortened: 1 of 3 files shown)


GrahamMcMynn (commented):
Finding: WD-CONN-101 maps only `statuses[0].error.code`, but the validated PowerApps cassette shape for this token failure has `error.code` = `Unauthorized` and the actionable Entra code only inside `error.message` (for example the cassette's AADSTS50173 case). With the current lookup on lines 392-397, live records in that shape will produce the generic Unauthorized remediation instead of the PR's promised AADSTS50173/AADSTS70008-specific hint. Please extract an `AADSTS\d+` code from `message` when `code` is a wrapper such as `Unauthorized` (and add a test using `error_code="Unauthorized"` plus an AADSTS message) before mapping to `TOKEN_HEALTH_ERROR_CODES`.

GrahamMcMynn:
Addressed in 3c83ed4: `_classify_token_health_error()` now scans `error.message` for `\bAADSTS\d{5,7}\b` first and prefers that code over `error.code`, falling back to `error.code` only when the message has no AADSTS code (which catches the `Unauthenticated` "never-signed-in" shape also captured in the cassette). Tests rewritten to use the realistic production shape (`code="Unauthorized"` + AADSTS in message); added a new test for the `Unauthenticated` fallback path. 11 tests pass; full flightcheck suite still green (46 passed, 8 skipped  same baseline).

amilandi (approved):

PR #90, opened 2026-05-15
flightcheck: add AUTH-005  Workday Enterprise App user assignment

### LINKED ISSUE #79: FlightCheck: Add check for Entra App registration security group on Workday Enterprise Application

**Frequency:** This misconfiguration has been observed in 1 customer incident(s) (Sev 2 deployment failure impacting ~3,000 users).

## What should be checked

Verify that the Microsoft Entra App registration backing the Workday Enterprise Application has the required user/security group assignment. Without this, the OBO/OAuth handshake on first agent access fails for end users, blocking the entire ESS deployment.

Checked from the customer's tenant:

- Locate the Entra Enterprise Application that fronts the Workday SOAP connector.
- Confirm "User assignment required?" is set to `Yes` and that the security group of intended ESS users is assigned to the application.
- If "User assignment required?" is `No`, warn that a deploy-time check cannot guarantee per-user access at runtime.

## Suggested implementation

- **Checkpoint ID:** AUTH-005
- **Priority:** CRITICAL
- **Validation logic:** Use Microsoft Graph `GET /servicePrincipals/{id}` (filter by displayName / appId for the Workday Enterprise Application)  check `appRoleAssignmentRequired`. Then `GET /servicePrincipals/{id}/appRoleAssignedTo`  confirm at least one assignment that targets the expected ESS user security group.

## Context

GrahamMcMynn:
Implements the FlightCheck check requested in #79.

**Checkpoint:** `AUTH-005`
**Category:** Authentication
**Priority:** Critical
**Conditional?** No  runs for every tenant; SKIPS cleanly if the Workday Enterprise App isn't provisioned yet.

**API tier:** validatable
**Mock source:**
- Microsoft Graph CSDL  `https://graph.microsoft.com/v1.0/$metadata` EntityType `servicePrincipal` (`appRoleAssignmentRequired` is Edm.Boolean) + EntityType `appRoleAssignment` (`principalType` is Edm.String, values `User`/`Group`/`ServicePrincipal`).
- MS Learn  `https://learn.microsoft.com/graph/api/serviceprincipal-get` and `https://learn.microsoft.com/graph/api/serviceprincipal-list-approleassignedto`. Example responses copied verbatim into the new mock builder docstrings in `tests/mocks/graph.py`.

**What it checks**
1. `GET /servicePrincipals?$filter=startswith(displayName,'Workday')`  locate the Workday Enterprise Application(s) in the tenant.
2. For each match, read `appRoleAssignmentRequired`:
   - `False`  **WARNING** ("User assignment required?" is set to No; deploy-time check cannot guarantee per-user access at runtime  recommend setting it to Yes and assigning the ESS user security group).
   - `True`  `GET /servicePrincipals/{id}/appRoleAssignedTo`:
     - No assignments  **FAILED** with the issue's customer scenario in the result text and Entra  Enterprise Applications  Users-and-groups in the remediation. This is the Sev 2 case AUTH-005 was filed to catch.
     - `Group` principal assigned  **PASSED** (lists up to three group display names).
     - Only `User` principals  **WARNING** (works today but doesn't scale to new ESS users).
3. No Workday SP in tenant  **SKIPPED** with remediation pointing at the Entra gallery install step.

**Tests**  `tests/flightcheck/checks/test_authentication.py` (new file). Covers:
- GOOD  group assigned  PASSED.
- BAD  no assignments  FAILED with remediation.
- WARNING  `appRoleAssignmentRequired = false`.
- WARNING  only individual users assigned (no group).
- SKIPPED  no Workday SP in tenant.
- SKIPPED  no Graph client at all (auth opted out).
- Multi-SP fan-out  tenant has both `Workday SSO` and `Workday OAuth` registered, each evaluated independently.

All 7 new tests pass; 110/110 of the pre-existing test suite still passes.

Fixes #79


solutions/ess-maker-skills/scripts/flightcheck/checks/authentication.py
```diff
@@ -131,6 +134,277 @@ def run_authentication_checks(runner) -> list[CheckResult]:
...
+        except Exception as e:
+            warning_items.append((
+                sp_name,
+                f"unable to list assigned users/groups ({e})",
+                "Re-run FlightCheck with a Graph token that holds "
+                "Application.Read.All or Directory.Read.All.",
+            ))
+            continue
+
+        if not assignments:
```
GrahamMcMynn, line 403: **Permissions error silently becomes a critical FAILED row.**

`graph.get_app_role_assignments(sp_id)` routes through `GraphClient.get_all()`, which on 401/403 returns the partial list silently (see `graph_client.py`:

```python
if resp.status_code in (401, 403):
    return items  # partial results
```

Your new method's own docstring acknowledges this: *"Returns `[]` on 401/403 (handled by `get_all`) so caller can branch on insufficient permissions."* But the caller here doesn't branch. The `try/except` above only catches raised exceptions; a 401/403 returns `[]` without raising, falls into this `if not assignments:` branch, and emits a `Critical / FAILED` row stating *"user assignment required, 0 users/groups assigned"* with remediation telling the operator to add a security group.

**Concrete failure scenario:** an auditor's token has `Application.Read.All` sufficient to list service principals (so we get past the SP lookup and find the Workday SP), but Conditional Access or a scoped directory role denies `appRoleAssignedTo` for that specific SP. The check then false-alarms a Sev-2-shaped finding telling the operator to fix an assignment that's already correct — exactly the wrong direction for a check whose own intro says it was filed to catch a real Sev 2.

**Fix — two reasonable options:**

1. Use `graph.get()` here instead — it returns `{"_error": "...", "_status": 401|403}` which you can detect explicitly and bucket as a WARNING with the permission remediation (same shape as the `except Exception` arm above).
2. Have `get_app_role_assignments` raise on 401/403 (e.g. wrap `get_all` and re-raise) so the existing `except` block catches it and produces the WARNING path you already wrote.

Worth adding a test that mocks a 403 on `/appRoleAssignedTo` and asserts the result is `WARNING` (permission), not `FAILED` (assignment missing).

GrahamMcMynn (commented):

solutions/ess-maker-skills/scripts/flightcheck/checks/authentication.py
GrahamMcMynn: Fixed in e56fef2.

The fix lives at the helper level (`graph_client.get_app_role_assignments` now probes via `graph.get()` first and raises `PermissionError` on 401/403) so any future caller is safe — not just AUTH-005. AUTH-005''s exception handler gets a dedicated `except PermissionError` arm that routes to WARNING with permission-specific remediation (mention Application.Read.All + Conditional Access scope), distinct from the generic Exception fallback.

Added regression test `test_403_on_appRoleAssignedTo_is_warning_not_failed` that mocks a 403 on `/servicePrincipals/{id}/appRoleAssignedTo` and asserts the result is WARNING (permission) and NOT FAILED with the "0 users/groups assigned" framing.

solutions/ess-maker-skills/scripts/flightcheck/graph_client.py
```diff
@@ -195,6 +195,43 @@ def get_service_principals(self, filter_expr: str = "") -> list:
...
+        uses the same defensive pattern via a ``graph.get()`` probe.
+
+        Docs: https://learn.microsoft.com/graph/api/serviceprincipal-list-approleassignedto
+        """
+        path = f"/servicePrincipals/{sp_id}/appRoleAssignedTo"
+        # Probe permission before delegating to get_all() — get_all
+        # silently returns [] on 401/403, which would let callers
+        # mistake "denied" for "no assignments exist." graph.get()
+        # preserves the _status field so we can branch explicitly.
+        probe = self.get(path, params={"$top": "1"})
```
qazizaahirah: nit: the probe fetches the first page, then get_all re-fetches from scratch, should we let get_all handle the status. minor performance concern

solutions/ess-maker-skills/scripts/flightcheck/checks/authentication.py
```diff
@@ -131,6 +134,302 @@ def run_authentication_checks(runner) -> list[CheckResult]:
...
+# ─────────────────────────────────────────────────────────────────────
+# AUTH-005 — Workday Enterprise App user assignment.
+# ─────────────────────────────────────────────────────────────────────
+
+
+# Display-name prefix(es) we recognize as the Workday Enterprise App
+# in a customer tenant. Most customers leave the SSO gallery name
+# ("Workday") in place; some prepend "Workday SSO" / "Workday OAuth".
+# Matches via the `startswith(displayName, '...')` Graph filter.
+_WORKDAY_SP_PREFIXES = ("Workday",)
```
qazizaahirah: just for my understanding, what if the entra app is named something else? aka no workday prefix

solutions/ess-maker-skills/scripts/flightcheck/checks/authentication.py
```diff
@@ -131,6 +134,302 @@ def run_authentication_checks(runner) -> list[CheckResult]:
...
+           * No assignments → FAILED (OBO will fail for all users).
+           * At least one Group assignment → PASSED.
+           * Only User-typed assignments → WARNING (works, but a security
+             group is the supportable pattern).
+
+    If no Workday SP is found, return SKIPPED — the customer tenant doesn't
+    have the Workday SSO app provisioned yet, so this gate isn't
+    applicable until they install it.
+    """
+    cp_id = "AUTH-005"
```
qazizaahirah, line 225: From copilot agency: AUTH-005 emits up to 3 CheckResult objects (one Failed, one Warning, one Passed) all with checkpoint_id="AUTH-005". If the report deduplicates or keys by checkpoint_id, only   one would survive. Verify the report renderer handles multiple results with the same ID.

solutions/ess-maker-skills/scripts/flightcheck/graph_client.py
GrahamMcMynn: Fixed in dcb1ce3.

Added an opt-in `raise_on_permission_error: bool = False` kwarg to `get_all()`. When set, the helper raises `PermissionError` on HTTP 401/403 instead of swallowing them as the existing partial-results path; the default behavior is unchanged for the 7+ other callers that rely on partial-results semantics. `get_app_role_assignments()` is now a single `get_all(path, raise_on_permission_error=True)` wrapped in `try/except PermissionError` that re-raises with a friendlier message. No more page-1 double-fetch.

solutions/ess-maker-skills/scripts/flightcheck/checks/authentication.py
GrahamMcMynn: Fixed in dcb1ce3.

You're right - displayName matching is fragile. The check now discovers the Workday Enterprise App by its immutable `servicePrincipal.applicationTemplateId`, which Entra sets at provisioning time when the SP is created from a gallery template and never changes thereafter (even if the customer renames the SP to `ESS SSO Provider`).

Two-step lookup:
1. `GET /applicationTemplates?$filter=startswith(displayName,'Workday')` returns the catalog of Workday gallery templates (tenant-independent metadata, no extra consent required).
2. Filter the result to templates whose `supportedSingleSignOnModes` array contains `saml` or `oidc` - this is the documented CSDL discriminator for federated-SSO templates per [the Graph `applicationTemplate` schema](https://learn.microsoft.com/graph/api/resources/applicationtemplate). (Initial implementation filtered by `categories` contains `Single sign-on`, but `Single sign-on` is not a valid `categories` value - the Workday SSO template is categorized as `Human resources`. Caught by re-testing against a real tenant.)
3. `GET /servicePrincipals?$filter=applicationTemplateId eq '<id>' or ...` for the resolved template id(s).

The displayName prefix fallback is removed entirely. Tenant rename now has no effect on detection. Test coverage added in `test_renamed_workday_sp_discovered_by_application_template_id`.

solutions/ess-maker-skills/scripts/flightcheck/checks/authentication.py
GrahamMcMynn: Verified in dcb1ce3.

The renderer is safe today - I confirmed by reading both code paths:

- `runner._generate_html_report` (runner.py:213-246) iterates the flat results list and writes one `<tr>` per result. No keying by `checkpoint_id`.
- `runner.run` (runner.py:120-130) aggregates category counts by iterating, not by grouping on `checkpoint_id`.

So multiple `CheckResult` rows sharing the same `checkpoint_id` render and serialize correctly today. To guard against future regressions:

1. Added an inline comment block in `authentication.py` near the AUTH-005 bucket-emission code documenting this contract (and noting it now applies to any check that emits buckets, e.g. AUTH-006).
2. Added `test_html_report_preserves_multiple_results_with_same_checkpoint_id` in `test_runner_manual_status.py` that builds a `RunResult` with 3 `CheckResult` objects all keyed `AUTH-005` (Failed + Warning + Passed) and asserts all 3 rows appear in both the HTML and the JSON output. If a future change accidentally introduces `checkpoint_id` keying in the renderer, that test will fail loudly.

qazizaahirah (approved):

PR #91, opened 2026-05-15
flightcheck: add SN-002  ServiceNow Entra User Sign In federated auth probe

### LINKED ISSUE #42: FlightCheck: Add check for ServiceNow connector Entra User Sign In federated auth

## What should be checked

When the ServiceNow Power Platform connector is configured to authenticate with the "Microsoft Entra User Sign In" authentication mode, FlightCheck should verify that the end-to-end federated identity flow actually succeeds. Customers using a third-party identity provider federated with Entra (e.g., Ping Federate) can have a connector that *appears* properly configured (basic auth works for testing) but the Entra User Sign In flow silently fails because token exchange between Entra and the external IdP does not complete and the request never reaches ServiceNow. The customer typically only discovers this after deployment when the connector returns "User Not Authenticated".

The check should:

1. Detect when a ServiceNow connection in the customer's environment uses the "Microsoft Entra User Sign In" auth mode.
2. Make a lightweight authenticated probe call against the configured ServiceNow instance using that connection.
3. Surface a clear error and remediation guidance if the call returns 401/403 or otherwise indicates the request did not reach ServiceNow.

## Suggested implementation

- **Checkpoint ID:** SN-002
- **Category:** External Systems
- **Priority:** HIGH
- **Validation logic:** Enumerate Power Platform connections of type ServiceNow in the target environment. For each one whose auth mode is "Microsoft Entra User Sign In", invoke a low-impact ServiceNow REST endpoint (e.g., GET /api/now/table/sys_user?sysparm_limit=1) and assert HTTP 200. If federated SSO (PingFederate, ADFS, Okta, etc.) is in use, include guidance that token exchange between Entra and the external IdP must be configured for the connector audience.

## Context

GrahamMcMynn:
Implements the FlightCheck check requested in #42.

**Checkpoint:** `SN-002`
**Category:** External Systems
**Priority:** HIGH
**Conditional?** No (skips silently when no `entraIDUserLogin` ServiceNow connections exist)

**API tier:** `validated`
**Mock source:**
  - `tests/fixtures/cassettes/flightcheck_pp_admin.yaml` line 2640 (Connected `entraIDUserLogin` ServiceNow connection)
  - `tests/fixtures/cassettes/flightcheck_pp_admin.yaml` line 2658 (Errored `entraIDUserLogin` ServiceNow connection with AADSTS50173 grant-expired)

## What this check does

Surfaces silent federated-SSO / token-exchange failures for ServiceNow Power Platform connections that use the `Microsoft Entra User Sign In` auth mode, before customers hit the runtime `User Not Authenticated` error described in the issue.

The check enumerates connections via the BAP admin API, filters for ServiceNow connections where `connectionParametersSet.name == `entraIDUserLogin`, and reports the stored token-status that BAP itself populates from periodic token refreshes  surfacing AADSTS50173 (grant expired), `invalid_grant`, and `is not authenticated` failure modes captured in the cassette. When the failure matches the AAD token-refresh fingerprint, the remediation explicitly calls out the federated-IdP token-exchange (PingFederate, ADFS, Okta) audience-config gap that is the root cause for federated-SSO customers.

| State | SN-002 status |
|---|---|
| All `entraIDUserLogin` SN connections `Connected` | `Passed` |
| All in `Error` state | `Failed` (with federated-SSO remediation if the error is token-related) |
| Mixed (1 Connected and 1 Error) | `Passed` (with callout listing the broken ones) |
| ServiceNow connections exist but none use `entraIDUserLogin` | silent (different auth mode  out of scope for SN-002) |
| No ServiceNow connections at all | silent (SN-001 already covers this) |

## Note on the issue's literal suggested implementation

The issue suggested `invoking a low-impact ServiceNow REST endpoint via the connection` (`GET /api/now/table/sys_user`) to assert the federated identity flow succeeds end-to-end. That probe is not directly executable from FlightCheck:

- `Microsoft Entra User Sign In` connections are user-delegated.
- Only the connection owner's bearer token can call ServiceNow on their behalf.
- FlightCheck (running with PP Admin scope) has no access to those tokens.

The implemented check achieves the same intent by reading the already-surfaced token-refresh status that BAP itself performs on the customer's behalf and persists on the connection record's `statuses[*]` field  a signal that fails for exactly the same federated-IdP token-exchange reason a runtime probe would.

## Tests

`tests/flightcheck/checks/test_servicenow_entra_user_signin.py`  12 tests, all passing locally:

- **GOOD:** single Connected, multiple Connected
- **BAD:** single Errored (with federated-SSO remediation), all Errored, non-token Error path (Failed without federated-specific remediation)
- **MIXED:** one Connected + one Errored
- **SKIP:** no connections, only Basic-auth ServiceNow, only non-ServiceNow, missing env_id, missing PP admin client
- **API filter:** `entraIDUserLogin` on a non-ServiceNow connector is correctly excluded

All 115 pre-existing tests continue to pass.

A code review pass will run automatically on this PR shortly.

Fixes #42


solutions/ess-maker-skills/scripts/flightcheck/checks/external_systems.py
```diff
@@ -149,3 +162,208 @@ def run_external_systems_checks(runner) -> list[CheckResult]:
...
+        or "invalid_grant", or "is not authenticated".
+    Any of those tokens count  we don't want to over-fit to a single
+    AADSTS number because the federated-IdP rejection paths surface
+    several different sub-codes.
+    """
+    err = status_entry.get("error") or {}
+    msg = (err.get("message") or "").lower()
+    code = (err.get("code") or "").lower()
+    target = (status_entry.get("target") or "").lower()
+    if target == "token" or "unauthorized" in code:
```
GrahamMcMynn: `_is_entra_token_failure()` currently returns true for any status whose `target` is `token` (or whose code merely contains `Unauthorized`), before checking for the AAD/refresh/invalid_grant/not-authenticated fingerprint below. That means a generic token/Unauthorized failure on an Entra User Sign In ServiceNow connection would get the federated-SSO IdP remediation even when the stored error does not identify the AAD token-refresh path; the non-token test only avoids this by changing `target` to `service`. Please require one of the captured message fingerprints (for example `aadsts`, `refresh access token`, `invalid_grant`, or `not authenticated`) before using the federated-specific remediation, and treat `target`/`code` as supporting context rather than sufficient evidence.

GrahamMcMynn (commented):

solutions/ess-maker-skills/scripts/flightcheck/checks/external_systems.py
GrahamMcMynn: Addressed in d85eeeb. Removed the `target == "token"` / `code == "Unauthorized"` early-return shortcut and now require an actual AAD/refresh fingerprint (`aadsts`, `refresh access token`, `invalid_grant`, or `not authenticated`) in the message before attributing the failure to the Entra federation. SN-002 still reports the connection as Failed in either case  only the remediation copy differs.

Added regression test `test_token_target_without_aad_fingerprint_skips_federated_guidance` that constructs an Errored entraIDUserLogin connection with `target="token"` + `code="Unauthorized"` + a non-AAD message and asserts the federated-specific remediation is suppressed while the failure is still surfaced. All 116 tests pass.

nehaoss (commented):
Well-structured new check (SN-002). Comments:

1. **Approach validation**: Rather than attempting a live ServiceNow call (which would require user-delegated tokens FlightCheck doesn't have), the check reads the BAP connection status — smart use of already-surfaced signal. The rationale docstring in \_check_sn_entra_user_signin_connections\ is excellent documentation for future contributors.

2. **Token failure detection (\_is_entra_token_failure\)**: Matching on substring tokens (\adsts\, \
efresh access token\, \invalid_grant\, \
ot authenticated\) in error messages is pragmatic. Be aware this is fragile if Microsoft changes error message wording. Consider also matching on \rror.code\ field (e.g., \Unauthorized\) as a secondary signal.

3. **Constants naming**: \SERVICENOW_API_FRAGMENT = 'service-now'\ and \ENTRA_USER_LOGIN_PARAMETERS_SET = 'entraIDUserLogin'\ — well-documented with comments explaining where these values come from in the BAP API response.

4. **Integration with existing check flow**: Clean extension point — \
esults.extend(_check_sn_entra_user_signin_connections(runner))\ after SN-001. The module docstring is updated to reflect the new check ID.

5. **Test coverage**: The 834 additions include test files — please ensure the test exercises both the healthy-connection path and the Entra-failure-detection path with realistic fixture data.

qazizaahirah:
Already fixed in this PR

PR #101, opened 2026-05-26
feat(flightcheck): add ENV-004 connections & connection references va…

qazizaahirah:
…lidation

Reports every Power Platform connection and connection reference in the target environment with binding state. Detects orphans in both directions:
- Orphan references pointing to missing connections (FAIL)
- Unbound references with no connection set (FAIL)
- Unbound connections with no reference (WARN)

Includes documentation update to validation-matrix.md.

## Related issue

Fixes #

## Type of change
- [ ] Documentation update
- [ ] Refactor / cleanup

## Checklist
- [ ] My code follows the existing style
- [ ] I have updated documentation as needed


solutions/ess-maker-skills/scripts/flightcheck/checks/environment.py
```diff
@@ -126,3 +130,227 @@ def run_environment_checks(runner) -> list[CheckResult]:
...
+    elif has_unbound_conns:
+        overall_status = Status.WARNING.value
+    else:
+        overall_status = Status.PASSED.value
+
+    # --- Summary ---
+    summary_parts = [
+        f"{len(all_conns)} connection(s)",
+        f"{len(conn_refs)} reference(s)",
+        f"{len(bound_conn_ids)} bound",
```
GrahamMcMynn: The "bound" count in the summary uses `len(bound_conn_ids)`, which counts the number of **distinct connections that are referenced**, not the number of references that are bound. When multiple references point to the same connection, the math the reader sees no longer adds up.

Example: 10 refs, 7 bound to 5 distinct connections, 1 orphan, 2 unbound → the summary reads `"10 reference(s) | 5 bound | 1 orphan ref(s) | 2 unbound ref(s)"`. 5 + 1 + 2 = 8, so the reader sees 2 "missing" refs that aren't actually missing.

Suggested fix — track bound refs as a counter (or derive it):

```python
bound_refs = len(conn_refs) - len(orphan_refs) - len(unbound_refs)
summary_parts = [
    f"{len(all_conns)} connection(s)",
    f"{len(conn_refs)} reference(s)",
    f"{bound_refs} bound ({len(bound_conn_ids)} distinct conn(s))",
]
```

GrahamMcMynn (commented):

solutions/ess-maker-skills/scripts/flightcheck/checks/environment.py
```diff
@@ -126,3 +130,227 @@ def run_environment_checks(runner) -> list[CheckResult]:
...
+            priority=Priority.HIGH.value, status=Status.WARNING.value,
+            description="Connections & connection references",
+            result=f"Error fetching connections: {e}",
+        ))
+        return results
+
+    # --- Fetch connection references from Dataverse ---
+    try:
+        sys.path.insert(0, os.path.join(os.path.dirname(__file__), "..", ".."))
+        from auth import query_all
```
GrahamMcMynn: Two things tangled inside this try-block hurt debuggability and add dead code:

1. `sys.path.insert(0, .../scripts/)` is already done at startup by `cli.py` (line 29), which is the only entry point that imports this module — the local insertion is dead defensive code.
2. `from auth import query_all` lives in the same try-block as the network call. If `auth` ever fails to import (rename/move/syntax error), the ImportError is reported as `"Error querying connection references: No module named 'auth'"`, sending an operator hunting for a Dataverse/permissions issue rather than a code/packaging issue.

Suggested fix — hoist the import to module top (matches the rest of the codebase pattern) and drop the sys.path mutation:

```python
# top of file, alongside the other imports
from auth import query_all   # scripts/auth.py, on path via cli.py
```

Then the try-block only wraps the actual `query_all(...)` call, and an `ImportError` surfaces as itself instead of being mis-reported as a query failure.

GrahamMcMynn (approved):

PR #105, opened 2026-05-28
ci(samples): Enforce static validation of agent-generated topic changes

srideshpande:
## Summary

Adds executable static validation for `samples/` topic changes and wires it into the PR flow.

This closes the gap where validation was previously described in agent guidance but not mechanically enforced.

## What changed

- Added `tools/validate_samples/` validator CLI.
- Added static checks for:
  - YAML parsing
  - `kind: AdaptiveDialog`
  - XML parsing
  - new XML filename convention
  - new topic folder convention + README
  - diff scope
  - secrets / internal URL detection
- Added `whitelist.yml` for documented pre-existing inconsistencies.
- Added pytest coverage for the validator. All 16 tests pass.
- Added `validate-samples.yml` GitHub Action.
- Updated `PULL_REQUEST_TEMPLATE.md` with a samples static validation section.
- Added `docs/validation.md` covering scope, local usage, failure interpretation, and runtime-validation boundaries.
- Updated `validate-sample-topic` skill and `AGENTS.md` to use the CLI as the validation source of truth.

## End-to-end flow

1. User opens a GitHub issue using the **Sample topic request or fix** template.
2. Issue is labeled `area:samples` and `needs-triage`.
3. Agent triages the issue and classifies it as new topic, fix, needs clarification, blocked, duplicate, out of scope, or not actionable.
4. If actionable, agent creates or updates the target topic inside `samples/` only.
5. Agent runs:

   ```bash
   python -m tools.validate_samples --diff-base origin/main
6. If validation fails, the agent fixes the issue or posts a blocker comment.
7. If validation passes, the agent opens a PR with:
    - summary tied to the issue
    - changed files
    - neighbor topic reference
    - validation summary block
    - Fixes # < issue-number >
8. CI runs the same validator on the PR diff.
9.  CI writes the validation summary to the job summary and sticky PR comment.
10. CI fails on any validation failure.
11.  Human reviewer reviews the PR, CI result, and diff.
12. On approval and green checks, the PR is merged and the issue closes automatically.

## Out of scope

This validation is static only. It does not validate:

  - Copilot Studio import
  - Power Platform runtime behavior
  - connector execution
  - tenant-specific configuration
  - end-to-end topic conversation behavior

Runtime validation remains a manual follow-up step.

## Validation
- All 16 validator tests pass (`test_validate_samples.py`).
- CLI was dogfooded against the existing samples/ tree.
- Workflow is ready to be marked as a required check on main after it runs green on a few real PRs.


tools/validate_samples/cli.py
```diff
@@ -0,0 +1,122 @@
...
+
+    if args.diff_base:
+        changed = _git_changed(repo_root, args.diff_base)
+    else:
+        changed = _parse_paths_file(Path(args.paths_file))
+
+    results = run_all_checks(repo_root, changed, whitelist)
+
+    if args.json:
+        print(results_to_json(results))
```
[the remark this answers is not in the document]
srideshpande: resolved.

tools/validate_samples/cli.py
```diff
@@ -0,0 +1,122 @@
...
+    else:
+        print(render_summary(results))
```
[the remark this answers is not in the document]
srideshpande: Resolved.

.github/workflows/validate-samples.yml
```diff
@@ -0,0 +1,83 @@
...
+# Licensed under the MIT License.
+
+name: Validate samples
+
+on:
+  pull_request:
+    branches: [ "main" ]
+    paths:
+      - "samples/**"
+      - "tools/validate_samples/**"
```
GrahamMcMynn, line 11: **[bug]** The workflow triggers on `samples/**`, `tools/validate_samples/**`, and `.github/workflows/validate-samples.yml`, but `check_diff_scope` rejects any path outside `samples/`. As a result, any future maintenance PR that touches only validator code or this workflow file (without also touching `samples/`) will trigger this workflow and then fail it — there are no samples to validate, but every changed validator/workflow path is reported as out-of-scope and the job exits non-zero. This PR's own validation run reproduced the failure (`Diff scope (samples/ only): FAIL — .github/PULL_REQUEST_TEMPLATE.md: outside samples/, …`).

Suggested fix: in `check_diff_scope`, treat the check as `N-A` when the change set contains no paths under `samples/` (the validator simply has nothing to evaluate), rather than failing each non-samples path. Alternatively, narrow this workflow's path filter to `samples/**` only and run a separate test job for validator-code changes.

tests/test_validate_samples.py
```diff
@@ -0,0 +1,265 @@
+# Copyright (c) Microsoft Corporation.
```
GrahamMcMynn, line 1: **[testing]** The new test suite (16 tests) is not wired into any CI workflow. The existing `ci.yml` runs only `ruff` and `python -m compileall` against a hardcoded `candidates` list, and the new `validate-samples.yml` runs only the validator binary — not `pytest`. Since the validator is becoming a gate on `samples/` PRs, undetected regressions in the validator itself (e.g., a check accidentally returning `PASS` instead of `FAIL`) would silently let bad samples merge.

`pyproject.toml` already declares `PyYAML` and `pytest` under `[project.optional-dependencies] test` and sets `testpaths = ["tests"]`, so wiring this in is small. Suggest adding either:
- a `pytest` step to `validate-samples.yml` (after the validator step), or
- a new dedicated test job that runs `pip install -e .[test] && pytest tests/test_validate_samples.py` on changes under `tools/validate_samples/**` or `tests/**`.

GrahamMcMynn (commented):
Code review by ESS Code Review Agent (2 comments). Existing comments from Copilot/CodeQL bots were respected and not duplicated.

.github/workflows/validate-samples.yml
```diff
@@ -0,0 +1,116 @@
...
+      - name: Install dependencies
+        run: pip install PyYAML
+
+      - name: Compute samples-only changed files
+        id: changed
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+        run: |
+          git diff --name-status "$BASE_SHA...HEAD" \
+            | awk -F'\t' 'BEGIN{OFS="\t"} { if ($1 ~ /^[RC]/) { p=$3 } else { p=$2 } if (p ~ /^samples\//) { if ($1 ~ /^[RC]/) print "A", p; else print $1, p } }' \
```
GrahamMcMynn: **[bug]** The `Compute samples-only changed files` step uses `awk` to filter `git diff --name-status` output down to **samples/-prefixed paths only** before passing it to the validator via `--paths-file samples-changed.txt`. This silently defeats `check_diff_scope` in CI: a PR that edits both `samples/Foo/topic.yaml` *and* (e.g.) `solutions/bar.txt` will have the `solutions/` entry stripped by awk, so the validator never sees it and `Diff scope (samples/ only)` reports **PASS** even though the diff is, in fact, out of scope.

This is asymmetric with the local invocation (`--diff-base origin/main`), where `_git_changed` returns *all* paths and the check works correctly — so the local-vs-CI parity claim ("Runs locally, in PRs ... All three call the same CLI so results cannot drift", `docs/validation.md`) is violated for this specific check.

**Fix:** drop the `if (p ~ /^samples\//)` predicate in the awk filter (let the validator see every changed path and decide), or use `--diff-base` in CI and rely on `check_diff_scope` to enforce. The `has_changes` gate can still skip the validate step when there are *no* samples paths at all, but only after considering the full diff.

tools/validate_samples/checks.py
```diff
@@ -0,0 +1,322 @@
...
+    xmls = [c for c in changed if c.path.endswith(".xml") and not c.is_deleted and c.path.startswith(SAMPLES_ROOT + "/")]
+    if not xmls:
+        return res
+    res.status = Status.PASS
+    for c in xmls:
+        text = _read_text(repo_root, c.path)
+        if text is None:
+            continue
+        try:
+            ET.fromstring(text)
```
GrahamMcMynn: **[security]** `check_xml_parse` uses `xml.etree.ElementTree.fromstring` to parse XML from PR content. The stdlib `xml.etree` parser is vulnerable to XML attacks — most notably the "billion laughs" / quadratic-blowup entity-expansion attack, which can cause unbounded memory / CPU consumption when parsing a small (~few KB) malicious file. Since this validator runs in CI on contributor-provided XML (the whole point is to validate untrusted PR content), a crafted `msdyn_*.xml` could trivially DoS the Actions runner.

Python explicitly recommends `defusedxml` for any XML "parsed from an untrusted source" — see [Python docs: XML vulnerabilities](https://docs.python.org/3/library/xml.html#xml-vulnerabilities).

**Fix:** add `defusedxml` to the validator''s deps and replace the import:

```python
from defusedxml.ElementTree import fromstring as _xml_fromstring
# ...
_xml_fromstring(text)
```

Then add `defusedxml` to the `pip install` line in `.github/workflows/validate-samples.yml` (both jobs).

GrahamMcMynn (commented):
Code review by ESS Code Review Agent (2 comments). Existing feedback from Copilot/CodeQL bots and prior reviews was respected and not duplicated.

nehaoss (commented):
Comprehensive and well-engineered. The single-CLI approach (local, CI, agent skill all call the same code) eliminates drift. Comments:

1. **PR template duplication**: The 'Pre-PR Checklist' block in \.github/PULL_REQUEST_TEMPLATE.md\ and \CONTRIBUTING.md\ are nearly identical. If one is updated, the other must be too. Consider referencing CONTRIBUTING.md from the template instead of duplicating.

2. **CI workflow — \self_edit\ detection**: The regex in \changed.txt\ grep matches tool/test/workflow paths to skip the diff-scope gate for maintainer changes. Smart approach. One edge case: a PR that touches BOTH \	ools/validate_samples/\ AND \samples/\ would skip validation on the sample changes. Is that intentional?

3. **Secrets check scans full file contents**: The design choice to scan the entire file (not just added lines) means pre-existing secrets in touched files will fail. This is a deliberate 'force cleanup' choice — worth calling out in the PR description for reviewers who might wonder about false positives on legacy files.

4. **Test coverage**: Good that \	est_validate_samples.py\ uses synthetic repos in \	mp_path\. The helper functions (\_write\, \_new_topic\) make tests readable.

5. **\ctions/checkout@v6\**: Double-check this version exists — as of my knowledge cutoff, v4 was latest. If v6 doesn't exist yet, the workflow will fail.

GrahamMcMynn (approved):

PR #113, opened 2026-06-01
FlightCheck: AUTH-006 — SAML NameID alignment with Workday (fixes #84)

### LINKED ISSUE #84: FlightCheck: Add check for SAML NameID alignment with Workday user identifier

## What should be checked

## Suggested implementation

- **Checkpoint ID:** AUTH-006
- **Priority:** HIGH

## Context

GrahamMcMynn:
Fixes #84.

## Approach

Issue #84 was originally blocked because the Workday admin SOAP `Security` service operations needed to read the Workday-side SAML NameID expectation are not exposed via WS-Security UsernameToken on a typical tenant, and FlightCheck checks can't prompt the operator for live SAML traces. This PR unblocks the check by introducing a new `MANUAL` result status: the kit reads the Entra-side half of the comparison via Microsoft Graph (which IS programmatically accessible) and tells the operator exactly which Workday screen + field to compare against. The Workday-side instructions were validated interactively against a live Workday tenant during PR review.

This MANUAL pattern is the template for any future Workday tenant-config check that hits the same admin-API gap.

## What changed

**Runner / report (general infrastructure, benefits any future MANUAL check):**
- New `Status.MANUAL` enum value with string `"Manual"`
- `CategorySummary.manual` + `RunResult.manual` tally fields
- MANUAL does NOT fail readiness (semantics: informational, not a gate)
- HTML report: new `Manual` summary card, `status-manual` CSS class, and a `cell-text` class on result/remediation cells so multi-line strings preserve authored line breaks and indentation instead of collapsing into a wall of text
- JSON output includes top-level `manual` count
- `solutions/ess-maker-skills/scripts/flightcheck/AGENTS.md` design principle #2 documents when to use MANUAL

**AUTH-006 check (the actual fix for #84):**
- Filters Entra service principals server-side via Graph v1.0 with `startswith(displayName,'Workday') and preferredSingleSignOnMode eq 'saml'`
- Reads each app's `claimsMappingPolicies` to detect any NameID override (absent custom policy = default `NameID = user.userPrincipalName`)
- **Coalesces N detected Workday apps into ONE MANUAL result** so the operator sees the manual-verification steps just once. Each detected app is listed with its `servicePrincipalNames` SAML entity ID — the join key against Workday's "SAML Identity Providers > Service Provider ID" column — so the operator can identify which Entra app the Workday tenant actually uses without leaving Workday.
- Remediation has two phases:
  - **Step 1** identifies the active app from inside Workday: open Edit Tenant Setup - Security → find the enabled IdP row → note its Service Provider ID → match to one of the entity IDs above.
  - **Step 2** walks the operator through the NameID comparison: pick a pilot user → look up their Entra value for the NameID attribute → open the `All Workday Accounts` report → compare the `User Name` column.
- `NOT_CONFIGURED` when no Workday SAML app exists; `SKIPPED` when the Graph client is unavailable.
- `doc_link` points at the authoritative MS Learn Workday SSO tutorial (the same page that explains the Entra-side fix path).

**Mocks / tests:**
- `tests/mocks/graph.py` gains validatable `service_principal()` and `claims_mapping_policy()` builders + `list_*` wrappers, citing the Graph v1.0 CSDL (`\`) and the relevant Learn operation docs per the validatable-tier convention.
- `graph_client.get_claims_mapping_policies()` helper.
- 12 new tests cover: NOT_CONFIGURED (no Workday app); MANUAL with default mapping; MANUAL with custom NameID override; MANUAL with non-NameID policy attached; multi-app coalesce (5 apps → 1 row, entity IDs surfaced, GUIDs filtered out); 403 on SPs; 403 on claimsMappingPolicies; `runner.graph` is None. Plus 4 runner tests pinning the MANUAL enum value, tally bucket, READY semantics, and HTML/JSON serialization.

## Validation

- `python -m pytest`: **146 passed, 8 skipped, 0 regressions**
- `ruff check` on all six touched files: **clean**
- AUTH-006 remediation steps eyeballed against a live Workday tenant by the PR author (this is what motivated the two design iterations to coalesce results and pivot from a fabricated "Identifier field" instruction to the validated "All Workday Accounts → User Name column" path).

## Notes for reviewer

- The new Microsoft Graph endpoints (`servicePrincipal`, `claimsMappingPolicy`) are both in Graph v1.0, which is already in the API tier registry as `validatable`. No registry change needed.
- Customer tenants need `Application.Read.All` and `Policy.Read.All` consented on the kit's Graph app registration for AUTH-006 to function fully. Missing consent surfaces as `NOT_CONFIGURED` (empty SP list) or as a per-app fallback to the default-mapping summary — both safe.
- No fabricated URLs: every link in the remediation text is verified live.
- No leaked tenant IDs in the shipped code (the early preview's `microsoft_dpt6` example string was replaced with `contoso_prod` per kit-wide redaction convention before commit).

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>


GrahamMcMynn (commented):
Code review by ESS Code Review Agent (audit mode, 2 comments posted).

solutions/ess-maker-skills/scripts/flightcheck/checks/authentication.py
```diff
@@ -111,4 +113,288 @@ def run_authentication_checks(runner) -> list[CheckResult]:
...
+            description=description,
+            result=f"Unable to query Entra service principals: {e}",
+            remediation=(
+                "Requires Application.Read.All (or Directory.Read.All) "
+                "consented on the Graph app registration."
+            ),
+            doc_link=doc_link,
+        )]
+
+    if not workday_sps:
```
GrahamMcMynn, line 218: `AUTH-006` reports a confident `NOT_CONFIGURED` ("No federated Workday enterprise app found in Entra...") when `workday_sps` is empty, but `graph_client.get_all()` returns `[]` on a 401/403 (see `graph_client.py:154-155`). A customer who deploys without `Application.Read.All` consented will get a check result that confidently asserts "you don't have a federated Workday app, so this check is N/A" — when the truth is "we couldn't list service principals at all."

The downstream consequence is exactly the silent-failure mode this check is meant to surface (end-user SAML SSO into Workday broken, ISU runtime calls still work, agent looks healthy). An operator reading the AUTH-006 NOT_CONFIGURED message has no signal to investigate further.

The same swallowed-403-becomes-empty-list pattern also drives the misleading test in `test_authentication_saml.py::TestErrorHandling::test_servicepriniclap_403_emits_warning` (separate comment below). Tier-1 fix: probe `/servicePrincipals?\$top=1` with `graph.get()` first and surface `_status == 403` as `WARNING` with an `Application.Read.All` remediation, before falling through to the `NOT_CONFIGURED` branch.

tests/flightcheck/checks/test_authentication_saml.py
```diff
@@ -0,0 +1,335 @@
...
+        # We check via the substring "entity IDs: guid-noise" — if it
+        # appears, the filter let a GUID through.
+        assert "entity IDs: guid-noise" not in r.result
+
+
+class TestErrorHandling:
+    """Graph failures surface as WARNING, not silent pass."""
+
+    @responses.activate
+    def test_servicepriniclap_403_emits_warning(
```
GrahamMcMynn: The `TestErrorHandling` class docstring says "Graph failures surface as **WARNING**, not silent pass" and its two tests are named `test_servicepriniclap_403_emits_warning` / `test_claims_mapping_403_emits_warning_for_that_sp` — but both tests actually assert that 403 becomes `NotConfigured` / `Manual`. The class+test names assert the *intended* behavior; the assertions pin the *current quirky* behavior. When either test fails in the future, the name will tell the next maintainer the opposite of what the test actually covers.

Two cleanups:
1. The class docstring overstates the contract (failures **don't** surface as WARNING today — that's the bug from the `authentication.py` comment). Either fix the production code and keep the name, or rename the class/tests to something like `test_servicepriniclap_403_currently_masquerades_as_notconfigured` and link to a follow-up issue.
2. Also: `test_servicepriniclap_403_emits_warning` has a typo (`servicepriniclap` → `service_principal`).

amilandi:
**Nit (non-blocking):** `_saml_entity_ids()` would admit a bare `urn:uuid:...` containing a GUID substring. Unlikely in practice (Workday gallery apps always use `http://www.workday.com/{tenant}`), and the filter is noise-reduction not security — just noting it.

amilandi (approved):

GrahamMcMynn:
> **Nit (non-blocking):** `_saml_entity_ids()` would admit a bare `urn:uuid:...` containing a GUID substring. Unlikely in practice (Workday gallery apps always use `http://www.workday.com/{tenant}`), and the filter is noise-reduction not security — just noting it.

I don't think that is worth fixing... as honestly if the urn:uuid: format was used it's technically valid.

PR #115, opened 2026-06-02
Add FlightCheck-only installer and pip dependency fix

amilandi:
## Summary

Adds a `-FlightCheckOnly` flag to the one-shot installer (`setup/Install-EssAdk.ps1`) that installs only the minimal toolchain needed to run FlightCheck without requiring VS Code, GitHub Copilot, or the full `/setup` flow.

Also fixes a gap where pip dependencies (msal, requests, PyYAML, defusedxml) were never installed by the installer, causing `/setup` to fail on fresh machines.

ADO task: https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7439972 

##  Demo video
https://github.com/user-attachments/assets/44e7fae8-6427-452c-a5da-ec533140e9a8

### Full ADK Maker Kit install (developers customizing ESS agents)

```powershell
iex (irm https://raw.githubusercontent.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/main/setup/bootstrap.ps1)
```

Installs Python, Git, GitHub CLI, PowerShell 7, VS Code + extensions, clones the repo, installs pip dependencies, and opens VS Code in the maker workspace.

### FlightCheck only (validating a deployed environment)

```powershell
iex (irm https://raw.githubusercontent.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/main/setup/bootstrap-flightcheck.ps1)
```

Single command that handles everything: installs Python + Git, signs you in, walks you through picking your environment and agent, then runs FlightCheck automatically. Re-run the same command to change your environment or agent.

### Installer (`setup/Install-EssAdk.ps1`)
- **Pip dependency install** — New step installs from requirements.txt after Python is available (deferred to post-clone when running via bootstrap)
- **`-FlightCheckOnly` flag** — Minimal toolchain (Python + Git only), interactive environment/agent discovery via discover.py, generates `.local/config.json` for FlightCheck
- **Auto-run FlightCheck** — After config generation, FlightCheck runs immediately so users see results without a second command
- **Reconfigure support** — Re-running with `-FlightCheckOnly` detects existing config and offers to reconfigure (change environment or agent)

### Robustness improvements
- **`Invoke-Native` helper** — Prevents PowerShell 5.1 from terminating the script when native commands (winget, pip, git, python) write to stderr under `$ErrorActionPreference = 'Stop'`
- **`Resolve-Python` helper** — Robust Python detection: tries `py -3.12`, `py -3`, validated `python` on PATH (excluding the Windows Store alias), and known install paths (`%LOCALAPPDATA%\Programs\Python\Python312\`)
- **Git hang prevention** — Sets `GIT_TERMINAL_PROMPT=0` and `GCM_INTERACTIVE=never` to prevent git clone from blocking on credential/host-key prompts
- **winget optional for FlightCheckOnly** — If Python and Git are already installed, continues without winget (supports Windows Server, LTSC, and Store-disabled environments)
- **Pre-installed tool detection** — Skips winget install for already-present packages (avoids unnecessary elevation prompts)
- **Network resilience** — Bootstrap downloads use `-TimeoutSec 60` with actionable guidance when blocked by firewalls/proxies
- **Manual install fallback** — When winget is unavailable or fails, provides direct download URLs for Python, Git, etc.
- **Output filtering** — Suppresses winget progress bar Unicode garbage and spinner frames that corrupt terminal output

### New files
- **`setup/bootstrap-flightcheck.ps1`** — One-liner entry point that downloads the installer and passes `-FlightCheckOnly`

### Unicode/encoding fixes
- **`scripts/auth.py`** — Replace Unicode arrows with ASCII (fixes crash on CP1252 consoles)
- **`scripts/discover.py`** — Replace box-drawing chars with ASCII hyphens
- **Config JSON** — Write without UTF-8 BOM (PS 5.1's `-Encoding utf8` adds BOM; Python's `json.load()` rejects it)

### Documentation
- **`setup/README.md`** — Rewritten to document both install modes with user-friendly language
- **`solutions/ess-maker-skills/README.md`** — Added standalone install section to FlightCheck docs

## Testing

Tested end-to-end on a clean Windows machine (no Python/Git pre-installed):
1. Bootstrap downloads installer from raw GitHub URL
2. Installs Python 3.12 + Git via winget
3. Clones repo from branch
4. Installs pip dependencies successfully
5. Interactive environment picker lists environments, user selects one
6. Agent discovery lists agents, user selects one (or skips)
7. Config written (BOM-free UTF-8), FlightCheck runs automatically and displays results

Also validated:
- Re-run detects existing config and offers reconfigure prompt
- Script works in both PowerShell 5.1 (Windows PowerShell) and PowerShell 7
- Output is clean (no Unicode garbage from winget progress bars)

### DIFF
--- a/setup/Install-EssAdk.ps1
+++ b/setup/Install-EssAdk.ps1
@@ -654,16 +654,18 @@ if ($FlightCheckOnly) {
     if ($pythonExe) {
         Push-Location $workspace
         try {
+            # Run FlightCheck with EAP=Continue so stderr doesn't terminate,
+            # but let output stream directly to console (FlightCheck is interactive)
+            $prevEAP = $ErrorActionPreference
+            $ErrorActionPreference = 'Continue'
             if ($pythonExe -eq 'py -3.12') {
-                $fcArgs = @('-3.12', 'scripts/flightcheck/cli.py', '--scope', 'full')
-                $fcOutput = Invoke-Native { & py @fcArgs }
+                & py -3.12 scripts/flightcheck/cli.py --scope full
             } elseif ($pythonExe -eq 'py -3') {
-                $fcArgs = @('-3', 'scripts/flightcheck/cli.py', '--scope', 'full')
-                $fcOutput = Invoke-Native { & py @fcArgs }
+                & py -3 scripts/flightcheck/cli.py --scope full
             } else {
-                $fcOutput = Invoke-Native { & $pythonExe scripts/flightcheck/cli.py --scope full }
+                & $pythonExe scripts/flightcheck/cli.py --scope full
             }
-            foreach ($line in $fcOutput) { Write-Host $line }
+            $ErrorActionPreference = $prevEAP
         } finally { Pop-Location }
     } else {
         Write-Warn2 'Python not found. Open a new terminal and run:'
--- a/setup/Install-EssAdk.ps1
+++ b/setup/Install-EssAdk.ps1
@@ -638,17 +638,24 @@ if ($FlightCheckOnly) {
         }
 
         # Match the structure setup.py produces: agents array + activeAgent slug
+        if ($botId) {
+            $agentsList = [System.Collections.ArrayList]@()
+            $agentsList.Add($agentEntry) | Out-Null
+        } else {
+            $agentsList = [System.Collections.ArrayList]@()
+        }
+
         $config = @{
             configVersion      = 1
             setup              = 'flightcheck-only'
             dataverseEndpoint  = $envUrl
             flightCheckOnly    = $true
             agent              = $agentEntry
-            agents             = if ($botId) { @(,$agentEntry) } else { @() }
+            agents             = $agentsList
             activeAgent        = if ($botId) { $slug } else { '' }
         }
 
-        $json = $config | ConvertTo-Json -Depth 4
+        $json = $config | ConvertTo-Json -Depth 4 -Compress:$false
         [System.IO.File]::WriteAllText($configPath, $json, (New-Object System.Text.UTF8Encoding $false))
         Write-Ok "Created $configPath"
     }
--- a/setup/Install-EssAdk.ps1
+++ b/setup/Install-EssAdk.ps1
@@ -627,20 +627,24 @@ if ($FlightCheckOnly) {
         }
 
         # Write minimal config.json sufficient for FlightCheck
+        $slug = 'flightcheck-only'
+        $agentEntry = @{
+            name       = $agentName
+            botId      = $botId
+            schemaName = $schemaName
+            isManaged  = $isManaged
+            slug       = $slug
+            folder     = ''
+        }
+
+        # Match the structure setup.py produces: agents array + activeAgent slug
         $config = @{
             setup              = 'complete'
             dataverseEndpoint  = $envUrl
             flightCheckOnly    = $true
-            agent              = @{
-                name       = $agentName
-                botId      = $botId
-                schemaName = $schemaName
-                isManaged  = $isManaged
-                slug       = 'flightcheck-only'
-                folder     = ''
-            }
-            agents             = @()
-            activeAgent        = ''
+            agent              = $agentEntry
+            agents             = if ($botId) { @(,$agentEntry) } else { @() }
+            activeAgent        = if ($botId) { $slug } else { '' }
         }
 
         $json = $config | ConvertTo-Json -Depth 4
--- a/setup/Install-EssAdk.ps1
+++ b/setup/Install-EssAdk.ps1
@@ -195,11 +195,8 @@ if (-not $wingetAvailable) {
     Write-Warn2 'winget not available. Checking for pre-installed tools...'
     $missingTools = @()
     foreach ($pkg in $packages) {
-        $cmd = Get-Command $pkg.Cmd -ErrorAction SilentlyContinue
-        if ($cmd) {
-            Write-Ok "$($pkg.Name) (already installed at $($cmd.Source))"
-        } elseif ($pkg.Cmd -eq 'python') {
-            # Use Resolve-Python for smarter detection
+        if ($pkg.Cmd -eq 'python') {
+            # Use Resolve-Python to exclude the non-functional Store alias
             $resolved = Resolve-Python
             if ($resolved) {
                 Write-Ok "$($pkg.Name) (found: $resolved)"
@@ -207,7 +204,12 @@ if (-not $wingetAvailable) {
                 $missingTools += $pkg.Name
             }
         } else {
-            $missingTools += $pkg.Name
+            $cmd = Get-Command $pkg.Cmd -ErrorAction SilentlyContinue
+            if ($cmd) {
+                Write-Ok "$($pkg.Name) (already installed at $($cmd.Source))"
+            } else {
+                $missingTools += $pkg.Name
+            }
         }
     }
     if ($missingTools.Count -gt 0) {
@@ -411,8 +413,22 @@ if (-not $SkipClone) {
         Push-Location $repoPath
         try {
             $gitOutput = Invoke-Native { & git fetch --quiet origin }
-            $gitOutput = Invoke-Native { & git checkout --quiet $Branch }
-            $gitOutput = Invoke-Native { & git pull --quiet --ff-only }
+            foreach ($line in $gitOutput) { if ($line) { Write-Host "      $line" } }
+            if ($LASTEXITCODE -ne 0) {
+                Write-Warn2 "git fetch failed (exit $LASTEXITCODE). Continuing with local copy."
+            } else {
+                $gitOutput = Invoke-Native { & git checkout --quiet $Branch }
+                foreach ($line in $gitOutput) { if ($line) { Write-Host "      $line" } }
+                if ($LASTEXITCODE -ne 0) {
+                    Write-Warn2 "git checkout $Branch failed (exit $LASTEXITCODE). Continuing on current branch."
+                } else {
+                    $gitOutput = Invoke-Native { & git pull --quiet --ff-only }
+                    foreach ($line in $gitOutput) { if ($line) { Write-Host "      $line" } }
+                    if ($LASTEXITCODE -ne 0) {
+                        Write-Warn2 "git pull failed (exit $LASTEXITCODE). Continuing with local copy."
+                    }
+                }
+            }
         } finally { Pop-Location }
     } else {
         $gitOutput = Invoke-Native { & git clone --branch $Branch --single-branch $RepoUrl $repoPath }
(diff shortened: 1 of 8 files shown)


setup/Install-EssAdk.ps1
```diff
@@ -93,27 +153,76 @@ if ($os.Platform -ne 'Win32NT' -or $os.Version.Build -lt 17763) {
...
+    )
+}
 
-if ($UseDsc) {
+if (-not $wingetAvailable) {
+    # winget not present - check if required tools are already installed
+    Write-Warn2 'winget not available. Checking for pre-installed tools...'
+    $missingTools = @()
+    foreach ($pkg in $packages) {
+        $cmd = Get-Command $pkg.Cmd -ErrorAction SilentlyContinue
```
GrahamMcMynn: In the `winget not available` branch, `Get-Command python` will match the Windows Store **python.exe alias** in `%LOCALAPPDATA%\Microsoft\WindowsApps\` and report it as "already installed", bypassing the `Resolve-Python` smart-detection in the `elseif` arm (which exists specifically to exclude the Store alias — see lines 87-101).

On a Store-disabled / LTSC machine — exactly the FlightCheck-Only target audience — the Store alias is non-functional. The script will print `[ok] Python 3.12 (already installed at …\WindowsApps\python.exe)` and then the deferred `pip install` will either pop the Store install prompt or fail silently.

Suggested fix: check `Resolve-Python` first for the Python package, e.g.

```powershell
foreach ($pkg in $packages) {
    if ($pkg.Cmd -eq 'python') {
        $resolved = Resolve-Python
        if ($resolved) { Write-Ok "$($pkg.Name) (found: $resolved)"; continue }
        $missingTools += $pkg.Name; continue
    }
    $cmd = Get-Command $pkg.Cmd -ErrorAction SilentlyContinue
    ...
}
```

setup/Install-EssAdk.ps1
```diff
@@ -240,13 +410,17 @@ if (-not $SkipClone) {
         Write-Ok "Repo already cloned at $repoPath - pulling latest"
         Push-Location $repoPath
         try {
-            & git fetch --quiet origin
-            & git checkout --quiet $Branch
-            & git pull --quiet --ff-only
+            $gitOutput = Invoke-Native { & git fetch --quiet origin }
+            $gitOutput = Invoke-Native { & git checkout --quiet $Branch }
+            $gitOutput = Invoke-Native { & git pull --quiet --ff-only }
```
GrahamMcMynn: The "already cloned, pulling latest" branch wraps each git call in `Invoke-Native`, captures into `$gitOutput`, then overwrites it on the next line and never checks `$LASTEXITCODE`:

```powershell
$gitOutput = Invoke-Native { & git fetch --quiet origin }
$gitOutput = Invoke-Native { & git checkout --quiet $Branch }
$gitOutput = Invoke-Native { & git pull --quiet --ff-only }
```

Compare with the fresh-clone branch on lines 418-425 just below, which echoes output and checks `$LASTEXITCODE`. In the refresh path:
- if `fetch` fails (network blip, auth pop), `checkout` still runs against a stale tree;
- if `checkout` fails (branch typo, dirty working tree), `pull` still runs;
- the captured stderr/stdout is discarded, so the user sees nothing.

The script then continues, runs `discover.py`, writes a config, and launches FlightCheck against potentially stale code with no indication anything went wrong. Idempotent re-run is the documented intended use case in the PR description, so this is the more common path long-term.

Suggested fix: echo `$gitOutput`, check `$LASTEXITCODE` after each call, and `throw` on failure — mirror what the fresh-clone branch already does.

setup/Install-EssAdk.ps1
```diff
@@ -256,15 +430,233 @@ if (-not $SkipClone) {
...
+    if ($pythonExe) {
+        Push-Location $workspace
+        try {
+            if ($pythonExe -eq 'py -3.12') {
+                & py -3.12 scripts/flightcheck/cli.py --scope full
+            } elseif ($pythonExe -eq 'py -3') {
+                & py -3 scripts/flightcheck/cli.py --scope full
+            } else {
+                & $pythonExe scripts/flightcheck/cli.py --scope full
+            }
```
GrahamMcMynn, line 679: The final FlightCheck launch is the only native-Python invocation in this script that does **not** go through `Invoke-Native`:

```powershell
if ($pythonExe -eq 'py -3.12') {
    & py -3.12 scripts/flightcheck/cli.py --scope full
} elseif ($pythonExe -eq 'py -3') {
    & py -3 scripts/flightcheck/cli.py --scope full
} else {
    & $pythonExe scripts/flightcheck/cli.py --scope full
}
```

`$ErrorActionPreference` is still `'Stop'` here (set at line 81 and never restored), and per `Invoke-Native`'s own docstring (line 50-51) the whole reason that helper exists is that PS 5.1 will terminate the script when a native command writes to stderr under EAP=Stop. FlightCheck writes WARNING / progress lines to stderr; any of them can abort the run mid-check, dropping the user out before they see results.

Suggested fix: wrap the same way every other native call in this script is wrapped, e.g.

```powershell
if ($pyCmd -eq 'py') {
    $fcArgs = $pyBaseArgs + @('scripts/flightcheck/cli.py','--scope','full')
    $fcOutput = Invoke-Native { & py @fcArgs }
} else {
    $fcOutput = Invoke-Native { & $pythonExe scripts/flightcheck/cli.py --scope full }
}
foreach ($line in $fcOutput) { Write-Host $line }
```

(Or just temporarily set `$ErrorActionPreference = 'Continue'` around the invocation.)

GrahamMcMynn (commented):
Code review by ESS Code Review Agent.

Four issues, all in the installer/bootstrap surface. The Unicode→ASCII and pip-deps fixes look correct.

setup/bootstrap-flightcheck.ps1
```diff
@@ -0,0 +1,66 @@
...
+)
+
+$ErrorActionPreference = 'Stop'
+[Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12
+
+$tempDir = Join-Path $env:TEMP "ess-adk-bootstrap-$([Guid]::NewGuid().ToString('N').Substring(0,8))"
+New-Item -ItemType Directory -Path $tempDir -Force | Out-Null
+
+$files = @(
+    'ess-adk-setup.winget.yaml',
```
GrahamMcMynn: `bootstrap-flightcheck.ps1` downloads two files:

```powershell
$files = @(
    'ess-adk-setup.winget.yaml',
    'Install-EssAdk.ps1'
)
```

then invokes:

```powershell
$installerArgs = @{ Branch = $Branch; FlightCheckOnly = $true }
& $installer @installerArgs
```

`ess-adk-setup.winget.yaml` is only ever read by `Install-EssAdk.ps1` inside the `elseif ($UseDsc)` branch (line 234 of `Install-EssAdk.ps1`) — and this bootstrap never passes `-UseDsc`. The YAML download is dead weight: it adds a second network round trip (with its own 60s timeout and `throw` failure surface) for a file the installer will never open. On a firewall that allows the first download but blocks the second, the customer gets `[ERR] Failed to download: …ess-adk-setup.winget.yaml` and the install aborts even though the only file actually needed was already fetched.

Suggested fix: drop `'ess-adk-setup.winget.yaml'` from `$files` in `bootstrap-flightcheck.ps1`. (`setup/bootstrap.ps1` has the same pre-existing issue, but that's out of scope for this PR — flagging just the new file.)

setup/Install-EssAdk.ps1
amilandi: Invoke-Native buffers all output until the command finishes, which hides FlightCheck's interactiveprompts (e.g. Workday test employee ID). Since FlightCheck needs to stream output and accept userinput in real-time, we can't capture its output. Instead, we temporarily set $ErrorActionPreference = 'Continue' around the call, which solves the same PS 5.1 stderr termination issue without breakinginteractivity.

The current implementation is the correct middle ground — stderr won't crash the script, but the user still sees output and can respond to prompts.

sthilkumar:
@amilandi this is cool.  Love the simplified run for FlightCheck installing the dependencies.  Couple of questions:
1. Does the login for a tenant need to be Power Platform Admin or a simple maker privileges work?
2. Like the option for running 'environment-wide' check without selecting an agent, can a tenant-wide check to be introduced right before environment selection to run tenant-wide pre-reqs. like licensing, role permissions, etc.?
3. Can this be also tested with few edge cases like where a maker/admin tries to run this in desktop that's managed by organization policies or UAC restricting executables/installers?

@GrahamMcMynn the tenant and environment pre-reqs are already covered in PowerShell version and can the converted python scripts be integrated in the flow too?  Just confirming

setup/Install-EssAdk.ps1
```diff
@@ -256,15 +446,238 @@ if (-not $SkipClone) {
...
+        } finally { Pop-Location }
+
+        # Create .local directory
+        if (-not (Test-Path $localDir)) {
+            New-Item -ItemType Directory -Path $localDir -Force | Out-Null
+        }
+
+        # Write minimal config.json sufficient for FlightCheck
+        $config = @{
+            setup              = 'complete'
```
nehaoss: **Missing `configVersion` - breaks `auth.load_config()` for downstream scripts**

This FlightCheck-only config omits `configVersion`. The canonical `auth.load_config()` in `scripts/auth.py` (line ~289-295) gates on this field:
```python
ver = cfg.get('configVersion', 0)
if ver != EXPECTED_CONFIG_VERSION:  # EXPECTED_CONFIG_VERSION = 1
    print(f'ERROR: {config_path} is schema v{ver}, expected v{EXPECTED_CONFIG_VERSION}. ...')
    sys.exit(1)
```Without `configVersion`, value defaults to 0, and any kit script calling `load_config()` will hard-exit with a confusing error. The full setup path in `setup.py` always stamps `configVersion = EXPECTED_CONFIG_VERSION`; this path should too.

**Suggestion:** Add `configVersion = 1` to this config hash. Also consider using `setup = 'flightcheck-only'` instead of `'complete'` so scripts can distinguish incomplete setup states.

File: setup/Install-EssAdk.ps1 [hunk 6]
```diff
@@ -256,15 +446,238 @@ if (-not $SkipClone) {
...
+            flightCheckOnly    = $true
+            agent              = @{
+                name       = $agentName
+                botId      = $botId
+                schemaName = $schemaName
+                isManaged  = $isManaged
+                slug       = 'flightcheck-only'
+                folder     = ''
+            }
+            agents             = @()
```
nehaoss: **`agents` array is empty while `agent` object has data - inconsistent config state**

This sets `agents = @()` (empty) while populating the `agent` object with the selected agent's details. This works today because `flightcheck/cli.py` has a backward-compatibility fallback that promotes `agent` to `agents` when the array is empty.

However, the `agents` array is the canonical location per `setup.py`'s `write_config()`. Any new code that reads `agents` directly (without the compat fallback) will see zero agents and skip agent-specific logic - even though the user selected one during the FlightCheck installer.

**Suggestion:** When the user selects an agent, append it to the `agents` array and set `activeAgent` to its slug, matching the structure `setup.py` produces.

nehaoss (commented):
The bootstrap script, Invoke-Native helper, Resolve-Python, git error handling, pip dependency installation, and discover.py/auth.py encoding fixes are all well-implemented. Found two issues in the FlightCheck-only config generation.

GrahamMcMynn (dismissed):

amilandi:
> @amilandi this is cool. Love the simplified run for FlightCheck installing the dependencies. Couple of questions:
> 
> 1. Does the login for a tenant need to be Power Platform Admin or a simple maker privileges work?
> 2. Like the option for running 'environment-wide' check without selecting an agent, can a tenant-wide check to be introduced right before environment selection to run tenant-wide pre-reqs. like licensing, role permissions, etc.?
> 3. Can this be also tested with few edge cases like where a maker/admin tries to run this in desktop that's managed by organization policies or UAC restricting executables/installers?
> 
> @GrahamMcMynn the tenant and environment pre-reqs are already covered in PowerShell version and can the converted python scripts be integrated in the flow too? Just confirming

1 (Admin vs Maker privileges): The installer itself only needs permissions toinstall packages via winget (local machine admin or user-scoped installs). TheFlightCheck discovery step (listing environments/agents) uses delegated auth — itwill show whatever environments the signed-in user has access to. A Power PlatformAdmin will see all environments; a maker will only see environments they have arole in.

2 (Tenant-wide pre-reqs before environment selection): Good idea — this is aFlightCheck enhancement rather than an installer change. Out of scope for this PRbut worth tracking as a follow-up issue.

3 (Corporate policies/UAC restricting installs): This PR already handles several of those cases: winget is optional in FlightCheckOnly mode if Python+Git arepre-installed, and the script provides manual install URLs when winget fails. Wehaven't tested every possible corporate lockdown scenario, but the fallback pathsare in place. Happy to address specific failure cases as they come up.

GrahamMcMynn (approved):

PR #117, opened 2026-06-03
Skip winget upgrade for already-installed tools & suppress parent-repo prompt

amilandi:
## Changes

1. **Skip winget install for already-present tools** — The pre-install check (`Resolve-Python` / `Get-Command`) was previously gated behind `-FlightCheckOnly`. In the full install path, winget always attempted an upgrade which throws if the upgrade fails (e.g. VS Code running or installed via a different channel). Now all modes skip tools that are already functional on PATH.

2. **Suppress VS Code parent-repo notification** — Set `git.openRepositoryInParentFolders` to `never` in the maker workspace settings so users aren't prompted to open the parent repo (which would break Copilot skills).

## Testing

Ran the full installer locally with VS Code already installed — installer correctly skips it with `[ok]` instead of attempting a winget upgrade that fails.

### DIFF
--- a/setup/Install-EssAdk.ps1
+++ b/setup/Install-EssAdk.ps1
@@ -245,12 +245,10 @@ if (-not $wingetAvailable) {
     # logs "already installed" for present packages.
     foreach ($pkg in $packages) {
         # Skip if already installed (avoids unnecessary winget calls + elevation prompts)
-        if ($FlightCheckOnly) {
-            $existing = if ($pkg.Cmd -eq 'python') { Resolve-Python } else { Get-Command $pkg.Cmd -ErrorAction SilentlyContinue }
-            if ($existing) {
-                Write-Ok "$($pkg.Name) (already installed)"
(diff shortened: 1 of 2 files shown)


nehaoss (commented):
Overall this is a clean, well-scoped change. Two observations:

1. **Behavioral change in full-install mode**: Previously, the skip-if-installed logic only ran in FlightCheckOnly mode. Now it runs unconditionally. This means a full install will also skip \winget install\ for present tools. This is likely intentional (avoids elevation prompts for tools already installed), but worth confirming: if a user has an *older* version of Python, should we still attempt \winget upgrade\? The current logic checks for the command's existence, not version compatibility.

2. **VS Code settings addition**: The \git.openRepositoryInParentFolders: never\ suppression is a nice UX improvement for the nested workspace. 👍

avneeshraiusit (approved):

qazizaahirah (approved):

PR #118, opened 2026-06-03
Add GitHub Codespaces support for the ESS Maker Kit

amilandi:
## Summary

Adds a GitHub Codespaces option for users who want the full maker kit experience without installing VS Code or toolchain locally. Also cleans up READMEs to be customer-facing ready.

### Codespaces support
- **`.devcontainer/devcontainer.json`** — Pre-configured dev container with Python 3.12, pip dependencies (absolute path), GitHub Copilot extensions, and GitHub CLI
- **`.gitignore`** — Added `.devcontainer/` to allowlist
- **`solutions/ess-maker-skills/.vscode/settings.json`** — Added `git.openRepositoryInParentFolders: never` to suppress parent-repo prompt

### README updates
- **`README.md`** (root) — New Getting Started table linking all setup options (one-shot installer, Codespaces, FlightCheck-only, manual setup)
- **`setup/README.md`** — Added Codespaces section with free tier info (60hrs/month on 2-core), open-folder instruction, and pricing link. Removed internal-only sections (design choices, open items, redundant customer commands)
- **`solutions/ess-maker-skills/README.md`** — Quick Start updated with Codespaces as Option B

### Removed
- **`setup/bootstrap-codespaces.ps1`** — Removed; target users prefer a direct browser link over a PowerShell command

### Known limitation
- `workspaceFolder` in devcontainer.json is not honored by GitHub Codespaces for subfolder paths (known VS Code bug). Users are instructed to open `solutions/ess-maker-skills` via File > Open Folder after launch.

## Validation

- Created a Codespace from this branch — confirmed pip deps installed, Copilot extension active, `/setup` working after opening the subfolder
- 2-core machine type is sufficient


nehaoss (commented):
Solid addition — Codespaces support is a great DX improvement. A few observations:

1. **devcontainer.json**: The \postCreateCommand\ installs pip deps and prints a welcome message — clean approach. Consider adding \--user\ flag or confirming the container runs as non-root (the mcr.microsoft.com/devcontainers/python image typically does, so this is likely fine).

2. **Removed sections from setup/README.md**: The 'Customer-facing commands' and 'Design choices worth flagging in review' sections were deleted (lines 111–148 in old file). These contained useful context about design decisions (idempotent behavior, pinned Python version, no admin elevation). If this info is documented elsewhere that's fine, otherwise consider preserving the 'Design choices' notes in a contributor doc.

3. **README getting-started table**: Clean presentation. The 'paid feature' callout for Codespaces is appropriately placed.

4. **Minor**: The Codespace link hardcodes \
ef=main\ — this means users on feature branches would still get main. That's probably correct for the published getting-started flow.

setup/README.md
```diff
@@ -18,6 +18,16 @@ iex (irm https://raw.githubusercontent.com/microsoft/Employee-Self-Service-Agent
 
 > **GitHub Copilot subscription is still required** for the in-editor maker experience. This script installs the toolchain and extension scaffolding; it does not (and cannot) grant the Copilot entitlement.
 
+## GitHub Codespaces (no local install)
+
+> **GitHub Codespaces is a paid feature.** Usage is billed to your GitHub account or organization. See [GitHub Codespaces pricing](https://github.com/features/codespaces#pricing) for details.
```
avneeshraiusit: We get some free quota. So, lets rephrase to ensure it doesn't deter the customer. We can provide the link to Codespaces policy.

README.md
```diff
@@ -12,6 +12,19 @@ A monorepo of solutions, samples, and tooling for the Microsoft Employee Self-Se
...
 Additional solutions (FlightCheck standalone, evaluation harnesses, etc.) will be added under `solutions/` over time.
 
+## Getting Started
+
+There are several ways to set up your environment depending on your needs:
+
+| Option | Best for | Guide |
+|--------|----------|-------|
+| **One-shot installer** (Windows) | Full maker kit — installs VS Code, Python, Git, and all dependencies | [Setup README](setup/README.md) |
+| **GitHub Codespaces** | Browser-based development — no local install required ([paid feature](https://github.com/features/codespaces#pricing)) | [Setup README](setup/README.md#github-codespaces-no-local-install) |
```
avneeshraiusit: Same here for rewording 'paid'

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "ESS Maker Kit",
```
avneeshraiusit, line 2: Thought:
Codespaces can be used not just for maker but also for flight check with appropriate dependencies installed?
Should we provide that as an option as well in flight check section?

setup/README.md
```diff
@@ -18,6 +18,19 @@ iex (irm https://raw.githubusercontent.com/microsoft/Employee-Self-Service-Agent
...
+> **Free tier:** Personal GitHub accounts include 60 hours/month of Codespaces on a 2-core machine (120 core-hours). Compute is only billed while the Codespace is running — it stops automatically after 30 minutes of inactivity. See [GitHub Codespaces pricing](https://github.com/features/codespaces#pricing) for details.
+
+For users who prefer a cloud-based development environment — no local toolchain or VS Code desktop install required. Just a browser and a GitHub account:
+
+👉 [**Create Codespace**](https://github.com/codespaces/new?repo=microsoft/Employee-Self-Service-Agent-Developer-Kit&ref=main&devcontainer_path=.devcontainer%2Fdevcontainer.json)
+
+The Codespace comes pre-configured with Python 3.12, pip dependencies, and GitHub Copilot. Select the **2-core** machine type (sufficient for the maker kit). Once it starts:
+
+1. Open the `solutions/ess-maker-skills` folder (File → Open Folder → `/workspaces/Employee-Self-Service-Agent-Developer-Kit/solutions/ess-maker-skills`)
+2. Run `/setup` in Copilot Chat to connect your Dataverse environment
```
avneeshraiusit, line 32: Suggestion [Can be a follow-up PR]: Is it beneficial to auto trigger the setup command when the codespace gets loaded? And have some context on the next steps listed in the chat window?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,25 @@
...
+  "image": "mcr.microsoft.com/devcontainers/python:3.12",
+
+  "workspaceFolder": "/workspaces/Employee-Self-Service-Agent-Developer-Kit/solutions/ess-maker-skills",
+
+  "postCreateCommand": "pip install --quiet --disable-pip-version-check -r /workspaces/Employee-Self-Service-Agent-Developer-Kit/solutions/ess-maker-skills/scripts/requirements.txt && echo '' && echo '=== ESS Maker Kit ready! ===' && echo 'Open File > Open Folder and select solutions/ess-maker-skills, then run /setup in Copilot Chat.' && echo ''",
```
avneeshraiusit, line 8: How does it validate that all the dependencies got successfully installed?

avneeshraiusit, line 8: For my understanding - where have we specified the full list of dependencies like python runtime, Github CoPilot, VSCode etc that is required to be installed in the docker?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "ESS Maker Kit",
+  "image": "mcr.microsoft.com/devcontainers/python:3.12",
+
+  "workspaceFolder": "/workspaces/Employee-Self-Service-Agent-Developer-Kit/solutions/ess-maker-skills",
```
qazizaahirah: Add a comment here why the hard coded space

.gitignore
```diff
@@ -6,6 +6,7 @@
 /*
 
 # Allow our top-level folders and root files
+!.devcontainer/
```
qazizaahirah, line 9: nit: Sometimes when you gitignore the entire folder, its  a security risk because secrets can be tracked within the folder without knowing

avneeshraiusit (dismissed):

amilandi:
Addressed the review feedback in the latest commit:

**@avneeshraiusit:**
1. **'Paid' rewording** — Changed to 'Free tier available' with a direct link to the GitHub billing docs (both in setup/README.md and root README.md). Should be more encouraging without misleading about potential costs for org accounts.
2. **FlightCheck via Codespaces** — Good idea! The current devcontainer installs full pip dependencies which covers FlightCheck too. We could add an explicit FlightCheck section pointing users to the Codespace link. I'll note this as a follow-up.
3. **Auto-trigger /setup on load** — Agreed this would be great UX. Noting as a follow-up PR since it requires Copilot Chat API integration.
4. **Dependency validation** — The \postCreateCommand\ uses \pip install\ which will exit non-zero if any package fails. The Codespace creation log shows the full output. If it fails, the container still starts but the user sees the error prominently in the creation log. Added a comment in devcontainer.json explaining this.
5. **Full dependency list** — The Python deps come from \scripts/requirements.txt\. The container image (\mcr.microsoft.com/devcontainers/python:3.12\) provides the Python runtime, git, and common dev tools. VS Code and Copilot extensions are specified in the \customizations.vscode.extensions\ array. Added comments to clarify this flow.

**@qazizaahirah:**
6. **Comment on workspaceFolder** — Added explanatory comment noting the known Codespaces limitation.
7. **gitignore security** — Good catch. Narrowed the allowlist from \!.devcontainer/\ (entire folder) to \!.devcontainer/devcontainer.json\ (specific file only). This prevents secrets or other files from being accidentally committed under that directory.

setup/README.md
amilandi: Fixed — rephrased to free tier available with billing docs link.

README.md
amilandi: Fixed — same rewording applied here, now says 'free tier available' with the billing docs link.

setup/README.md
amilandi: As of now it's not possible to do this since the user has to manually switch to the correct VSCode workspace before they can run /setup, but it's possible for the one shot installer. I created https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7454935 to track

.devcontainer/devcontainer.json
amilandi: postCreateCommand uses pip install which exits non-zero if any package fails. Codespaces surfaces this in the creation log — the container still starts but the user sees the error. There isn't a separate validation step beyond pip's own exit code.

.devcontainer/devcontainer.json
amilandi: Removed this line entirely — workspaceFolder is ignored by GitHub Codespaces (known VS Code bug microsoft/vscode-remote-release#7798). It was misleading dead code. Users are instructed to open the subfolder manually in step 1.

.gitignore
amilandi: Good catch — fixed. Changed to a 3-rule pattern that only allows devcontainer.json specifically. Any other files placed in .devcontainer/ (secrets, env files, etc.) will remain ignored.

setup/README.md
amilandi: Fixed — rephrased to 'Free tier available' with a direct link to the GitHub billing docs.

README.md
amilandi: Fixed — same rewording applied here.

.devcontainer/devcontainer.json
amilandi: Added a 'FlightCheck via Codespaces' subsection to the README. The same Codespace has all FlightCheck deps pre-installed — users just open a terminal and run the CLI directly.

.devcontainer/devcontainer.json
amilandi: postCreateCommand uses pip install which exits non-zero if any package fails. Codespaces surfaces this in the creation log — the container still starts but the user sees the error prominently. There isn't a separate validation step beyond pip's own exit code.

amilandi: Added a full Dependencies section to setup/README.md that documents all system tools, Python packages, VS Code extensions, and Codespaces environment components with their purposes and where they're specified.

.devcontainer/devcontainer.json
amilandi: Removed workspaceFolder entirely — it's ignored by GitHub Codespaces (known VS Code bug microsoft/vscode-remote-release#7798). Users are instructed to open the subfolder manually in step 1.

.gitignore
amilandi: Good catch — changed to a 3-rule pattern that only allows devcontainer.json specifically. Any other files placed in .devcontainer/ (secrets, env files, etc.) will remain ignored.

qazizaahirah (approved):

PR #119, opened 2026-06-03
Fix one-shot winget installer reliability

amilandi:
## Summary

Fixes several reliability issues with the one-shot winget installer (ootstrap.ps1 → Install-EssAdk.ps1) discovered during real-world testing on clean machines and PS 5.1.

### Bug fixes
- **Fix Test-Path '' crash on PS 5.1**: PowerShell 5.1's `-or` does not short-circuit, so `Test-Path ''` throws a parameter binding error when `` is empty (pre-clone scenario). Fixed by using `-and` which does short-circuit.
- **Fix pip dependency list**: Added all required packages (google-adk, zure-identity, etc.) so that `/setup` works without needing to re-install manually.

### Improvements
- **Elapsed-time progress indicator**: Replaced the broken runspace-based spinner (doesn't work in ScriptBlock::Create context) with a simple elapsed-time display (e.g., `[12s] installing Python 3.12 ...`) that works in both PS 5.1 and PS 7.
- **Deferred pip install**: When Python is freshly installed and `requirements.txt` isn't available yet (pre-clone), pip install is deferred until after the repo is cloned.
- **Simplified README**: Removed execution-policy notes (bootstrap handles it), streamlined instructions.

## Testing

Tested on Windows 11 with:
- Clean machine (Python, VS Code uninstalled) — full install flow
- Already-installed machine — skip/verify flow
- PowerShell 5.1 and PowerShell 7


setup/bootstrap.ps1
```diff
@@ -57,7 +57,13 @@ foreach ($f in $files) {
 }
 
 $installer = Join-Path $tempDir 'Install-EssAdk.ps1'
+
+# Run the installer in-memory (as a script block) so execution policy never
+# applies — the script content is never "executed from disk".
+$scriptContent = Get-Content $installer -Raw
+$scriptBlock = [ScriptBlock]::Create($scriptContent)
+
```
nehaoss, line 65: Nice approach - `[ScriptBlock]::Create()` sidesteps execution policy cleanly since it's treated as in-memory code, not a disk-based script. I verified there are no other `$PSScriptRoot` or `$MyInvocation` references in the installer beyond the two you've guarded - good coverage.

setup/Install-EssAdk.ps1
```diff
@@ -228,8 +228,8 @@ if (-not $wingetAvailable) {
     # Declarative path - requires `winget configure --enable` (one-time opt-in
     # that pulls DSC modules from the Microsoft Store). Provided for IT shops
     # that prefer the auditable YAML manifest.
-    $configFile = Join-Path $PSScriptRoot 'ess-adk-setup.winget.yaml'
-    if (-not (Test-Path $configFile)) {
+    $configFile = if ($PSScriptRoot) { Join-Path $PSScriptRoot 'ess-adk-setup.winget.yaml' } else { '' }
```
nehaoss, line 251: When `$PSScriptRoot` is empty (ScriptBlock invocation), this throws `Configuration file not found: ` with a blank path, which is confusing. Consider making the message more actionable for the empty-path case, e.g.: `Cannot locate winget config - $PSScriptRoot is empty. -UseDsc requires running Install-EssAdk.ps1 directly from disk, not via bootstrap.`

This tells the user *why* it failed and what to do about it.

setup/Install-EssAdk.ps1
```diff
@@ -304,8 +304,12 @@ if (-not $pythonExe) {
     $deferPip = $false
 } else {
     Write-Ok "Using Python: $pythonExe"
-    $requirementsFile = Join-Path $PSScriptRoot '..\solutions\ess-maker-skills\scripts\requirements.txt'
-    if (-not (Test-Path $requirementsFile)) {
+    if ($PSScriptRoot) {
```
nehaoss, line 329: The fallback logic here is good - when `$PSScriptRoot` is empty, `$deferPip` is set to `true` and pip install is retried after clone. This is correct since the bootstrap flow always clones the repo downstream.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "ESS Maker Kit",
+  "image": "mcr.microsoft.com/devcontainers/python:3.12",
+
+  "workspaceFolder": "/workspaces/Employee-Self-Service-Agent-Developer-Kit/solutions/ess-maker-skills",
+
+  "postCreateCommand": "pip install --quiet --disable-pip-version-check -r /workspaces/Employee-Self-Service-Agent-Developer-Kit/solutions/ess-maker-skills/scripts/requirements.txt && echo '' && echo '=== ESS Maker Kit ready! ===' && echo 'Open File > Open Folder and select solutions/ess-maker-skills, then run /setup in Copilot Chat.' && echo ''",
```
nehaoss, line 8: Two suggestions:

1. **Readability**: This is a long inline command. Consider extracting to a shell script (e.g. `setup/codespace-setup.sh`) for easier maintenance.

2. **Hardcoded repo name**: The path `/workspaces/Employee-Self-Service-Agent-Developer-Kit` is hardcoded here and in `workspaceFolder`. If someone forks the repo under a different name, this breaks silently.

setup/README.md
```diff
@@ -1,22 +1,27 @@
...
+For users who prefer a cloud-based development environment — no local toolchain or VS Code desktop install required. Just a browser and a GitHub account:
 
-...and they land in VS Code at `solutions/ess-maker-skills/` with everything wired up except the final Dataverse auth step (`/setup` in Copilot Chat).
+👉 [**Create Codespace**](https://github.com/codespaces/new?repo=microsoft/Employee-Self-Service-Agent-Developer-Kit&ref=main&devcontainer_path=.devcontainer%2Fdevcontainer.json)
 
-> **GitHub Copilot subscription is still required** for the in-editor maker experience. This script installs the toolchain and extension scaffolding; it does not (and cannot) grant the Copilot entitlement.
+The Codespace comes pre-configured with Python 3.12, pip dependencies, and GitHub Copilot. Select the **2-core** machine type (sufficient for the maker kit). Once it starts:
+
+1. Open the `solutions/ess-maker-skills` folder (File → Open Folder → `/workspaces/Employee-Self-Service-Agent-Developer-Kit/solutions/ess-maker-skills`)
+2. Run `/setup` in Copilot Chat to connect your Dataverse environment
```
nehaoss, line 24: The `devcontainer.json` already sets `workspaceFolder` to `solutions/ess-maker-skills`, so the Codespace should open in the right folder automatically. If that's the case, step 1 here (Open Folder) is unnecessary and could confuse users. Consider simplifying to just: 'Once it starts, run /setup in Copilot Chat to connect your Dataverse environment.'

nehaoss (commented):
## Overall Assessment

Solid fix for the execution policy issue. The `[ScriptBlock]::Create()` approach is the right pattern - it avoids execution policy entirely without requiring `-ExecutionPolicy Bypass` in a child process, which as noted can be overridden by GPO. The `$PSScriptRoot` null guards in the installer are a necessary follow-up that's easy to miss.

A few suggestions inline, mostly around error messages and the devcontainer config. The core fix is sound.

setup/README.md
amilandi: that code in devcontainer.json was not working to set the workspace. removed the broken code

nehaoss (dismissed):

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "ESS Maker Kit",
```
avneeshraiusit, line 2: devcontainer.json is unintended in this PR I assume given its part of PR: https://github.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/pull/118

avneeshraiusit (changes requested):

.devcontainer/devcontainer.json
amilandi: fixed.

avneeshraiusit (approved):

PR #123, opened 2026-06-04
flightcheck: WD-PKG-001 + WD-CONN-012 — detect Workday install flavor

GrahamMcMynn:
## Summary

Adds two new FlightCheck checks that distinguish the **simplified** (OOTB) Workday integration from the **full / legacy SOAP+custom** integration, so downstream Workday checks can branch on which package shape the customer has installed.

- [Microsoft Learn — Workday simplified setup](https://learn.microsoft.com/en-us/microsoft-365/copilot/employee-self-service/workday-simplified-setup)
- [Microsoft Learn — Workday full / legacy setup](https://learn.microsoft.com/en-us/microsoft-365/copilot/employee-self-service/workday)

### `WD-PKG-001` — Workday install-flavor detection

Queries Dataverse `connectionreferences` filtered to the `shared_workdaysoap` connector, extracts the trailing 5-hex suffix from each `connectionreferencelogicalname`, and matches the resulting set against the Microsoft-shipped fingerprints:

| Suffix set | Verdict | Reason |
|---|---|---|
| `{ff0df}` | **simplified** (PASS) | 1 ref: OAuthUser / OBO |
| `{ff0df, 0786a, d6081}` | **full** (PASS) | 3 refs: OBO + 2 ISU roles |
| (empty) | **none** (NOT_CONFIGURED) | No Workday integration |
| strict subset of legacy | **partial** (FAIL) | Incomplete install |
| anything else | **unknown** (WARNING) | Customer-modified or future MS release |

**Design notes** (per rubber-duck review):

- **Suffix-only matching** keeps the check resilient to publisher-prefix changes — a customer cloning the solution under a custom prefix retains the same Microsoft-shipped suffix.
- **Both flavors share `ff0df`** so a `{ff0df}`-only shape is inherently ambiguous (could also be a partial / broken full install). The result text surfaces this explicitly. Closing the ambiguity deterministically would require an install-time-identity signal (e.g. `solutions`-table ownership of the ref) and is deferred.
- We considered `environmentvariabledefinitions` as a second signal but it would false-FAIL on the legacy → simplified upgrade case (env var defs survive uninstall as data residue, indistinguishable from a partial install). Deferred for the same reason.

The check sets `runner._workday_package_flavor` and caches `runner._workday_connection_refs` for downstream consumers.

### `WD-CONN-012` — Package connection-reference binding completeness

Reads the cached refs from WD-PKG-001 and verifies every connection reference Microsoft shipped for the detected flavor is bound to an active connection (`connectionid` present, `statuscode == 1`).

Intentionally a **Dataverse-side binding-completeness check**, NOT a BAP-side connection-health check; `WD-CONN-001..NNN` and `WD-CONN-101` remain the home for BAP health and OAuth token validation.

### Pipeline ordering

`run_workday_checks` is restructured so `WD-PKG-001` runs first (whenever Dataverse credentials are available, **independent of flow detection**) and `WD-CONN-012` runs last. Early-return now requires both no Workday flows AND no Workday connection refs — so a tenant with refs installed but no flows yet deployed still gets a verdict.

## Evidence

Captured against real test tenants (one simplified, one full / legacy), PII-redacted, and committed:

| File | Shape |
|---|---|
| `tests/fixtures/cassettes/dataverse_workday_connection_refs_simplified.yaml` | 1 Workday ref: `_ff0df` (OAuthUser / OBO) |
| `tests/fixtures/cassettes/dataverse_workday_connection_refs_full.yaml` | 3 Workday refs: `_ff0df`, `_0786a`, `_d6081` |
| `tests/captures/record_dataverse_workday_connection_refs.py` | Recorder that produced them |

The Dataverse API contract itself remains **documented**-tier (per MS Learn `connectionreference` reference); these cassettes are **fingerprint-value evidence**, not contract evidence. `INDEX.md` is updated to reflect this and to correct the expected-count text (was 2 / 4, conflating Dataverse refs with BAP connections; now 1 / 3).

## Bug fix included

While validating the new cassettes, found that the GUID-scrub regex in `tests/captures/_common.py` failed to redact GUIDs preceded by URL-encoded prefixes like `%20<guid>` — the 1-character negative lookbehind matched `0` (last char of `%20`) as a hex digit and suppressed the substitution, leaking real solution GUIDs into URLs.

Widened the lookaround on both sides to 3 characters, closes the leak without breaking the existing 58 redaction tests. Three new regression tests cover the URL-encoded prefix patterns. Scanned all existing committed cassettes for the leak — zero hits, so the fix is forward-looking only.

## Validation

- **Automated**: 230 tests pass (16 new), 8 skipped per mock-validation policy. Ruff lint clean on all touched code paths.
- **Manual**: Both checks validated against a real simplified-install tenant and a real full / legacy-install tenant. WD-PKG-001 correctly classifies each; WD-CONN-012 correctly verifies bindings for the detected flavor.

## File map

```
solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py  (+ ~290 LOC: constants, 2 checks, restructured run_workday_checks)
tests/flightcheck/checks/test_workday_package.py                  (new, 16 tests)
tests/mocks/dataverse.py                                          (+ connection_ref / workday_connection_ref helpers)
tests/fixtures/cassettes/dataverse_workday_connection_refs_*.yaml (new, 2 cassettes)
tests/captures/record_dataverse_workday_connection_refs.py        (new recorder)
tests/captures/_common.py                                         (redactor fix)
tests/test_captures_redaction.py                                  (3 regression tests)
tests/fixtures/cassettes/INDEX.md                                 (row update)
```


tests/captures/record_dataverse_workday_connection_refs.py
```diff
@@ -0,0 +1,252 @@
...
+What DOES distinguish them is the **set of ``connectionreference``
+rows** shipped by the two different Microsoft-owned solutions that
+install the simplified vs. full integration. Connection references are
+emitted by the solution publisher at build time and embedded in the
+solution XML. They exist in Dataverse from the moment the solution is
+installed, regardless of whether the customer has wired up any
+connections — so they're a deterministic install-time fingerprint that
+survives partial wire-up. The simplified solution ships two
+connectionrefs (one OBO, one service-account); the full/legacy
+solution ships four (OBO, two SOAP ISU roles, one auxiliary). The
```
GrahamMcMynn: The recorder added in this PR still references the **old 2 / 4** ref counts that the PR description explicitly says were corrected to **1 / 3**, in three places:

- Module docstring lines 32–34 (this hunk): _"The simplified solution ships **two** connectionrefs (one OBO, one service-account); the full/legacy solution ships **four** (OBO, two SOAP ISU roles, **one auxiliary**)."_
- Operator workflow step 4 (~line 74): _"(2 rows on a simplified tenant, 4 rows on a legacy tenant)"_
- `NEXT STEPS` stdout (~line 209): `print("   you captured against (expect 2 for simplified, 4 for legacy).")`

Everything else in this same PR uses 1 / 3:

- `workday.py`: `SIMPLIFIED_REF_SUFFIXES = {"ff0df"}` (1 ref), `LEGACY_REF_SUFFIXES = {"ff0df","0786a","d6081"}` (3 refs).
- The committed cassettes: `dataverse_workday_connection_refs_simplified.yaml` has 1 Workday-flagged row; `..._full.yaml` has 3.
- `INDEX.md` was updated specifically to fix the 2/4 → 1/3 wording.
- PR body: _"corrected the expected-count text (was 2 / 4, conflating Dataverse refs with BAP connections; now 1 / 3)"_.

An operator running this recorder against a real simplified tenant will see 1 Workday-flagged row, read "expect 2", and either discard a correct capture or go hunting for a phantom "missing" ref. Same "doc rule change partially applied" failure mode the `INDEX.md` edit was trying to close — it just didn't get threaded into the recorder file.

**Fix:** rewrite the three occurrences to "1 / 3" (one OBO ref for simplified; three refs — OBO + two ISU roles — for full / legacy), and drop the _"one auxiliary"_ phrase in the module docstring (the full install ships 3, not 4 — the "4" was the BAP-connection count, not the Dataverse-ref count).

GrahamMcMynn (commented):
Code review by ESS Code Review Agent — 1 comment.

tests/captures/record_dataverse_workday_connection_refs.py
GrahamMcMynn: Good catch — fixed in a8b7862. All three sites now say 1 / 3 (module docstring lines 32-34, operator workflow step 4 around line 78, and the NEXT STEPS stdout). Dropped the inaccurate `one auxiliary` phrasing and added a short parenthetical noting the BAP-side counts ARE 2 / 4 because each install layers an additional unbound service-account connection — that's the origin of the original 2/4 figure, captured here for future readers who see both numbers in the codebase. Tests still 230 passing, ruff clean.

amilandi (approved):

PR #124, opened 2026-06-04
flightcheck: gate ISU/RaaS Workday checks on WD-PKG-001 verdict

GrahamMcMynn:
# flightcheck: gate ISU/RaaS Workday checks on WD-PKG-001 verdict

## Summary

Follow-up to #123. WD-PKG-001 classifies the customer's Workday install as `simplified` (1 connection ref, OBO via the signed-in user) or `full` / `legacy` (3 refs: OBO + 2 ISU service-account refs that drive RaaS reports). Several downstream Workday checks are only meaningful on the full install — on a simplified-install tenant they emit FAILs whose remediations point operators at the wrong setup path.

This PR gates those checks on `runner._workday_package_flavor == "simplified"` so they SKIP cleanly with a message that names WD-PKG-001 as the gating decision and acknowledges the install-shape ambiguity.

## Checks gated

| Check | Why it's ISU/RaaS-specific |
|---|---|
| **WD-ENV-001 / 002 / 003** (`_check_env_vars`) | The three `EmployeeContextRequest*` env vars (`AccountName`, `ReportName`, `ReportInstanceName`) are the three pieces of one RaaS `customreport2` URL + Basic-Auth pair — they're not independent settings. The simplified install replaces this entire RaaS call with an OBO `Get_Workers` SOAP call against the signed-in user's identity, so none of the three are read. Verified against `src/reference/ess-docs/integrations/workday.md` (lines 335-337) + the reference MCP `_raas_url()` / `_raas_auth()` helpers in `src/mcp/workday/client.py`. |
| **WD-ENV-101** (`_check_isu_username_format`) | Validates an ISU service-account username's alignment with Entra UPN format. The simplified install has no ISU service account. |
| **WD-WF-000..017** (`_check_workflows`) | Tests Workday SOAP with ISU credentials. Without gating, the natural credential-missing skip emits `"Workday ISU credentials not provided"` with a remediation telling the operator to provide ISU creds — actively misleading on a tenant that intentionally has no ISU. |

## Checks intentionally NOT gated

- `_check_connections` (WD-CONN-001..NNN) — package-agnostic; iterates ALL Workday connections
- `_check_connection_token_health` (WD-CONN-101) — package-agnostic auth-health
- `_check_flow_status` (WD-FLOW-*) — package-agnostic
- `_check_package_connection_completeness` (WD-CONN-012) — already gated correctly in #123
- `_check_package_flavor` (WD-PKG-001) — the source of truth itself

## Skip semantics (per rubber-duck critique)

The single safety rule: **skip ONLY on a positive match for `"simplified"`**. Any other verdict — `None`, `"full"`, `"partial"`, `"unknown"`, `"none"`, `"skipped"` — runs the existing logic. Operators debugging a broken install need maximum signal, not silence.

- `getattr(runner, "_workday_package_flavor", None)` defaults to `None` → run. This keeps the pre-PR test minimal-runners (no attribute set) on the run-by-default path → zero existing-test regressions.
- The gate is placed **before any API reads** so simplified tenants don't hit Dataverse / Graph unnecessarily and so the flavor-not-applicable SKIP isn't masked by a token-related one.
- The SKIP `remediation` explicitly surfaces the `{ff0df}`-only ambiguity: the 1-ref shape WD-PKG-001 classifies as `simplified` also matches a broken full install where the 2 ISU refs failed to deploy. Without this note an operator who intended the full install could dismiss the SKIP as benign.

## Migration edge cases (intentional behavior)

| Scenario | What happens | Why this is OK |
|---|---|---|
| Simplified → full mid-deploy, only `{ff0df}` so far | WD-PKG-001 reports `simplified`; consumers SKIP | WD-PKG-001's own diagnostic + the SKIP remediation's ambiguity note tell the operator their full install is incomplete. No regression vs. pre-PR behavior. |
| Full → simplified with stale ISU refs left over | WD-PKG-001 classifies as `full` (3 refs still present); consumers run as before | Stale-ref cleanup is out of scope for this PR. |

## AGENTS.md

New design principle **#11** added to `solutions/ess-maker-skills/scripts/flightcheck/AGENTS.md`: **"Gate flavor-specific checks on an install-fingerprint verdict."** Documents the convention for future flavor-gating work (Workday is the canonical example; the principle generalizes to any product that ships under multiple flavors that share the same connector — e.g. ServiceNow if a similar split ever lands). Covers:

- Fingerprint check runs first; canonical attribute name `runner._<product>_package_flavor`
- Skip only on positive incompatible match; ambiguous values RUN (this is the load-bearing safety rule)
- `None`-defaults-to-run is for test backwards-compat; production callers must run the fingerprint first
- SKIP `result` (observation) vs `remediation` (actionable contingency) per principle #8
- Gate must precede any API reads

## Tests

`TestSimplifiedInstallGate` class added to two existing files; one new file for the workflow gate. Each class pins **four** branches:

1. **`flavor == "simplified"` → SKIP** with full message contract verified (priority, `WD-PKG-001` mentioned in `result`, ambiguity-note phrases in `remediation`, `doc_link` points to simplified-setup)
2. **`flavor == "full"` → existing logic runs unchanged** (happy-path mock)
3. **`flavor in ("partial", "unknown", "none", "skipped")` → existing logic runs unchanged** (parametrized; protects against a future careless `if flavor != "full": skip` regression)
4. **Attribute absent → existing logic runs unchanged** (backwards-compat for the existing test minimal-runner shape)

## Verification

- `python -m ruff check solutions/ess-maker-skills/scripts solutions/ess-maker-skills/src/mcp` → All checks passed
- `python -m pytest tests/ -q` → **246 passed, 8 skipped** (16 new tests over the 230-test baseline, 0 regressions)
- Manually validated against both a simplified-install environment and a full-install environment

## Files changed

- `solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py` — helper + 3 gates + docstring updates
- `solutions/ess-maker-skills/scripts/flightcheck/AGENTS.md` — new design principle #11
- `tests/flightcheck/checks/test_workday_env_vars.py` — `TestSimplifiedInstallGate` (5 tests)
- `tests/flightcheck/checks/test_workday_isu_username_format.py` — `TestSimplifiedInstallGate` (4 tests)
- `tests/flightcheck/checks/test_workday_workflows_gate.py` — new file, `TestSimplifiedInstallGate` (2 tests)

### DIFF
--- a/solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py
+++ b/solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py
@@ -605,8 +605,98 @@ def _check_package_connection_completeness(runner) -> list[CheckResult]:
     )]
 
 
+# ─────────────────────────────────────────────────────────────────────────
+# Install-flavor gating helper (see AGENTS.md design principle #11)
+# ─────────────────────────────────────────────────────────────────────────
+#
+# Several Workday checks below are only meaningful on the full / legacy
+# install (the one with the 2 ISU service-account refs + RaaS report
+# env vars). On the simplified install, OBO uses the signed-in user's
+# identity — no ISU, no RaaS — so these checks would false-FAIL with
+# remediations pointing operators at the wrong setup path.
+#
+# We gate them on the WD-PKG-001 verdict cached on
+# `runner._workday_package_flavor`. To avoid suppressing real bugs:
+#
+#   * We skip ONLY on a positive `"simplified"` verdict. Any other
+#     verdict (None, "full", "partial", "unknown", "none", "skipped")
+#     runs the existing logic — operators debugging a broken install
+#     need maximum signal, not silence.
+#   * Tests that build minimal runners without setting the attribute
+#     get the default-None branch and observe the pre-gating behavior
+#     (backwards-compatible).
+#   * The SKIP remediation must acknowledge the `{ff0df}`-only
+#     ambiguity: the same 1-ref shape WD-PKG-001 classifies as
+#     "simplified" ALSO matches a broken full install where the 2 ISU
+#     refs failed to deploy. An operator who intended the full install
+#     needs to see this from the SKIP, not just from WD-PKG-001's
+#     standalone diagnostic.
+
+def _simplified_install_skip(
+    *, checkpoint_id: str, description: str,
+    priority: str = Priority.HIGH.value,
+    category: str = "Workday",
+) -> CheckResult:
+    """Build a SKIPPED CheckResult for an ISU/RaaS check gated on
+    `runner._workday_package_flavor == "simplified"`.
+
+    `category` defaults to "Workday" but can be overridden for the
+    workflow-test check, which uses the distinct "Workday Workflows"
+    category in the report so SKIP / pass rows stay grouped with
+    other WD-WF-* output.
+
+    The message split follows the AGENTS.md `result` / `remediation`
+    contract (principle #8): `result` reports the observation
+    (WD-PKG-001's verdict); `remediation` carries the actionable
+    contingency for an operator who intended the OTHER install flavor.
+    """
+    return CheckResult(
+        checkpoint_id=checkpoint_id,
+        category=category,
+        priority=priority,
+        status=Status.SKIPPED.value,
+        description=description,
+        result=(
+            "WD-PKG-001 detected the simplified Workday install shape "
+            "(1 connection reference, OBO/OAuthUser). This check is "
+            "ISU/RaaS-specific and does not apply on the simplified "
+            "install — OBO uses the signed-in user's identity."
+        ),
+        remediation=(
+            "If you intended to install the full / legacy SOAP+custom "
+            "integration, the same 1-ref shape also matches a broken "
+            "full install where the 2 ISU connection references "
+            "(Generic User, Context Generic User) failed to deploy. "
+            "See WD-PKG-001's diagnostic to confirm your install "
+            "intent before treating this SKIP as benign."
+        ),
+        doc_link=f"{DOC_BASE}/workday-simplified-setup",
+    )
+
+
 def _check_env_vars(runner) -> list[CheckResult]:
-    """Validate Workday environment variables in Dataverse."""
+    """Validate Workday environment variables in Dataverse.
+
+    Gated on `runner._workday_package_flavor`: skipped on
+    `"simplified"` because the three env vars (ISU account name,
+    RaaS report name, RaaS report instance) are only consumed by the
+    full / legacy install's RaaS code path. See
+    `_simplified_install_skip` for the SKIP message contract.
+    """
+    flavor = getattr(runner, "_workday_package_flavor", None)
+    if flavor == "simplified":
+        return [
+            _simplified_install_skip(
+                checkpoint_id=meta["id"],
+                description=meta["description"],
+                priority=(
+                    Priority.CRITICAL.value if meta["critical"]
+                    else Priority.HIGH.value
+                ),
+            )
+            for meta in ENV_VARS.values()
+        ]
+
     results = []
     env_url = runner.env_url
     dv_token = runner.dv_token
@@ -721,7 +811,19 @@ def _check_isu_username_format(runner) -> list[CheckResult]:
     surface when those inputs are formalised; this checkpoint covers
     the static format-alignment gap that can be detected without ISU
     credentials.
+
+    Gated on `runner._workday_package_flavor`: skipped on
+    `"simplified"` because the simplified install has no ISU service
+    account (OBO uses the signed-in user's identity directly). See
+    `_simplified_install_skip` for the SKIP message contract.
     """
+    flavor = getattr(runner, "_workday_package_flavor", None)
+    if flavor == "simplified":
+        return [_simplified_install_skip(
+            checkpoint_id="WD-ENV-101",
+            description="ISU username vs Entra UPN format alignment",
+        )]
+
     results: list[CheckResult] = []
     env_url = runner.env_url
     dv_token = runner.dv_token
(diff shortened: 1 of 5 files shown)


tests/flightcheck/checks/test_workday_workflows_gate.py
```diff
@@ -0,0 +1,94 @@
...
+        configured" SKIP for `WD-WF-000` (because our minimal runner
+        has no MCP / config-resolved Workday URL).
+
+        Pin both that the gate didn't suppress the result AND that
+        the result is the credential-missing one, not the gate's
+        flavor-not-applicable one.
+        """
+        from flightcheck.checks.workday import _check_workflows
+
+        runner = _MinimalRunner()
```
GrahamMcMynn, line 113: `test_attribute_absent_falls_through_to_existing_logic` is sensitive to the developer''s ambient environment. When the gate doesn''t fire, `_check_workflows` calls `_resolve_workday_metadata(runner)`, which reads `os.environ.get("WORKDAY_BASE_URL" / "WORKDAY_TENANT" / "WORKDAY_TEST_EMPLOYEE_ID")`, then `_read_mcp_workday_env()` (parses `.vscode/mcp.json` from CWD), then `runner.config`.

A developer with any of those env vars set (very plausible for someone working on Workday code) or running pytest from a workspace containing a `.vscode/mcp.json` with `WORKDAY_BASE_URL`/`WORKDAY_TENANT` will get a different SKIP message ("No test employee ID provided" / "Workday ISU credentials not provided") and the `"not configured" in r.result.lower()` assertion fails. Worse, with `pytest -s` the `input("  Test Employee ID …")` prompt in `_resolve_workday_metadata` will activate and hang the test. There is no autouse conftest fixture that scrubs these.

Recommend tightening the test:

```python
def test_attribute_absent_falls_through_to_existing_logic(
    self, monkeypatch: pytest.MonkeyPatch, tmp_path: Path,
) -> None:
    monkeypatch.delenv("WORKDAY_BASE_URL", raising=False)
    monkeypatch.delenv("WORKDAY_TENANT", raising=False)
    monkeypatch.delenv("WORKDAY_TEST_EMPLOYEE_ID", raising=False)
    monkeypatch.chdir(tmp_path)
    ...
```

Same isolation applies to the second proposed parametrized test below.

tests/flightcheck/checks/test_workday_workflows_gate.py
```diff
@@ -0,0 +1,94 @@
...
+
+        assert len(results) == 1
+        r = results[0]
+        assert r.checkpoint_id == "WD-WF-000"
+        assert r.status == "Skipped"
+        # The pre-existing credential-missing message (NOT the gate's
+        # flavor message). Catches a regression where the gate text
+        # leaks into the no-attribute branch.
+        assert "WD-PKG-001" not in r.result
+        assert "not configured" in r.result.lower()
```
GrahamMcMynn, line 126: The PR description says "Each class pins **four** branches" (simplified→SKIP, full→runs, ambiguous (`partial`/`unknown`/`none`/`skipped`)→runs, attribute-absent→runs), and that contract enforces the load-bearing safety rule "skip ONLY on a positive `simplified` match" (AGENTS.md principle #11.b). The env_vars and isu_username_format `TestSimplifiedInstallGate` classes pin all four; this one pins only branches 1 and 4.

As written, a future careless rewrite of `_check_workflows` like `if flavor != "full": skip` (or `if flavor in {"simplified", "partial", "unknown"}: skip`) would pass both existing tests here but silently suppress the workflow credential diagnostics on `partial` / `unknown` tenants — exactly the failure mode the safety rule exists to prevent.

Recommend adding a parametrized test mirroring `attribute_absent` but with the attribute set:

```python
@pytest.mark.parametrize("flavor", ["full", "partial", "unknown", "none", "skipped", None])
def test_non_simplified_verdicts_fall_through(
    self, monkeypatch: pytest.MonkeyPatch, tmp_path: Path, flavor,
) -> None:
    # ...same env/cwd isolation as comment 1...
    runner = _MinimalRunner()
    if flavor is not None:
        runner._workday_package_flavor = flavor
    results = _check_workflows(runner)
    assert len(results) == 1
    r = results[0]
    assert r.checkpoint_id == "WD-WF-000"
    assert "WD-PKG-001" not in r.result, f"flavor={flavor!r} leaked gate text"
    assert "not configured" in r.result.lower()
```

GrahamMcMynn (commented):
Code review by ESS Code Review Agent (audit mode)

amilandi (approved):
Clean implementation. The gating logic is well-designed: skip only on positive "simplified" match, default-to-run for any other value including None (backwards-compat with minimal test runners). The shared `_simplified_install_skip` helper keeps the SKIP message contract DRY across all three consumers. Tests are thorough — parametrized ambiguous-verdict tests catch the most likely regression pattern (`if flavor != "full": skip`).

One minor note: the `{ff0df}` references in test docstrings/comments appear to be leftover unicode escapes — might want to replace with the actual descriptive text (e.g. "1-connection-ref-only") for readability, but not a blocker.

solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py
```diff
@@ -605,8 +605,98 @@ def _check_package_connection_completeness(runner) -> list[CheckResult]:
...
+
+    The message split follows the AGENTS.md `result` / `remediation`
+    contract (principle #8): `result` reports the observation
+    (WD-PKG-001's verdict); `remediation` carries the actionable
+    contingency for an operator who intended the OTHER install flavor.
+    """
+    return CheckResult(
+        checkpoint_id=checkpoint_id,
+        category=category,
+        priority=priority,
```
nehaoss, line 656: The `doc_link` points to `{DOC_BASE}/workday-simplified-setup` but every other check in this file links to anchors like `/workday#step-4-environment-variables`. If `/workday-simplified-setup` isn't a published Learn page, operators clicking the SKIP's doc link will hit a 404. Verify the URL exists or use an existing anchor (e.g. `/workday#simplified-install`).

solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py
```diff
@@ -605,8 +605,98 @@ def _check_package_connection_completeness(runner) -> list[CheckResult]:
...
+
+    Gated on `runner._workday_package_flavor`: skipped on
+    `"simplified"` because the three env vars (ISU account name,
+    RaaS report name, RaaS report instance) are only consumed by the
+    full / legacy install's RaaS code path. See
+    `_simplified_install_skip` for the SKIP message contract.
+    """
+    flavor = getattr(runner, "_workday_package_flavor", None)
+    if flavor == "simplified":
+        return [
```
nehaoss, line 688: `_simplified_install_skip` defaults `category='Workday'` which happens to match this call site — but this is implicit coupling. If someone changes the helper's default without updating this comprehension, report grouping breaks silently. Consider passing `category='Workday'` explicitly here (same pattern as `_check_workflows` which explicitly passes `category='Workday Workflows'`).

PR #125, opened 2026-06-04
flightcheck: WD-CONN-010  Workday single-Entra-tenant federation alignment (MANUAL)

### LINKED ISSUE #43: FlightCheck: Add check for Workday tenant single Entra integration constraint

## What should be checked

Workday does not support an Entra Integrated connection from two different Entra tenants to the same Workday tenant simultaneously. When a second Entra tenant is wired up (for example, an "admin" or "test" tenant added on top of an existing production EmployeeHub tenant), the existing tenant's federation breaks and previously working Entra Integrated Workday connections fail. Customers and internal test environments have hit this when standing up additional environments against the same Workday tenant.

FlightCheck should detect that the target Workday tenant already has an active Entra Integrated federation to a *different* Entra tenant before allowing or recommending a new Entra Integrated Workday connection in the current environment. If a second integration is detected (or recently configured), the check should warn that this configuration is unsupported and will break the existing connection.

## Suggested implementation

- **Checkpoint ID:** WD-CONN-010
- **Category:** Workday
- **Priority:** HIGH
- **Validation logic:** For the Workday tenant referenced by the environment's Workday connections, query Workday tenant security configuration (or the Workday connection metadata) to discover any existing Entra federation/redirection setup. If a federation is configured pointing at a different Entra tenant ID than the current environment's tenant, fail the check and explain that Workday only supports one Entra tenant integration at a time. Recommend reverting the conflicting integration via "Edit Tenant Setup - Security" in Workday before proceeding.

## Context

### LINKED ISSUE #84: FlightCheck: Add check for SAML NameID alignment with Workday user identifier

GrahamMcMynn:
Implements the FlightCheck check requested in #43.

**Checkpoint:** WD-CONN-010
**Category:** Workday
**Priority:** High
**Status emitted:** Manual (per AGENTS.md design principle #2  the kit gathers everything programmatically observable on the Entra side and delegates the Workday-side comparison to the operator; MANUAL does not fail readiness)
**Conditional?** No  runs before the no-Workday early-return gate in 
un_workday_checks so the warning fires for the pre-install / foreign-tenant scenario the issue describes.

## Why this is now shippable (was BLOCKED on 2026-05-15)

The previous BLOCKED comment correctly observed that no Workday API surface exposes the tenant's SAML Identity Provider configuration  the SOAP `Identity_Management.Get_Authentication_Policies` returns HTTP 500 across v36/v40/v44, the REST `/ccx/api/security/v1/{tenant}/authentication-policies` returns `Invalid request`, and the WQL admin surface is gated by the documented chicken-and-egg auth blocker.

Since that BLOCKED was filed, commit `53f3762` (AUTH-006, fixes #84) introduced the `MANUAL` status pattern specifically for this case. `solutions/ess-maker-skills/scripts/flightcheck/AGENTS.md` design principle #2 explicitly calls AUTH-006 the canonical pattern for *"Workday/ServiceNow/SAP tenant-config validation that has no programmatic admin surface"*  exactly issue #43's situation. WD-CONN-010 is the same shape applied to the foreign-tenant federation scenario.

## API tier

**Microsoft Graph v1.0  `validatable` tier per `tests/fixtures/cassettes/INDEX.md`.**

No cassette / no registry changes needed. Reuses the existing `tests/mocks/graph.py` `service_principal()` / `list_service_principals()` / `insufficient_permissions()` builders (already validatable-tier, CSDL-cited, and exercised by AUTH-005 + AUTH-006). The `service_principal()` docstring is updated to list `WD-CONN-010` as a consumer per the cardinal-rule traceability convention.

## What the check observes vs. delegates

**Entra side (observed via Microsoft Graph):**
- Current Entra tenant ID (`runner.graph.tenant_id`).
- Every federated Workday enterprise app in this Entra tenant  same SAML filter as AUTH-006: `startswith(displayName,'Workday') and preferredSingleSignOnMode eq 'saml'`.
- Each app's SAML entity IDs from `servicePrincipalNames` (the Workday "Service Provider ID" join key).

**Workday side (MANUAL  operator):**
1. Open Workday "Edit Tenant Setup - Security"  "SAML Identity Providers", find the enabled IdP row covering the ESS environment, note its "Service Provider ID".
2. Read the IdP's "Issuer" / federation metadata URL (Entra-issued values have the form `https://sts.windows.net/{tenantId}/`). Compare the embedded `{tenantId}` against the current Entra tenant surfaced in the result. If they DIFFER, configuring an Entra Integrated Workday connection from this tenant will silently break the foreign tenant's federation (Workday supports exactly one Entra-tenant integration per Workday tenant).

## Pattern decisions (informed by rubber-duck review)

- **NOT_CONFIGURED wording** explicitly does NOT claim "the conflict scenario doesn't apply" when no local Workday SAML app is found  a foreign tenant could already own the Workday-side federation, so the remediation walks the operator through the manual Workday-side verification for the pre-install / foreign-tenant scenario as well.
- **Tenant-ID comparison points at the SAML issuer URL**, not the entity ID / appId / cert subject. The Entra tenant ID is reliably embedded in the SAML issuer (`https://sts.windows.net/{tenantId}/`); the entity IDs serve as join keys to identify the active local Entra app, not as proof of tenant identity.
- **Permission probe pattern inherited from AUTH-006**: probe `/servicePrincipals` with `graph.get()` first so a missing `Application.Read.All` consent surfaces as `WARNING` with a remediation pointing at the specific permission needed  never as a silent `NOT_CONFIGURED`.
- **Placement before the no-Workday early-return gate**: the conflict applies pre-install too, so gating on existing Workday refs would defeat the purpose. A wiring/placement regression test pins this contract  a future refactor that moves the call below the gate fails the test loudly.
- **Helper duplication over cross-module imports**: `_WORKDAY_SAML_SP_FILTER` and `_saml_entity_ids` mirror AUTH-006 (`checks/authentication.py`) and are duplicated locally with a comment crediting AUTH-006, so the Workday-category check stays self-contained.

## Tests

`tests/flightcheck/checks/test_workday_entra_federation.py` covers:

- **NOT_CONFIGURED** — no local Workday SP; guards against falsely claiming safety; confirms remediation still walks the operator through Workday-side verification for the foreign-tenant scenario.
- **MANUAL with one app**  entity IDs surfaced, current Entra tenant ID in both `result` and `remediation`, issuer-URL pivot present, conflict warning present, doc_link points at the MS Learn Workday SSO tutorial.
- **Multi-app coalesce**  N apps  1 row (principle 7); both entity IDs listed; GUID noise filtered out of the entity-IDs CSV.
- **403 on /servicePrincipals  WARNING** with `Application.Read.All` remediation; no "Detected apps" list emitted (silent-failure guard).
- **graph=None  SKIPPED** (clean degradation, no crash).
- **Wiring/placement regression guard**  runs `run_workday_checks` with empty flows + flavor=none and asserts WD-CONN-010 appears (proving placement before the early-return gate) AND downstream Workday checks do NOT appear (proving the gate still fires correctly).

Local suite: **257 passed, 8 skipped, 0 regressions.** `ruff check` clean on all touched files.

Fixes #43


amilandi (commented):
**Overall**: Solid implementation. The MANUAL pattern is well-applied, the probe-before-filter pattern correctly prevents the silent-NOT_CONFIGURED trap, and the wiring/placement test is smart insurance against future refactors. A few observations:

---

**1. Duplicated `_saml_entity_ids` (repo instruction violation)**

The repo's copilot-instructions explicitly prohibit duplicating logic across files (see "No duplicate functions" in the root instructions). The PR acknowledges this in comments ("duplicated rather than imported to keep this Workday-category check self-contained") but that's exactly the drift-risk the rule exists to prevent. Consider extracting to a shared location (e.g. `checks/_saml_utils.py`) that both `authentication.py` and `workday.py` import from.

**2. Double `/servicePrincipals` call**

The probe (`$top=1`, no filter) and then the filtered `get_service_principals()` call means two HTTP round-trips. Could the probe be skipped by inspecting the return value of `get_service_principals()` directly? The comment says `get_all()` swallows 401/403 into an empty list — if that's the underlying issue, fixing `get_all()` to propagate status codes would be a cleaner long-term fix (benefits all consumers). Not a blocker for this PR, but worth a follow-up.

**3. `_WORKDAY_SAML_SP_FILTER` also duplicated**

Same filter string exists in `authentication.py` as `_SAML_SP_FILTER`. Same concern as point 1.

---

None of these are blocking — the logic is correct and tests are comprehensive. Approve with the suggestion to extract the shared helpers in a follow-up.

amilandi (approved):

PR #126, opened 2026-06-05
Add macOS one-shot installer (Homebrew-based)

amilandi:
## Summary

Adds a macOS equivalent of the Windows one-shot installer, using Homebrew for toolchain management.

## New files

- `setup/install-ess-adk.sh` - Main orchestrator (installs Homebrew, Python 3.12, Git, gh, VS Code, pip deps, clones repo, launches VS Code)
- `setup/bootstrap-mac.sh` - One-liner entry point (full maker kit)
- `setup/bootstrap-flightcheck-mac.sh` - One-liner entry point (FlightCheck only)

## Usage

**Full maker kit:**
`ash
/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/main/setup/bootstrap-mac.sh)"
``n
**FlightCheck only:**
`ash
/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/main/setup/bootstrap-flightcheck-mac.sh)"
``n
## Key design decisions

- **PEP 668 compliance**: Uses a virtualenv (.venv/) instead of pip install --user to avoid externally-managed-environment errors on modern macOS
- **Apple Silicon + Intel**: Handles both Homebrew paths (/opt/homebrew vs /usr/local)
- **Git via Xcode CLT**: Skips brew install git if git is already available (common via Xcode Command Line Tools)
- **Copilot extensions as hard errors**: Fails the install if Copilot extensions can't be installed (user needs a valid subscription)
- **FlightCheck mode**: Installs only Python + Git, clones repo, runs cli.py --scope full

## README updates

- Root README: Added macOS row to Getting Started table
- setup/README: Added macOS testing section, fixed gh CLI dependency table, added macOS re-run command

### DIFF
--- a/setup/README.md
+++ b/setup/README.md
@@ -1,9 +1,17 @@
 # ESS ADK — One-Shot Installer
 
-A single PowerShell command that installs everything needed for the ESS Maker Kit: VS Code, Python 3.12, Git, GitHub CLI, Copilot extensions, pip dependencies, and clones the repo.
+A single command that installs everything needed for the ESS Maker Kit: VS Code, Python 3.12, Git, GitHub CLI, Copilot extensions, pip dependencies, and clones the repo.
+
+**Windows** (PowerShell):
 
 ```powershell
 iex (irm https://raw.githubusercontent.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/main/setup/bootstrap.ps1)
+```
+
+**macOS** (Terminal):
+
+```bash
+/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/main/setup/bootstrap-mac.sh)"
 ```
 
 Once complete, VS Code opens at `solutions/ess-maker-skills/`. Run `/setup` in Copilot Chat to connect your Dataverse environment.
@@ -36,10 +44,18 @@ The same Codespace environment can be used to run FlightCheck without the full m
 
 ## FlightCheck-Only Mode
 
-For users who want to run a pre-deployment readiness check on their Power Platform environment — no coding tools or VS Code required. Just run this command in PowerShell:
+For users who want to run a pre-deployment readiness check on their Power Platform environment — no coding tools or VS Code required. Just run this command:
+
+**Windows** (PowerShell):
 
 ```powershell
 iex (irm https://raw.githubusercontent.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/main/setup/bootstrap-flightcheck.ps1)
+```
+
+**macOS** (Terminal):
+
+```bash
+/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/main/setup/bootstrap-flightcheck-mac.sh)"
 ```
 
 That's it. The script handles everything automatically:
@@ -71,10 +87,13 @@ python scripts/flightcheck/cli.py --scope full
 
 | File | Purpose |
 |---|---|
-| `ess-adk-setup.winget.yaml` | Declarative DSC config consumed by `winget configure`. Installs VS Code, Python 3.12, PowerShell 7, Git, GitHub CLI. |
-| `Install-EssAdk.ps1` | Orchestrator. Installs toolchain via winget, installs pip dependencies, clones the repo, installs VS Code extensions, launches VS Code. Idempotent. With `-FlightCheckOnly`, installs minimal toolchain and runs interactive environment/agent discovery. |
-| `bootstrap.ps1` | One-liner entry point for the full maker kit install. Downloads the installer into `$env:TEMP` and runs it. |
-| `bootstrap-flightcheck.ps1` | One-liner entry point for FlightCheck-only install. Downloads the installer and runs it with `-FlightCheckOnly`. |
+| `Install-EssAdk.ps1` | Windows orchestrator. Installs toolchain via winget, pip dependencies, clones repo, installs extensions, launches VS Code. With `-FlightCheckOnly`, installs minimal toolchain and runs FlightCheck. |
+| `install-ess-adk.sh` | macOS orchestrator. Same as above but uses Homebrew. Set `FLIGHTCHECK_ONLY=true` for FlightCheck-only mode. |
+| `bootstrap.ps1` | Windows one-liner entry point (full maker kit). |
+| `bootstrap-flightcheck.ps1` | Windows one-liner entry point (FlightCheck only). |
+| `bootstrap-mac.sh` | macOS one-liner entry point (full maker kit). |
+| `bootstrap-flightcheck-mac.sh` | macOS one-liner entry point (FlightCheck only). |
+| `ess-adk-setup.winget.yaml` | Declarative DSC config consumed by `winget configure` (optional Windows path). |
 | `.devcontainer/devcontainer.json` | Codespace configuration. Pre-installs Python 3.12, pip dependencies, and Copilot extensions. |
 
 ## Dependencies
--- a/setup/README.md
+++ b/setup/README.md
@@ -78,9 +78,16 @@ iex (irm https://raw.githubusercontent.com/microsoft/Employee-Self-Service-Agent
 
 Once you've run the installer once, you can re-run FlightCheck directly without going through setup again:
 
+**Windows:**
 ```powershell
 cd $env:USERPROFILE\source\Employee-Self-Service-Agent-Developer-Kit\solutions\ess-maker-skills
 python scripts/flightcheck/cli.py --scope full
+```
+
+**macOS:**
+```bash
+cd ~/source/Employee-Self-Service-Agent-Developer-Kit/solutions/ess-maker-skills
+.venv/bin/python scripts/flightcheck/cli.py --scope full
 ```
 
 ## Files
@@ -106,7 +113,7 @@ The installer provisions the following dependencies. Users do not need to instal
 |------|---------|---------|:-----------------:|
 | Python | 3.12 | Runtime for FlightCheck and maker scripts | ✅ |
 | Git | Latest | Clone the repo, version control | ✅ |
-| GitHub CLI (`gh`) | Latest | Device-code auth flow for private repo clone | ✅ |
+| GitHub CLI (`gh`) | Latest | Device-code auth flow for private repo clone | ❌ |
 | VS Code | Latest | Editor and Copilot host | ❌ |
 | PowerShell 7 | Latest | Script execution (Windows only) | ❌ |
 
@@ -138,6 +145,8 @@ The devcontainer provides an equivalent pre-built environment:
 
 ## How to test it locally
 
+### Windows
+
 From this folder:
 
 ```powershell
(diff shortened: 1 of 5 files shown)


setup/install-ess-adk.sh
```diff
@@ -0,0 +1,263 @@
...
+BREW_PREFIX="$(brew --prefix)"
+
+# Try brew Python 3.12 first
+if [[ -x "$BREW_PREFIX/bin/python3.12" ]]; then
+    PYTHON="$BREW_PREFIX/bin/python3.12"
+elif command -v python3.12 &>/dev/null; then
+    PYTHON="$(command -v python3.12)"
+elif command -v python3 &>/dev/null; then
+    PY_VER="$(python3 --version 2>&1 | grep -oE '[0-9]+\.[0-9]+')"
+    if [[ "$PY_VER" == "3.12" || "$PY_VER" > "3.12" ]]; then
```
nehaoss: Bug: Bash lexicographic string comparison (`>`) doesn't work correctly for version numbers. `"3.9" > "3.12"` evaluates to **true** because '9' > '1' character-by-character, so this would incorrectly accept Python 3.9 as >= 3.12.

**Fix:** Use `sort -V` or split into major/minor integers:
```bash
if printf '%s\n' "3.12" "$PY_VER" | sort -V | head -n1 | grep -q "3.12"; then
```

setup/install-ess-adk.sh
```diff
@@ -0,0 +1,263 @@
...
+    step "Running FlightCheck"
+
+    MAKER_KIT_PATH="$REPO_PATH/solutions/ess-maker-skills"
+    cd "$MAKER_KIT_PATH"
+    FLIGHTCHECK_PYTHON="$VENV_PATH/bin/python"
+    if [[ ! -x "$FLIGHTCHECK_PYTHON" ]]; then
+        FLIGHTCHECK_PYTHON="$PYTHON"
+    fi
+    "$FLIGHTCHECK_PYTHON" scripts/flightcheck/cli.py --scope full
+    exit $?
```
nehaoss, line 252: The FlightCheck-only path does `cd $MAKER_KIT_PATH` without checking if the directory exists first. If the clone failed silently (e.g., the `|| warn` on line 170 swallowed an error), this will fail with a confusing 'No such file or directory' error.

**Fix:** Add a guard:
```bash
if [[ ! -d "$MAKER_KIT_PATH" ]]; then
    err "Maker kit path not found at $MAKER_KIT_PATH. Was the clone successful?"
    exit 1
fi
```

setup/install-ess-adk.sh
```diff
@@ -0,0 +1,263 @@
...
+    fi
+
+    if [[ -n "$CODE_CMD" ]]; then
+        # Required extensions — fail if these can't be installed
+        REQUIRED_EXTENSIONS=("GitHub.copilot" "GitHub.copilot-chat")
+        for ext in "${REQUIRED_EXTENSIONS[@]}"; do
+            if ! "$CODE_CMD" --install-extension "$ext" --force 2>/dev/null; then
+                err "Failed to install required extension: $ext"
+                err "Sign in to VS Code with a GitHub account that has Copilot access, then re-run this script."
+                exit 1
```
nehaoss, line 220: The error message tells the user to 'Sign in to VS Code with a GitHub account that has Copilot access, then re-run this script' -- but this is a CLI install (`code --install-extension`). The most common failure cause is that the extension marketplace is unreachable (corporate proxy/firewall). The message should mention marketplace connectivity as a possible cause too, since the user hasn't opened VS Code yet at this point in the flow.

setup/bootstrap-mac.sh
```diff
@@ -0,0 +1,27 @@
...
+
+if ! curl -fsSL "$INSTALLER_URL" -o "$TEMP_DIR/install-ess-adk.sh"; then
+    echo "  [ERR] Failed to download: $INSTALLER_URL" >&2
+    echo "  If raw.githubusercontent.com is blocked by your firewall/proxy," >&2
+    echo "  clone the repo manually and run: setup/install-ess-adk.sh" >&2
+    exit 1
+fi
+
+# Run in-memory (source) to avoid any permission issues
+source "$TEMP_DIR/install-ess-adk.sh"
```
nehaoss, line 34: Security concern: the downloaded script is `source`'d (executed) without any integrity check. If the network delivers a corrupted or tampered file (e.g., corporate proxy injecting an HTML error page), it runs as arbitrary code.

**Fix:** At minimum, check the file is non-empty and starts with a shebang before sourcing:
```bash
if [[ ! -s "$TEMP_DIR/install-ess-adk.sh" ]] || ! head -1 "$TEMP_DIR/install-ess-adk.sh" | grep -q '^#!/'; then
    echo "[ERR] Downloaded file appears invalid" >&2
    exit 1
fi
```

setup/install-ess-adk.sh
```diff
@@ -0,0 +1,263 @@
...
+#   Homebrew, Python 3.12, Git, GitHub CLI, VS Code, Copilot extensions,
+#   pip dependencies, clones the repo, and launches VS Code.
+#
+# Usage (full maker kit):
+#   /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/main/setup/bootstrap-mac.sh)"
+#
+# Usage (FlightCheck only):
+#   /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/main/setup/bootstrap-flightcheck-mac.sh)"
+# ---------------------------------------------------------------------------
+set -euo pipefail
```
nehaoss, line 15: The script uses `set -euo pipefail` but `CODE_CMD` is only set inside the `FLIGHTCHECK_ONLY != true` block (line 201). If the control flow around the FlightCheck early-exit (line 243) ever changes, referencing `$CODE_CMD` at line 249 would crash under `set -u`. Consider initializing `CODE_CMD=""` near the top of the script (alongside the other config vars) to make this robust against future refactoring.

setup/install-ess-adk.sh
qazizaahirah: +1 on this might turn out to be a bug

setup/README.md
```diff
@@ -62,19 +78,29 @@ iex (irm https://raw.githubusercontent.com/microsoft/Employee-Self-Service-Agent
...
+**Windows:**
 ```powershell
 cd $env:USERPROFILE\source\Employee-Self-Service-Agent-Developer-Kit\solutions\ess-maker-skills
 python scripts/flightcheck/cli.py --scope full
 ```
 
+**macOS:**
+```bash
+cd ~/source/Employee-Self-Service-Agent-Developer-Kit/solutions/ess-maker-skills
+.venv/bin/python scripts/flightcheck/cli.py --scope full
```
GrahamMcMynn: The documented macOS re-run command points at a .venv path that won't exist after running the installer.

install-ess-adk.sh creates the venv at the repo root (`VENV_PATH="$REPO_PATH/.venv"`, i.e. `~/source/Employee-Self-Service-Agent-Developer-Kit/.venv`), but this snippet `cd`s into `solutions/ess-maker-skills/` and then runs `.venv/bin/python`, which resolves to `~/source/Employee-Self-Service-Agent-Developer-Kit/solutions/ess-maker-skills/.venv/bin/python` — that path doesn't exist, so the command fails with `No such file or directory`.

**Fix** — either run from the repo root:
```bash
cd ~/source/Employee-Self-Service-Agent-Developer-Kit
.venv/bin/python solutions/ess-maker-skills/scripts/flightcheck/cli.py --scope full
```
or reference the venv via a relative path:
```bash
cd ~/source/Employee-Self-Service-Agent-Developer-Kit/solutions/ess-maker-skills
../../.venv/bin/python scripts/flightcheck/cli.py --scope full
```

GrahamMcMynn (commented):

setup/install-ess-adk.sh
```diff
@@ -0,0 +1,263 @@
...
+        fi
+    fi
+}
+
+install_brew_cask() {
+    local cask="$1"
+    local name="${2:-$1}"
+    if brew list --cask "$cask" &>/dev/null; then
+        ok "$name (already installed)"
+    elif [[ -d "/Applications/Visual Studio Code.app" ]]; then
```
GrahamMcMynn: `install_brew_cask` looks generic — it accepts a `$cask` parameter — but it hardcodes `/Applications/Visual Studio Code.app` in two places (this line and again in the post-install check on line 103):

```bash
elif [[ -d "/Applications/Visual Studio Code.app" ]]; then
    ok "$name (already installed outside Homebrew)"
...
if brew list --cask "$cask" &>/dev/null || [[ -d "/Applications/Visual Studio Code.app" ]]; then
```

Today it's only called with `visual-studio-code` so it works. But a future caller (`install_brew_cask "iterm2"` etc.) would silently report a successful install whenever VS Code happens to be present, and would incorrectly fall through to the `err`/`exit 1` branch if the new cask installs fine but VS Code isn't around.

**Fix** — either drop the VS-Code-specific bundle check (rely solely on `brew list --cask`), or accept the app-bundle path as a third parameter:
```bash
install_brew_cask() {
    local cask="$1" name="${2:-$1}" app_path="${3:-}"
    ...
    elif [[ -n "$app_path" && -d "$app_path" ]]; then
        ok "$name (already installed outside Homebrew)"
    ...
}
# caller:
install_brew_cask "visual-studio-code" "Visual Studio Code" "/Applications/Visual Studio Code.app"
```

GrahamMcMynn (commented):

setup/install-ess-adk.sh
amilandi: Fixed — switched to `sort -V` for proper semantic version comparison.

setup/install-ess-adk.sh
amilandi: Fixed — added a guard that checks if the directory exists before `cd`.

setup/install-ess-adk.sh
amilandi: Fixed — error message now mentions marketplace connectivity (proxy/firewall) as a possible cause.

setup/bootstrap-mac.sh
amilandi: Fixed — both bootstrap scripts now verify the downloaded file is non-empty and starts with a shebang before sourcing.

setup/install-ess-adk.sh
amilandi: Fixed — `CODE_CMD` is now initialized to empty string at the top of the script alongside other config vars.

setup/README.md
amilandi: Fixed — changed to `../../.venv/bin/python` since the venv lives at repo root.

setup/install-ess-adk.sh
amilandi: Fixed — `install_brew_cask` now accepts an optional third `app_path` parameter. The VS Code call passes `/Applications/Visual Studio Code.app` explicitly.

GrahamMcMynn (approved):

PR #127, opened 2026-06-05
FlightCheck: Check Workday SOAP/REST integration certificate validity…

### LINKED ISSUE #84: FlightCheck: Add check for SAML NameID alignment with Workday user identifier

apurvabanka:
Checkpoint: WD-CONN-102
Category: Workday
Priority: High
Status emitted: MANUAL on the healthy path (operator compares the Entra thumbprint against the matching Workday "Edit Tenant Setup - Security → SAML Identity Providers" row). Auto-escalates to FAILED (no AsymmetricX509Cert keyCredentials / all expired / active selection expired with rollover live), WARNING (active cert within CERT_EXPIRY_WARN_DAYS=30 of NotAfter, or NotBefore in the future, or 401/403 consent), NOT_CONFIGURED (no federated Workday SAML enterprise app), SKIPPED (Graph unavailable).
Conditional? No — runs before the no-Workday early-return gate in run_workday_checks so the warning fires pre-install.

Why MANUAL is the right shape on the healthy path

ESS Workday connectors use Basic auth + Entra SSO; there is no certificate stored on the Power Platform connection record. The cert the spec's failure mode describes is the SAML signing cert on the federated Workday enterprise app's servicePrincipal (reachable via Microsoft Graph). The matching cert uploaded to Workday's tenant security setup has no Workday API surface (SOAP RaaS / Worker services don't expose tenant security config; WQL admin is gated by the documented chicken-and-egg blocker). AGENTS.md design principle #2 introduces MANUAL specifically for this case — canonical reference AUTH-006, commit 53f3762, fixes #84. WD-CONN-102 reuses the pattern with the addition that Entra-side data is rich enough to programmatically classify expiry/missing/all-expired states into FAILED/WARNING directly.

API tier

Microsoft Graph v1.0 — validatable tier per tests/fixtures/cassettes/INDEX.md. No cassette / no registry changes. Extends existing tests/mocks/graph.py service_principal() with key_credentials and preferred_token_signing_key_thumbprint kwargs, adds a key_credential() builder for the keyCredential complex type — both CSDL-cited. Docstrings updated to list WD-CONN-102 as a consumer.

What the check observes vs. delegates

Entra side (observed via Microsoft Graph):

 - Every federated Workday enterprise app — same SAML filter as AUTH-006: startswith(displayName,'Workday') and preferredSingleSignOnMode eq 'saml'.
 - keyCredentials filtered to type == "AsymmetricX509Cert", with Sign + Verify entries sharing a customKeyIdentifier coalesced into one logical cert.
 - preferredTokenSigningKeyThumbprint to identify the active cert during rollover windows.
 - customKeyIdentifier decoded from base64 → 20-byte SHA-1 digest → colon-uppercase-hex (matches Workday's display format).
 - startDateTime / endDateTime parsed as UTC, days-to-expiry computed against datetime.now(timezone.utc).

Workday side (MANUAL operator):

 - Open "Edit Tenant Setup - Security" → "SAML Identity Providers", match the enabled IdP row's "Service Provider ID" against the entity IDs surfaced in the result to identify the active local Entra app.
 - View that row's X509 Certificate thumbprint, compare byte-for-byte against the active Entra thumbprint. Mismatch → end-user SAML SSO silently broken (ISU runtime calls still work, so the agent appears healthy).

Pattern decisions

 - $select forced on the listing call. Graph GET /servicePrincipals omits keyCredentials unless explicitly projected — without it, every healthy tenant would falsely report "no signing certificate." New GraphClient.get_workday_saml_service_principals() wraps the call with WORKDAY_SAML_SP_SELECT; pinned by TestSelectClauseInRequestUrl.
 - Sign+Verify coalescing. Single uploaded cert appears as two keyCredentials entries sharing customKeyIdentifier; _group_workday_cert_keycredentials() groups by CKI and prefers the Sign entry's metadata.
 - Active-cert selection during rollover. preferredTokenSigningKeyThumbprint is canonical; fallback to first cert whose [start, end] contains now; last-resort fallback to latest endDateTime.
 - Active expired with rollover live → FAILED, not WARNING. Operator forgot to flip the active selection after rotating; surfaced with explicit "rollover certificate exists" hint.
 - NotBefore in future → WARNING ("not yet valid for N more days").
 - Status bucketing for multi-SP tenants (AGENTS.md principle 7). At most one CheckResult per non-empty bucket; each lists only its bucket's SPs.
 - Permission probe pattern inherited from AUTH-006. Missing Application.Read.All surfaces as WARNING, never silent NOT_CONFIGURED.
 - Placement before the no-Workday early-return gate. Cert can be unhealthy pre-install; pinned by TestWireup.
 - Helper localization with lazy import. Cert helpers are local to checks/workday.py; _saml_entity_ids imported lazily from checks/authentication.py to avoid a circular import while reusing AUTH-006's tested filter.
 - Malformed customKeyIdentifier is not fatal. Renders as (malformed) rather than crashing.

Tests

tests/flightcheck/checks/test_workday_saml_certificate.py covers:

 - TestNotConfigured, TestPermissionDenied, TestNoGraphClientSkipped — defensive paths.
 - TestHealthyCertManual — MANUAL with thumbprint, NotAfter math, entity IDs, doc_link.
 - TestSignVerifyCoalescing — pair with same CKI collapses to one logical cert.
 - TestPreferredThumbprintSelectsActive — disambiguation during rollover, rollover: label on the others.
 - TestExpiringSoonWarning, TestNotYetValidWarning — WARNING with hardening framing (principle 9).
 - TestNoCertsFailed, TestAllExpiredFailed — FAILED with explicit reason in result text.
 - TestBucketing — 3 SPs in 3 statuses → 3 CheckResults, each listing only its bucket's SPs.
 - TestSelectClauseInRequestUrl — regression guard for the $select bug.
 - TestMalformedThumbprintIsNotFatal — malformed CKI doesn't crash the Workday block.
 - TestWireup — WD-CONN-102 fires pre-gate; downstream Workday checks do NOT fire (gate still works).

remediation-guide.md and validation-matrix.md updated. Local suite: N passed, M skipped, 0 regressions (fill in after pytest). ruff check clean on touched files.


apurvabanka:

solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py
```diff
@@ -111,6 +114,79 @@
...
+    "employee-self-service/workday#task-1-create-the-x509-public-key"
+)
+
+# Filter SAML-only — OIDC variants of the Workday enterprise app
+# don't use the SAML signing cert WD-CONN-102 inspects, so listing
+# them would emit confusing "no cert" diagnostics for apps that
+# don't need a cert. Used to pre-flight the /servicePrincipals probe
+# (the actual listing call uses graph.get_workday_saml_service_principals()
+# which carries the same filter plus the $select clause).
+_WORKDAY_SAML_SP_PROBE_FILTER = (
```
nehaoss: This filter string is duplicated -- the same expression exists inline in `graph_client.py`'s `get_workday_saml_service_principals()` method. Per the repo's 'No duplicate functions' copilot-instructions rule, define it once (e.g., as a constant on `GraphClient` or a shared module) and reference from both the probe call here and the listing method. If one drifts without the other, the probe could succeed while the listing filters differently.

solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py
```diff
@@ -1529,6 +1621,625 @@ def _format_unhealthy_remediation(entry: dict, env_id: str) -> str:
...
+            # rolling over. Surface as FAILED with rollover hint.
+            failed_entries.append({
+                "sp_name": sp_name,
+                "summary": (
+                    f"{sp_header}\n"
+                    f"    active: {cert_line}{rollover_block}\n"
+                    "    Note: a rollover certificate exists; the "
+                    "active selection has not been updated."
+                ),
+                "reason": "active_expired_rollover_exists",
```
nehaoss, line 2375: This `active_expired_rollover_exists` branch doesn't have a dedicated test. `TestPreferredThumbprintSelectsActive` uses a still-valid old cert so it never hits this path.

**Action:** Add a test where `preferredTokenSigningKeyThumbprint` points at an expired cert while a second non-expired cert exists. Assert FAILED status and the 'active selection has not been updated' note in result text.

solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py
```diff
@@ -1529,6 +1621,625 @@ def _format_unhealthy_remediation(entry: dict, env_id: str) -> str:
...
+            ),
+            doc_link=doc_link,
+        )]
+
+    # Probe /servicePrincipals with graph.get() so a missing
+    # Application.Read.All consent surfaces as a clear WARNING rather
+    # than silently masquerading as "no Workday SAML app exists"
+    # (which would falsely produce NOT_CONFIGURED). Same defensive
+    # pattern AUTH-006 uses — get_all() swallows 401/403 into [].
+    sp_probe = graph.get(
```
nehaoss: The probe call (`graph.get(..., top=1)`) checks permissions; on success its result is discarded and `get_workday_saml_service_principals()` re-fetches the same data. For tenants with many SPs this is two round-trips to Graph.

Consider catching 401/403 from the listing call directly, or document why the two-call pattern is intentional (e.g., because `get_all` pagination swallows HTTP errors into `[]`).

tests/flightcheck/checks/test_workday_saml_certificate.py
```diff
@@ -0,0 +1,603 @@
...
+        # runs (and gets SKIPPED because graph=None) even though the
+        # downstream Workday block early-returns.
+        with patch(
+            "flightcheck.checks.workday._check_package_flavor",
+            return_value=[],
+        ):
+            results = run_workday_checks(bare)
+
+        ids = [r.checkpoint_id for r in results]
+        assert "WD-CONN-102" in ids
```
nehaoss, line 664: The docstring says 'downstream Workday checks do NOT fire (gate still works)' but the test only asserts `WD-CONN-102 in ids`. If the early-return gate regresses this won't catch it.

**Action:** Add `assert not any(i.startswith('WD-FLOW-') for i in ids)` to pin the gate behavior.

tests/flightcheck/checks/test_workday_saml_certificate.py
```diff
@@ -0,0 +1,603 @@
...
+
+    def test_wired_in_before_early_return_guard(self) -> None:
+        from flightcheck.checks.workday import run_workday_checks
+
+        @dataclass
+        class _Bare:
+            graph: Any = None
+            pp_admin: Any = None
+            env_id: str | None = None
+            _workday_flows: list = None
```
nehaoss: Nit: `_workday_flows: list = None` should be `list | None = None` (same for `config: dict`). Minor, but avoids type-checker noise.

apurvabanka: added

solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py
apurvabanka: added

solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py
apurvabanka: removed the duplicate.

solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py
apurvabanka: done

solutions/ess-maker-skills/scripts/flightcheck/graph_client.py
```diff
@@ -221,10 +221,22 @@ def get_service_principals(
...
         raise_on_permission_error: bool = False,
     ) -> list:
         """List service principals (enterprise apps) with optional filter.
 
+        ``select`` controls the OData ``$select`` projection. Graph
+        ``GET /servicePrincipals`` listing responses OMIT several
+        navigation-adjacent properties unless explicitly requested,
+        most relevantly ``keyCredentials`` and
+        ``preferredTokenSigningKeyThumbprint`` — both used by
+        WD-CONN-102 (SAML signing certificate health). Without an
```
srideshpande, line 234: Q: In the context of simplified Workday setup, where we are using end user's OAuth credentials to connect to WD, I would like to know which ESS code path actually breaks when this cert expires?

solutions/ess-maker-skills/src/reference/ess-docs/flightcheck/remediation-guide.md
```diff
@@ -106,6 +106,33 @@ after installing the Workday extension pack.
 5. Or run `/connect workday` for guided re-setup
 **Verify:** Re-run `/flightcheck --scope workday`
 
+### WD-CONN-102: Workday SAML signing certificate health
```
srideshpande: Q: Should we mention that in here that,  this check is about the Workday SAML / Entra federation setup, not the normal ESS runtime API call path. Otherwise, a failure here may be misread as “ESS can’t call Workday,” which sends customers toward the wrong root cause.

srideshpande (commented):
The PR title/description feel a bit confusing to me. Are we actually checking a certificate used by the Workday SOAP/REST integration itself, or are we checking the SAML signing certificate on the federated Workday Enterprise App in Entra?

From the implementation, it seems like the latter. If so, could we clarify the title/description to say “Workday SAML / Enterprise App certificate health” or something in those lines, rather than “SOAP/REST integration certificate” so the scope is clearer?

solutions/ess-maker-skills/scripts/flightcheck/graph_client.py
apurvabanka: As far as I know, the ADK flightcheck didn't break. The failure was at the ESS agent for all Workday topics. That's why we are adding a manual check. Is that what you are asking?

solutions/ess-maker-skills/src/reference/ess-docs/flightcheck/remediation-guide.md
apurvabanka: hey Srikanth, this makes sense. I have changed the documentation to point to Workday SAML / Entra federation setup. Let me know if there is anything else I should add.

solutions/ess-maker-skills/scripts/flightcheck/graph_client.py
srideshpande: Thanks. I agree the check is useful. I’m just trying to make sure we describe it clearly so others can understand why it exists later.

My main question is: this check validates the SAML cert on the Workday federated enterprise app in Entra. But how does that cert impact the actual ESS runtime flow for Workday topics?

For example, when this cert expires, what is the first thing that breaks?

- User sign-in to Workday
- Token acquisition for the user-context connection
- The actual SOAP/REST call to Workday

If we can identify the first failing step, then the PR title and description become much clearer. Something like:

`FlightCheck: Validate Workday SAML certificate for user-context auth` OR
`FlightCheck: Validate Workday SAML certificate used by ESS Workday topics` OR

simply -
`FlightCheck: Check Workday SAML certificate validity`

I feel the word integration is incorrect in the title. Anyways, I hope you understand what I am trying to get at, with all possible places the cert can affect.

srideshpande (approved):

PR #128, opened 2026-06-05
feat(flightcheck): add ENV-009 preferred customization solution check

daeunJe0ng:
## Description

Adds an environment-scope FlightCheck (`ENV-009`) that validates whether the current FlightCheck caller has selected a customer-owned unmanaged solution — bound to a customer publisher — as their preferred customization solution in Dataverse. Without this, new Copilot Studio / Maker portal customizations land in the Default Solution (or get bound to the env's auto-generated Default Publisher, prefix `new`), neither of which is exportable as a clean ALM artifact.

Note: the preferred-solution setting is stored per-user (on the `usersettings` table behind the scenes), not per-environment, so the check explicitly scopes its findings to the current caller's identity.

**Verdicts (always exactly one CheckResult emitted):**

- **SKIPPED**: `env_url` or `dv_token` missing.
- **FAILED**: no eligible customer-created unmanaged solutions exist in the environment.
- **WARNING**: eligible solution(s) exist but
  - the caller has not selected any of them as their preferred solution, OR
  - the caller's preferred solution exists but is not in the eligible set, OR
  - the caller's preferred solution is bound to the environment's **Default Publisher** (auto-generated, prefix `new`) — surfaces publisher display name and prefix so the recommended remediation is unambiguous, OR
  - the Dataverse call returned 401 (token expired), 403 (insufficient privilege), or 5xx (transient). HTTP status code is surfaced when available so 403 is distinguishable from 5xx.
- **PASSED**: the caller's preferred solution matches one of the eligible solutions AND is bound to a customer (non-default) publisher. Publisher-lookup failures degrade gracefully (PASSED still returned when solution-eligibility is satisfied) so a transient publisher fetch does not regress a known-good env.

**Implementation:**

- New helper `auth.dataverse_get()` for non-paginated Dataverse Web API GETs (function calls, single-record reads); reuses `_SESSION`, `HEADERS_BASE`, `_validate_https_url`; raises `AuthExpiredError` on 401. Two `assert` guards reject paths that already include a leading `/` or the `api/data/` prefix (case-insensitive) to catch developer mistakes that `.lstrip('/')` would otherwise mask.
- `_check_preferred_solution()` short-circuits when no eligible solutions exist, then uses a **single** `GetPreferredSolution()` call (bound to the caller via the bearer token — no separate `WhoAmI()` lookup or `usersettingscollection({UserId})` round-trip needed). When a preferred solution is returned, the check additionally resolves the publisher via `$expand=publisherid($select=friendlyname,customizationprefix)` and flags `prefix='new'` (the auto-generated Default Publisher signature) as a WARNING.
- GUID equality between the `solutions` collection and `GetPreferredSolution()` is normalised through `uuid.UUID(...)` so case/format differences between the two endpoints can never cause a false WARNING.

### UNCERTAINTY

The MS Learn reference for [`GetPreferredSolution`](https://learn.microsoft.com/power-apps/developer/data-platform/webapi/reference/getpreferredsolution) documents the return type as `crmbaseentity` but **does not include an example response body**. The code treats the response defensively: a body containing `solutionid` is the selected solution; a body without it (or any decode-edge case) is reported via the catch-all WARNING with the HTTP status code surfaced. The mock builder in `tests/mocks/dataverse.py:get_preferred_solution` composes the response shape from three MS-documented sources (.NET SDK `GetPreferredSolutionResponse.PreferredSolution` + `solution` entity reference + Web API single-entity serialisation convention). Live-tenant smoke test has now been run on a non-prod env against every reachable branch.

## Related issue

N/A

## Type of change

### Lint + syntax (CONTRIBUTING.md §1)

Both CI commands run clean from repo root:

```pwsh
ruff check solutions/ess-maker-skills/scripts/ solutions/ess-maker-skills/src/mcp/
python -m compileall -q solutions/ess-maker-skills/scripts/ solutions/ess-maker-skills/src/mcp/
```

### Unit tests

**15** unit tests in `tests/flightcheck/checks/test_preferred_solution.py` covering every verdict path (SKIPPED, FAILED, WARNING sub-cases including 401 / 403 / 500 / no-preferred / not-in-eligible / Default-Publisher, PASSED including the cross-endpoint case-insensitive-GUID path and the graceful-degradation publisher-fetch path) with mocked Dataverse responses. All mocks built via the documented-tier builders in `tests/mocks/dataverse.py` per `tests/AGENTS.md`.

```
pytest tests/flightcheck/checks/test_preferred_solution.py -> 15 passed
pytest tests/ --ignore=tests/captures               -> 316 passed, 8 skipped
```

### End-to-end manual validation

The affected command is `/flightcheck`. The matrix below is the minimum set that exercises every distinct branch in `_check_preferred_solution()`, verified against a non-prod env.

| Test | What this test covers | How to set up | Expected verdict | Matches actual result? |
|------|-----------------------|---------------|------------------|------------------------|
| **T1** | No customer-created unmanaged solutions exist | Use a fresh env before creating any solutions | **FAILED** | [x] verified on a fresh non-prod env before any customer solutions were created |
| **T2** | Caller's preferred solution is a valid customer solution bound to a customer publisher | Create a customer publisher (any prefix other than `new`), create one customer unmanaged solution with that publisher, then set it as preferred | **PASSED** | [x] verified on a non-prod env |
| **T3** | Customer solution exists but caller has not chosen any preferred solution (`GetPreferredSolution()` returns no `solutionid`) | Create one customer unmanaged solution, do **not** tick "set as preferred" (or clear it via API) | **WARNING** "no preferred solution selected" | [x] verified via unit test `test_warning_when_no_preferred_solution_selected` (`tests/flightcheck/checks/test_preferred_solution.py`) — live tenant repro skipped because maker portal does not expose a UI affordance to leave preferred unset once a solution has been created in that env |
| **T4** | Caller's preferred solution is the Default Solution itself (not a customer solution) | Create one customer unmanaged solution, PATCH `usersettings` to point `preferredsolution_userid@odata.bind` at the env's Default solution id | **WARNING** "preferred not in eligible set" | [x] verified on a non-prod env |
| **T5** | Caller lacks privilege to read `solution` / call `GetPreferredSolution()` | Run `/flightcheck` as a user with no `prvReadSolution` | **WARNING** with `[HTTP 403]` surfaced in the result text | [x] verified via unit test `test_warning_when_get_preferred_solution_returns_403` (`tests/flightcheck/checks/test_preferred_solution.py`) — live tenant repro skipped because provisioning a restricted user requires Global Admin / User Admin rights not held by the PR author |
| **T6** | Dataverse token has expired mid-run | Wait until the cached token in `.local/.token_cache.bin` expires, or manually corrupt it, then run `/flightcheck` | **WARNING** "auth expired" | [x] verified via unit test `test_warning_when_dataverse_returns_401` (`tests/flightcheck/checks/test_preferred_solution.py`) — live tenant repro skipped because VS Code / MSAL auto-refreshes the token transparently before the FlightCheck call lands |

## Checklist

- [X] I have updated documentation as needed
- [x] No fabricated URLs (CONTRIBUTING.md §5): only `https://make.powerapps.com` is referenced in the eligible-solutions hardening message; it is the canonical Power Apps maker portal entry point
- [x] Minimal, surgical changes (CONTRIBUTING.md §6): scope limited to one new check + one new helper + tests; no unrelated edits

## Follow-up (out of scope of this PR)

- Tracking issue to be filed: collapse the five sibling helpers in `auth.py` (`query_all`, `update_record`, `create_record`, `delete_record`, `dataverse_get`) into a shared private `_dv_request(env_url, token, method, path, ...)` to remove the duplicated preamble (per @qazi-microsoft review feedback).

N/A — this PR does not touch `samples/`.

### DIFF
--- a/solutions/ess-maker-skills/scripts/auth.py
+++ b/solutions/ess-maker-skills/scripts/auth.py
@@ -250,6 +250,19 @@ def dataverse_get(env_url, token, path, params=None):
         For other non-2xx responses (raised via ``raise_for_status``).
     """
     _validate_https_url(env_url)
+    # Catch developer mistakes where the absolute base path is passed in.
+    # `.lstrip('/')` below would silently strip a leading slash; without
+    # these asserts, `'/api/data/v9.2/WhoAmI()'` becomes a double-prefixed
+    # URL and `'api/data/v9.2/WhoAmI()'` becomes a malformed one. The doc
+    # says "path relative to /api/data/v9.2/" — enforce it.
+    assert not path.startswith("/"), (
+        f"dataverse_get path must be relative to /api/data/v9.2/ "
+        f"(no leading slash), got: {path!r}"
+    )
+    assert not path.lower().startswith("api/data/"), (
+        f"dataverse_get path must be relative to /api/data/v9.2/ "
+        f"(do not include it), got: {path!r}"
+    )
     headers = {**HEADERS_BASE, "Authorization": f"Bearer {token}"}
     url = f"{env_url}/api/data/v9.2/{path.lstrip('/')}"
     resp = _SESSION.get(url, headers=headers, params=params, timeout=60, verify=True)
--- a/solutions/ess-maker-skills/scripts/auth.py
+++ b/solutions/ess-maker-skills/scripts/auth.py
@@ -217,6 +217,48 @@ def query_all(env_url, token, entity_set, select, filter_expr=None):
     return all_records
 
 
+def dataverse_get(env_url, token, path, params=None):
+    """GET a Dataverse Web API endpoint that is not a paged table query.
+
+    Use this for function calls (e.g. ``WhoAmI()``) and single-record reads
+    (e.g. ``usersettingscollection({systemuserid})``) where ``query_all`` does
+    not fit — ``query_all`` always appends ``$select`` and follows
+    ``@odata.nextLink``.
+
+    Parameters
+    ----------
+    env_url : str
+        Base environment URL (e.g. ``https://contoso.crm.dynamics.com``).
+    token : str
+        Dataverse bearer token.
+    path : str
+        Path relative to ``/api/data/v9.2/`` (no leading slash). Example:
+        ``"WhoAmI()"`` or ``"usersettingscollection(11111111-...)``.
+    params : dict | None
+        Optional querystring parameters (e.g. ``{"$select": "_preferredsolution_value"}``).
+
+    Returns
+    -------
+    dict
+        Parsed JSON response body.
+
+    Raises
+    ------
+    AuthExpiredError
+        If the response is 401.
+    requests.HTTPError
+        For other non-2xx responses (raised via ``raise_for_status``).
+    """
+    _validate_https_url(env_url)
+    headers = {**HEADERS_BASE, "Authorization": f"Bearer {token}"}
+    url = f"{env_url}/api/data/v9.2/{path.lstrip('/')}"
+    resp = _SESSION.get(url, headers=headers, params=params, timeout=60, verify=True)
+    if resp.status_code == 401:
+        raise AuthExpiredError("Dataverse returned 401 (token expired or invalid)")
+    resp.raise_for_status()
+    return resp.json()
+
+
 def update_record(env_url, token, entity_set, record_id, data):
     """Update a single Dataverse record via PATCH.
 
--- a/solutions/ess-maker-skills/scripts/flightcheck/checks/environment.py
+++ b/solutions/ess-maker-skills/scripts/flightcheck/checks/environment.py
@@ -400,7 +400,14 @@ def _check_connections_and_refs(runner) -> list[CheckResult]:
 #                one of them.
 # ---------------------------------------------------------------------------
 
-_PREFSOL_DOC_LINK = f"{DOC_BASE}/install#set-up-a-preferred-solution"
+# Pinned to the current canonical MS Learn path. The module-level DOC_BASE
+# points at /copilot/microsoft-365/employee-self-service which 301-redirects
+# to /microsoft-365/copilot/employee-self-service; bypass the redirect here
+# so this link stays stable without rewriting DOC_BASE (out of scope for this PR).
+_PREFSOL_DOC_LINK = (
+    "https://learn.microsoft.com/en-us/microsoft-365/copilot/"
+    "employee-self-service/install#set-up-a-preferred-solution"
+)
 _PREFSOL_DESCRIPTION = "Maker has preferred customization solution selected"
 
 # OData filter that excludes Microsoft-installed and system solutions, leaving
--- a/solutions/ess-maker-skills/scripts/flightcheck/checks/environment.py
+++ b/solutions/ess-maker-skills/scripts/flightcheck/checks/environment.py
@@ -7,8 +7,6 @@ ESS FlightCheck — Environment Configuration Validation (ENV-xxx)
 Checks Power Platform environment, Dataverse, DLP policies, and related config.
 """
 
-import requests
-
 from ..runner import CheckResult, Status, Priority
 from .connections import get_connection_status
 from auth import query_all, dataverse_get, AuthExpiredError  # scripts/auth.py, on path via cli.py
@@ -365,8 +363,8 @@ def _check_connections_and_refs(runner) -> list[CheckResult]:
 # lines 19-45) tells operators to create a customer-owned unmanaged solution and
 # select it as their *preferred solution* so that new Copilot Studio / Maker
 # portal customizations land in that solution instead of in the Default
-# Solution. The preferred-solution selection is stored per user on the
-# `usersettings` table, not per environment, so this check validates the
+# Solution. The preferred-solution selection is stored per user (on the
+# `usersettings` table), not per environment, so this check validates the
 # *current FlightCheck caller's* selection - it is honest about that scope in
 # the description and result text.
 #
@@ -375,12 +373,19 @@ def _check_connections_and_refs(runner) -> list[CheckResult]:
 #     GET /solutions?$filter=ismanaged eq false and isvisible eq true
 #         and uniquename ne 'Default' and uniquename ne 'Active'
 #         and solutiontype eq 0 and _parentsolutionid_value eq null
-#  2. Caller identity
-#     GET /WhoAmI()
-#     -> https://learn.microsoft.com/power-apps/developer/data-platform/webapi/use-web-api-functions
-#  3. Caller's preferred solution
-#     GET /usersettingscollection({UserId})?$select=_preferredsolution_value
-#     -> https://learn.microsoft.com/power-apps/developer/data-platform/reference/entities/usersettings
+#  2. Caller's preferred solution (single round-trip, bound to the caller
+#     via the bearer token - no separate WhoAmI() lookup needed)
+#     GET /GetPreferredSolution()
+#     -> https://learn.microsoft.com/power-apps/developer/data-platform/webapi/reference/getpreferredsolution
+#
+# UNCERTAINTY: the MS Learn reference for GetPreferredSolution() documents
+# the return type as `crmbaseentity` but does not include an example
+# response body. The code below treats the response defensively:
+#  * A body that contains `solutionid` is the selected solution; we then
+#    compare it against the eligible set above.
+#  * A body that omits `solutionid` (or any decode/empty-body edge case
+#    that bubbles up as an exception) is reported via the generic catch-all
+#    WARNING with the HTTP status code surfaced (per PR #128 review).
 #
 # Verdict map (always exactly one CheckResult emitted):
 #   * SKIPPED  - env_url or dv_token missing.
@@ -432,7 +437,7 @@ def _check_preferred_solution(runner) -> list[CheckResult]:
         )]
 
     try:
-        # Eligible unmanaged customer solutions.
+        # Signal 1: eligible unmanaged customer solutions.
         solutions = query_all(
             env_url, token,
             "solutions",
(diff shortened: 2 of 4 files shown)


nehaoss (commented):
~Superseded by inline review comments below.~

solutions/ess-maker-skills/scripts/flightcheck/checks/environment.py
```diff
@@ -7,9 +7,11 @@
 Checks Power Platform environment, Dataverse, DLP policies, and related config.
 """
 
+import requests
```
nehaoss: The `requests` module is imported here solely for `requests.HTTPError` in the 404 catch (~line 468). Since `auth.py` already depends on `requests` and re-raises `HTTPError`, consider re-exporting `HTTPError` from `auth` (e.g. `from auth import ..., HTTPError`) to avoid coupling this module directly to the `requests` library. This keeps the dependency boundary clean — if `auth.py` ever switches HTTP libs, only one file needs updating.

solutions/ess-maker-skills/scripts/flightcheck/checks/environment.py
```diff
@@ -130,6 +132,9 @@ def run_environment_checks(runner) -> list[CheckResult]:
             remediation="Requires Power Platform Admin role.",
         ))
 
+    # ---- ENV-009: Maker has preferred customization solution selected ----
```
nehaoss, line 345: Inside `_check_preferred_solution` (called here), `whoami.get('UserId')` relies on exact PascalCase from the Dataverse response. If the API ever returns different casing, this silently returns `None` and falls into a WARNING branch with a confusing message.

Consider a case-insensitive lookup or at minimum logging the raw keys in the fallback WARNING result:
```python
user_id = whoami.get('UserId') or whoami.get('userid')
```

solutions/ess-maker-skills/scripts/flightcheck/checks/environment.py
```diff
@@ -351,3 +356,228 @@ def _check_connections_and_refs(runner) -> list[CheckResult]:
...
+                ),
+                doc_link=_PREFSOL_DOC_LINK,
+            )]
+
+        # Signal 2 + 3: caller identity, then caller's preferred solution lookup.
+        whoami = dataverse_get(env_url, token, "WhoAmI()")
+        user_id = whoami.get("UserId")
+        if not user_id:
+            return [CheckResult(
+                checkpoint_id="ENV-009", category="Environment",
```
nehaoss: The inner `except requests.HTTPError` only handles 404 and re-raises everything else. Those re-raised errors get caught by the outer `except Exception` — producing a generic 'Unable to validate preferred solution' message.

A **403** (insufficient privileges on the `usersettings` table) is a common operational scenario that deserves a specific, actionable remediation. Consider either:
1. Catching non-404 `HTTPError` explicitly here with a permissions-oriented message, or
2. At minimum, surfacing `e.response.status_code` in the generic handler's result text so operators can self-diagnose.

tests/flightcheck/checks/test_preferred_solution.py
```diff
@@ -0,0 +1,315 @@
...
+    _register_solutions(solutions=[
+        _solution_record(SOLUTION_ID_ELIGIBLE, "ESSCustomization"),
+        _solution_record(SOLUTION_ID_OTHER, "ContosoExtensions"),
+    ])
+    _register_whoami()
+    _register_usersettings(selected_solution_id=SOLUTION_ID_OTHER)
+
+    r = _check_preferred_solution(runner)[0]
+    assert r.status == "Passed"
+    assert "'ContosoExtensions'" in r.result
```
nehaoss, line 271: The error-case tests (500 and 401 scenarios) directly reference `dv._query_url(...)` — a private helper — to construct mock URLs. If this internal function is renamed or refactored, these tests break without an obvious connection to the production code.

Consider either:
1. Exposing a public `build_query_url` alias in the `dv` mock module, or
2. Adding a brief comment here documenting that the coupling is intentional for URL-shape consistency.

solutions/ess-maker-skills/scripts/auth.py
```diff
@@ -217,6 +217,48 @@ def query_all(env_url, token, entity_set, select, filter_expr=None):
...
+        Parsed JSON response body.
+
+    Raises
+    ------
+    AuthExpiredError
+        If the response is 401.
+    requests.HTTPError
+        For other non-2xx responses (raised via ``raise_for_status``).
+    """
+    _validate_https_url(env_url)
```
nehaoss, line 252: `path.lstrip('/')` silently produces a doubled `/api/data/v9.2/` prefix if a caller accidentally passes a full API path (e.g. copy-pasting from a browser network tab). The docstring says 'no leading slash' but there's no runtime enforcement.

Consider a cheap assertion to fail fast on misuse:
```python
assert not path.startswith('/api/'), f'path must be relative: {path}'
```

nehaoss (commented):
Review: 5 actionable items flagged inline below.

solutions/ess-maker-skills/scripts/auth.py
```diff
@@ -217,6 +217,48 @@ def query_all(env_url, token, entity_set, select, filter_expr=None):
     return all_records
 
 
+def dataverse_get(env_url, token, path, params=None):
```
qazizaahirah, line 220: Do we actually need this? I am pretty sure we have functions for getting dataverse values?

qazizaahirah (dismissed):

solutions/ess-maker-skills/scripts/auth.py
qazizaahirah: Minor refactoring opportunity (out of scope, but worth noting): auth.py now has five sibling functions (query_all, update_record, create_record, delete_record, dataverse_get) that each duplicate the same 4-line preamble (_validate_https_url, header construction, URL prefix, 401-to-AuthExpiredError). A private _dv_request(env_url, token, method, path, ...) helper would collapse all five. PR #128 does not make this duplication worse — it follows the same convention as the existing four — but the duplication is now five-fold instead of four-fold. Worth a follow-up issue if anyone cares to clean it up; not something to block this PR on.

daeunJe0ng:
> @daeunJe0ng please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

solutions/ess-maker-skills/scripts/auth.py
daeunJe0ng: I actually wanted to reuse of one ([query_all](https://github.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/blob/main/solutions/ess-maker-skills/scripts/auth.py#L188-L217) in auth.py) but it does not handle a single non-paginated GET.
And I want to call `WhoAmI() `or a single-record read like `usersettingscollection({id})`, so made a separate one.

solutions/ess-maker-skills/scripts/auth.py
daeunJe0ng: I get what you mean, it can be `https://env/api/data/v9.2/api/data/v9.2/WhoAmI()`
Thanks for catching this, will make a change!

solutions/ess-maker-skills/scripts/auth.py
daeunJe0ng: > Minor refactoring opportunity (out of scope, but worth noting): auth.py now has five sibling functions (query_all, update_record, create_record, delete_record, dataverse_get) that each duplicate the same 4-line preamble (_validate_https_url, header construction, URL prefix, 401-to-AuthExpiredError). A private _dv_request(env_url, token, method, path, ...) helper would collapse all five. PR [#128](https://github.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/pull/128) does not make this duplication worse — it follows the same convention as the existing four — but the duplication is now five-fold instead of four-fold. Worth a follow-up issue if anyone cares to clean it up; not something to block this PR on.

Filed as #<[132](https://github.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/issues/132)> with the five functions enumerated and the preamble + 401-mapping + URL-validation contract that the shared helper needs to preserve. 
Linked your comment as the source.

solutions/ess-maker-skills/scripts/flightcheck/checks/environment.py
daeunJe0ng: Resolved, direct import requests removed.

solutions/ess-maker-skills/scripts/flightcheck/checks/environment.py
daeunJe0ng: Naturally resolved, WhoAmI no longer called.

solutions/ess-maker-skills/scripts/flightcheck/checks/environment.py
daeunJe0ng: Resolved differently, `[HTTP {status_code}]` surfaced in generic handler (`test_warning_when_get_preferred_solution_returns_403 `covers it)

tests/flightcheck/checks/test_preferred_solution.py
daeunJe0ng: Resolved, build_query_url public alias added

tests/mocks/dataverse.py
```diff
@@ -295,6 +301,82 @@ def whoami(*, base_url: str, **kwargs: Any) -> dict[str, Any]:
     }
 
 
+def usersettings(
```
qazizaahirah: this might be dead code now. I dont see it getting used anywhere

qazizaahirah (approved):

PR #130, opened 2026-06-08
flightcheck: triage-first report layout + actionable CONFIG-005/007 + env_id correctness

GrahamMcMynn:
## Why

Two real-tenant scenarios surfaced when running FlightCheck against a live ESS agent:

1. **The report was hard to act on.** All results — passed, skipped, failed, warning — were dumped in a flat list. An operator had to scan every row to find the ones that needed action.
2. **CONFIG-007 lied.** It reported "No instructions block found in agent.mcs.yml" for an agent that obviously had hundreds of words of instructions in Copilot Studio, and the remediation didn't even point at the right agent.

Both turned out to span more than the surface symptom (a regex-on-YAML bug, a silent fallback that fabricated a wrong env id, and a missed BAP hostname match). This PR fixes them and ships a triage-friendly report layout.

## What changed

**Report layout** ( runner.py, cli.py)
- Verdict banner + three sections: **Action required**, **Manual verification**, **Passed**.
- Operators can see at a glance whether ESS is healthy and, if not, what they need to do.
- JSON output gains `skipped` + `errors` fields so downstream tooling stays in parity.

**CONFIG-007 / CONFIG-005** (checks/local_files.py)
- Remediations now include a verified Copilot Studio deep link:
  `https://copilotstudio.microsoft.com/environments/{envId}/bots/{botId}/overview` (with graceful fallback to the homepage if env_id / bot_id can't be resolved).
- Replace the regex YAML probe with `yaml.safe_load`. The old pattern (`instructions:\s*[|>]?\s*\n…`) silently missed YAML's chomping-indicator block scalars (`|-`, `|+`, `>-`, `>+`) — the exact form Copilot Studio emits.

**env_id correctness** (pp_admin_client.py, cli.py)
- `find_environment_id_by_dataverse_url` now matches on **both** `instanceUrl` AND `instanceApiUrl`. BAP advertises the bare web hostname on one and the `.api.` Web API hostname on the other; configs typically hold the `.api.` form, so checking only `instanceUrl` silently missed every match on those tenants.
- `derive_environment_id` no longer silently falls through to the Dataverse OrganizationId when `pp_admin` is supplied but the lookup misses. That fabricated value was corrupting both BAP admin calls (404s) and any URL that embeds the env id. CLI now reports the miss explicitly and tells the operator how to override with `--environment-id`.

## Tests

- `test_runner_report_layout.py` (16): bucket assignment, verdict banner, JSON shape, AUTH-005 multi-row pattern.
- `test_local_files_studio_link.py` (9): deep-link helper + CONFIG-007 deep-link remediation.
- `test_local_files_agent_identity.py` (11): pins the `|-` / `|+` / `>-` / `>+` / quoted-string regressions and starter-prompt counting from parsed YAML.
- `test_pp_admin_client.py` (+6): pins `instanceApiUrl` matching and no-silent-OrgId-fallback behavior.

**299 passed / 8 skipped.** Ruff clean on touched files.

## Follow-ups (deliberately out of scope, will land separately)

- Remaining remediation-string improvements (the user is iterating on these one at a time).
- `run_local_file_checks` and `checks/servicenow.py` walk `workspace/agents/` on disk instead of iterating `runner.config["agents"]` — surfaces stale folders from prior `/setup` runs and misses agents listed in config.
- `workspace/agents/` vs `my/agents/` naming inconsistency across the kit (setup.py / docs / SKILL.md files).
- Avery's `-FlightCheckOnly` flow should make local-file checks SKIP cleanly when `folder=""`.

### DIFF
--- a/solutions/ess-maker-skills/scripts/flightcheck/runner.py
+++ b/solutions/ess-maker-skills/scripts/flightcheck/runner.py
@@ -144,10 +144,19 @@ class FlightCheckRunner:
         total_failed = sum(c.failed for c in cat_map.values())
         total_warnings = sum(c.warnings for c in cat_map.values())
         total_passed = sum(c.passed for c in cat_map.values())
+        # Tallied here so the verdict logic can consult errors. Errors
+        # (a check raised mid-run) mean we don't actually know whether
+        # ESS is healthy in that area, so they MUST count as
+        # "not ready" — not "ready" or "ready with warnings". Before
+        # this was added, an error-only run rendered as green READY
+        # with all the errored rows visible under ACTION REQUIRED
+        # directly below the green banner — exactly the at-a-glance
+        # contradiction the prioritized report is meant to eliminate.
+        total_errors = sum(c.errors for c in cat_map.values())
 
-        if total_failed == 0 and total_warnings == 0:
+        if total_failed == 0 and total_errors == 0 and total_warnings == 0:
             overall = "READY"
-        elif total_failed == 0:
+        elif total_failed == 0 and total_errors == 0:
             overall = "READY_WITH_WARNINGS"
         else:
             overall = "NOT_READY"
@@ -165,7 +174,7 @@ class FlightCheckRunner:
             not_configured=sum(c.not_configured for c in cat_map.values()),
             manual=sum(c.manual for c in cat_map.values()),
             skipped=sum(c.skipped for c in cat_map.values()),
-            errors=sum(c.errors for c in cat_map.values()),
+            errors=total_errors,
             overall=overall,
         )
 
--- a/solutions/ess-maker-skills/scripts/flightcheck/runner.py
+++ b/solutions/ess-maker-skills/scripts/flightcheck/runner.py
@@ -78,6 +78,8 @@ class RunResult:
     warnings: int = 0
     not_configured: int = 0
     manual: int = 0
+    skipped: int = 0
+    errors: int = 0
     overall: str = ""  # READY / READY_WITH_WARNINGS / NOT_READY
 
 
@@ -162,6 +164,8 @@ class FlightCheckRunner:
             warnings=total_warnings,
             not_configured=sum(c.not_configured for c in cat_map.values()),
             manual=sum(c.manual for c in cat_map.values()),
+            skipped=sum(c.skipped for c in cat_map.values()),
+            errors=sum(c.errors for c in cat_map.values()),
             overall=overall,
         )
 
@@ -184,6 +188,8 @@ def save_results(run_result: RunResult, output_dir: str = "workspace/flightcheck
         "warnings": run_result.warnings,
         "not_configured": run_result.not_configured,
         "manual": run_result.manual,
+        "skipped": run_result.skipped,
+        "errors": run_result.errors,
         "categories": [asdict(c) for c in run_result.categories],
         "results": [asdict(r) for r in run_result.results],
     }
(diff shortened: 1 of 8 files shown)


GrahamMcMynn (commented):

solutions/ess-maker-skills/scripts/flightcheck/runner.py
```diff
@@ -210,42 +216,165 @@ def save_results(run_result: RunResult, output_dir: str = "workspace/flightcheck
...
-            Priority.CRITICAL.value: "priority-critical",
-            Priority.HIGH.value: "priority-high",
-        }.get(res.priority, "")
 
-        remediation = _md_links_to_html(res.remediation or "")
-        if res.doc_link:
-            remediation += f' <a href="{res.doc_link}" target="_blank">[docs]</a>'
+def bucket_results(
+    results: list[CheckResult],
+) -> dict[str, list[CheckResult]]:
```
GrahamMcMynn, line 297: The overall verdict computation ignores `Status.ERROR` entirely, which means the new verdict banner -- documented at `_verdict_text` as the report's single biggest signal -- lies on error-only runs.

In `FlightCheckRunner.run`:

```python
if total_failed == 0 and total_warnings == 0:
    overall = "READY"
elif total_failed == 0:
    overall = "READY_WITH_WARNINGS"
else:
    overall = "NOT_READY"
```

A category whose check raises is recorded with `status=Status.ERROR.value` (line ~184). On an error-only run (no Failed, no Warning), `total_failed == 0 and total_warnings == 0` so `overall == "READY"` -- the HTML renders a green `verdict-ready` banner saying *"All N check(s) passed. Your environment looks ready to deploy."* and `_print_prioritized_summary` prints `[READY] Ready for deployment`, even while the ACTION REQUIRED section below shows the errored rows.

`READY_WITH_WARNINGS` has the same blind spot: an error + one warning would render *"Ready with warnings -- 1 warning(s) to review"* and never mention the errored check.

This is pre-existing in `_aggregate_categories`, but this PR (a) introduces the verdict banner as the headline signal, (b) starts tallying errors here on this very line, and (c) is the right moment to align overall classification with the new buckets. None of the new tests in `test_runner_report_layout.py` exercise an error-only run.

Suggested fix:

```python
total_errors = sum(c.errors for c in cat_map.values())
if total_failed == 0 and total_errors == 0 and total_warnings == 0:
    overall = "READY"
elif total_failed == 0 and total_errors == 0:
    overall = "READY_WITH_WARNINGS"
else:
    overall = "NOT_READY"
```

…and pin it with a regression test that asserts `result.overall == "NOT_READY"` for an Error-only run.

GrahamMcMynn: Good catch — confirmed. Pushed fix in 58ffa01:

- `runner.py`: tally `total_errors` alongside `total_failed` / `total_warnings` and include it in both verdict gates. An error-only or error-plus-warning run now correctly classifies as `NOT_READY`.
- `_verdict_text` and `_print_prioritized_summary` already handle errors in their `NOT_READY` branches (`r.failed + r.errors`), so no wording change needed downstream.
- Two regression tests in `test_runner_report_layout.py` pin the exact failure mode (error-only ⇒ `NOT_READY`) and the warnings-boundary variant (error+warning ⇒ `NOT_READY`, not `READY_WITH_WARNINGS`).

301 passed / 8 skipped. Ruff clean.

qazizaahirah (approved):

PR #134, opened 2026-06-09
FlightCheck installer: fetch agent solution before running checks

amilandi:
## Summary

The FlightCheck-only installer now calls `fetch_and_setup.py` to download the agent solution from Dataverse before running FlightCheck. This enables the **Local Files** checks (topics, variables, agent identity) that previously always showed as SKIPPED.

Also fixes the environment listing table always showing "Unknown" for the Type column.

## Problem

1. FlightCheck has two categories of checks:
   - **API-based** (Dataverse, Graph, Power Platform) — always worked
   - **Local file checks** (inspect `workspace/agents/`) — always SKIPPED because the installer never downloaded the solution

2. The environment picker table showed "Unknown" for every environment type because it read `environmentType` instead of `environmentSku` from the BAP API response.

## Changes

- `setup/Install-EssAdk.ps1`: Added fetch_and_setup call between config creation and FlightCheck run. Refactored python command resolution to top of FlightCheckOnly block for reuse.
- `setup/install-ess-adk.sh`: Added full discovery + config + fetch flow to macOS FlightCheck-only mode (mirrors the Windows behavior). Uses Python `json.dump()` for safe config JSON generation.
- `scripts/list_environments.py`: Read `environmentSku` (the actual BAP API field) instead of `environmentType`.

## Testing

- Tested end-to-end on Windows: environment picker -> agent picker -> fetch solution -> FlightCheck runs with Local Files checks enabled (e.g. CONFIG-007 now fires instead of LOCAL-001 SKIPPED).
- macOS flow mirrors Windows structure (same discovery/config/fetch steps).

### DIFF
--- a/setup/install-ess-adk.sh
+++ b/setup/install-ess-adk.sh
@@ -304,8 +304,7 @@ if [[ "$FLIGHTCHECK_ONLY" == "true" ]]; then
         fi
 
         # Re-run with --select
-        SELECT_OUTPUT=$("$FLIGHTCHECK_PYTHON" "$DISCOVER_PY" --list-environments --select "$ENV_CHOICE" 2>&1)
-        if [[ $? -ne 0 ]]; then
+        if ! SELECT_OUTPUT=$("$FLIGHTCHECK_PYTHON" "$DISCOVER_PY" --list-environments --select "$ENV_CHOICE" 2>&1); then
             err "Environment selection failed."
             exit 1
         fi
@@ -331,8 +330,7 @@ if [[ "$FLIGHTCHECK_ONLY" == "true" ]]; then
         echo "    Discovering agents in this environment..."
         echo ""
 
-        AGENT_OUTPUT=$("$FLIGHTCHECK_PYTHON" "$DISCOVER_PY" --url "$ENV_URL" 2>&1) || true
-        AGENT_EXIT=$?
+        AGENT_OUTPUT=$("$FLIGHTCHECK_PYTHON" "$DISCOVER_PY" --url "$ENV_URL" 2>&1) && AGENT_EXIT=0 || AGENT_EXIT=$?
         echo "$AGENT_OUTPUT"
 
         if [[ $AGENT_EXIT -ne 0 ]] || echo "$AGENT_OUTPUT" | grep -q "^ERROR:"; then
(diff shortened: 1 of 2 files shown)


setup/install-ess-adk.sh
```diff
@@ -233,21 +233,210 @@ if [[ "$FLIGHTCHECK_ONLY" != "true" ]]; then
...
+        read -rp "    Select environment number from the list above: " ENV_CHOICE
+        ENV_CHOICE="${ENV_CHOICE// /}"
+        if [[ -z "$ENV_CHOICE" || ! "$ENV_CHOICE" =~ ^[0-9]+$ ]]; then
+            err "Invalid selection. Please re-run and pick a number."
+            exit 1
+        fi
+
+        # Re-run with --select
+        SELECT_OUTPUT=$("$FLIGHTCHECK_PYTHON" "$DISCOVER_PY" --list-environments --select "$ENV_CHOICE" 2>&1)
+        if [[ $? -ne 0 ]]; then
```
GrahamMcMynn: **Bug (set -e):** This graceful error handler is dead code. The bare assignment `SELECT_OUTPUT=$("$FLIGHTCHECK_PYTHON" "$DISCOVER_PY" --list-environments --select "$ENV_CHOICE" 2>&1)` runs under `set -euo pipefail` (line 32), so if `discover.py` exits non-zero the script aborts at the assignment line and this `if [[ $? -ne 0 ]]; then err "Environment selection failed."; exit 1; fi` block never executes — the user sees no friendly message.

Guard the assignment instead:

```bash
if ! SELECT_OUTPUT=$("$FLIGHTCHECK_PYTHON" "$DISCOVER_PY" --list-environments --select "$ENV_CHOICE" 2>&1); then
    err "Environment selection failed."
    exit 1
fi
```

setup/install-ess-adk.sh
```diff
@@ -233,21 +233,210 @@ if [[ "$FLIGHTCHECK_ONLY" != "true" ]]; then
...
+        ok "Environment: $ENV_DISPLAY"
+        ok "URL: $ENV_URL"
+
+        # --- Step 2: List agents ---
+        echo ""
+        echo "    Discovering agents in this environment..."
+        echo ""
+
+        AGENT_OUTPUT=$("$FLIGHTCHECK_PYTHON" "$DISCOVER_PY" --url "$ENV_URL" 2>&1) || true
+        AGENT_EXIT=$?
```
GrahamMcMynn: **Bug:** `AGENT_EXIT` is always 0, so the `if [[ $AGENT_EXIT -ne 0 ]]` check below is dead. Because the assignment above ends in `|| true`:

```bash
AGENT_OUTPUT=$(... 2>&1) || true
AGENT_EXIT=$?
```

`$?` reflects the `|| true` (always 0), not `discover.py`'s exit code. Agent-discovery failures are then only caught by the `grep -q "^ERROR:"` half of the condition; a non-zero exit without an `ERROR:` line slips through.

Capture the real exit code, e.g.:

```bash
AGENT_OUTPUT=$("$FLIGHTCHECK_PYTHON" "$DISCOVER_PY" --url "$ENV_URL" 2>&1) && AGENT_EXIT=0 || AGENT_EXIT=$?
```

setup/install-ess-adk.sh
```diff
@@ -233,21 +233,210 @@ if [[ "$FLIGHTCHECK_ONLY" != "true" ]]; then
...
+        fi
+
+        cat > "$CONFIG_PATH" <<EOF
+{
+    "configVersion": 1,
+    "setup": "flightcheck-only",
+    "dataverseEndpoint": "$ENV_URL",
+    "flightCheckOnly": true,
+    "agent": {
+        "name": "$AGENT_NAME",
```
GrahamMcMynn: **Bug (robustness):** This heredoc interpolates `$AGENT_NAME` (and `$SCHEMA_NAME`) directly into JSON without escaping. Power Platform agent display names are free-form and may contain `"` or `\`; if they do, the generated `.local/config.json` becomes invalid JSON and breaks every downstream consumer (the re-run config read, FlightCheck). The PowerShell path avoids this by building a hashtable and using `ConvertTo-Json`, which escapes safely — this path should match.

Emit the config with a JSON-aware tool instead of string interpolation (e.g. `"$FLIGHTCHECK_PYTHON" -c 'import json,sys; ...'` or `jq`).

GrahamMcMynn (commented):
Code review by ESS Code Review Agent. Both new bugs are surfaced by set -euo pipefail in install-ess-adk.sh.

setup/install-ess-adk.sh
amilandi: Fixed in 720b100 — using `if ! SELECT_OUTPUT=$(...)` pattern so the error handler fires under `set -e`.

setup/install-ess-adk.sh
amilandi: Fixed in 720b100 — using `cmd && AGENT_EXIT=0 || AGENT_EXIT=$?` to capture the real exit code.

setup/install-ess-adk.sh
amilandi: Fixed in 720b100 — replaced heredoc with inline Python `json.dump()` that passes all values as `sys.argv` and lets the json module handle escaping safely.

GrahamMcMynn (approved):

PR #135, opened 2026-06-09
User-friendly HTTP error messages for API failures

amilandi:
## Summary

ADO tasks https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7469877 and https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7469904

Adds a centralized HTTP error handler that translates raw status codes into actionable messages with troubleshooting tips. This significantly improves the experience when users hit permission or connectivity issues during install/setup.

## Problem

When API calls fail (e.g., user lacks Dataverse read permission), the scripts showed raw tracebacks like:
`
requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: https://org.crm.dynamics.com/api/data/v9.2/bots?=...
`

Non-technical users don't know what a 403 means or how to fix it.

## Solution

New `scripts/http_errors.py` module with:
- `APIError` class (subclasses `requests.HTTPError` for backward compat)
- `raise_api_error(response, resource_name, operation, required_role)` helper
- Friendly entity name mapping (`bots` → "agents", `workflows` → "cloud flows")
- Operation-aware messages (read/create/update/delete)
- Compact technical detail line for debugging

### Example output (403 on bots query):
`
  ERROR: You signed in successfully, but your account doesn't have permission to read agents.
  Tip:   Ask your Power Platform admin to assign the Bot Author, Bot Contributor, or System Administrator security role to your account in this environment.
  Detail: HTTP 403 — GET https://org.crm.dynamics.com/api/data/v9.2/bots
`

### Status codes covered:
| Code | Message theme |
|------|--------------|
| 400 | Version mismatch / bad query |
| 401 | Session expired, re-sign-in |
| 403 | Permission denied + role suggestion |
| 404 | Resource/solution not deployed |
| 429 | Rate limiting + wait guidance |
| 5xx | Transient server error + retry advice |

## Files changed
- **New:** `scripts/http_errors.py` — central error handler
- `scripts/auth.py` — replace `raise_for_status()` with `raise_api_error()`
- `scripts/discover.py` — catch `APIError`, show friendly output
- `scripts/fetch_and_setup.py` — catch `APIError` in refresh + normal modes

## Design decisions
- FlightCheck clients (graph_client, pp_admin_client) NOT changed — they already handle errors via `_error` dicts
- `sync_docs.py` / `sync_samples.py` NOT changed — internal tooling, not user-facing during install
- 401 still uses existing `AuthExpiredError` (callers depend on it for re-auth)
- `APIError` subclasses `HTTPError` so existing `except HTTPError` handlers still work

## Testing
Verified via unit mock: all status codes produce correct messages, 2xx doesn't raise, `except HTTPError` still catches `APIError`.

### DIFF
--- a/solutions/ess-maker-skills/scripts/auth.py
+++ b/solutions/ess-maker-skills/scripts/auth.py
@@ -34,6 +34,8 @@ except ImportError:
     print("ERROR: 'urllib3' / 'requests' not found. Run: pip install requests")
     sys.exit(1)
 
+from http_errors import raise_api_error  # noqa: E402
+
 
 # Microsoft public client ID for Power Platform CLI / Dataverse delegated access.
 # Source: https://learn.microsoft.com/power-platform/admin/programmability-authentication-v2
@@ -203,7 +205,7 @@ def query_all(env_url, token, entity_set, select, filter_expr=None):
         resp = _SESSION.get(url, headers=headers, timeout=120, verify=True)
         if resp.status_code == 401:
             raise AuthExpiredError("Dataverse returned 401 (token expired or invalid)")
-        resp.raise_for_status()
+        raise_api_error(resp, resource_name=entity_set, operation="read")
         data = resp.json()
         records = data.get("value", [])
         all_records.extend(records)
@@ -324,7 +326,7 @@ def update_record(env_url, token, entity_set, record_id, data):
     resp = _SESSION.patch(url, headers=headers, json=data, timeout=60, verify=True)
     if resp.status_code == 401:
         raise AuthExpiredError("Dataverse returned 401 (token expired or invalid)")
-    resp.raise_for_status()
+    raise_api_error(resp, resource_name=entity_set, operation="update")
     return True
 
 
@@ -341,7 +343,7 @@ def create_record(env_url, token, entity_set, data):
     resp = _SESSION.post(url, headers=headers, json=data, timeout=60, verify=True)
     if resp.status_code == 401:
         raise AuthExpiredError("Dataverse returned 401 (token expired or invalid)")
-    resp.raise_for_status()
+    raise_api_error(resp, resource_name=entity_set, operation="create")
     result = resp.json()
     return result.get("botcomponentid", result.get("id"))
 
@@ -357,7 +359,7 @@ def delete_record(env_url, token, entity_set, record_id):
     resp = _SESSION.delete(url, headers=headers, timeout=60, verify=True)
     if resp.status_code == 401:
         raise AuthExpiredError("Dataverse returned 401 (token expired or invalid)")
-    resp.raise_for_status()
+    raise_api_error(resp, resource_name=entity_set, operation="delete")
     return True
 
 
--- a/solutions/ess-maker-skills/scripts/auth.py
+++ b/solutions/ess-maker-skills/scripts/auth.py
@@ -34,7 +34,7 @@ except ImportError:
     print("ERROR: 'urllib3' / 'requests' not found. Run: pip install requests")
     sys.exit(1)
 
-from http_errors import raise_api_error  # noqa: E402
+from http_errors import APIError, raise_api_error  # noqa: E402
 
 
 # Microsoft public client ID for Power Platform CLI / Dataverse delegated access.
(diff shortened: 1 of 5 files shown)


solutions/ess-maker-skills/scripts/auth.py
```diff
@@ -203,7 +205,7 @@ def query_all(env_url, token, entity_set, select, filter_expr=None):
         resp = _SESSION.get(url, headers=headers, timeout=120, verify=True)
         if resp.status_code == 401:
             raise AuthExpiredError("Dataverse returned 401 (token expired or invalid)")
-        resp.raise_for_status()
+        raise_api_error(resp, resource_name=entity_set, operation="read")
```
qazizaahirah, line 226: 401 still produces a raw traceback — the exact failure mode the PR aims to fix. Here, the 401 check raises AuthExpiredError BEFORE reaching raise_api_error, but the new try/except in discover.py   and fetch_and_setup.py only catches APIError. Walking origin/main, neither caller has an except AuthExpiredError handler. So token-expired 401s escape the friendly handler entirely. PR description claims   "callers depend on it for re-auth" but no such handler exists in the changed files. Fix: catch both, or make AuthExpiredError inherit from APIError.

solutions/ess-maker-skills/scripts/http_errors.py
```diff
@@ -0,0 +1,192 @@
...
+    "bots": "agents",
+    "botcomponents": "agent components",
+    "workflows": "cloud flows",
+    "msdyn_employeeselfservicetemplateconfigs": "ESS template configurations",
+    "connections": "connections",
+    "connectionreferences": "connection references",
+}
+
+
+def _friendly_name(resource_name):
```
qazizaahirah, line 30: Should we add unit tests here

solutions/ess-maker-skills/scripts/http_errors.py
```diff
@@ -0,0 +1,192 @@
...
+            "",
+            f"  ERROR: {self._friendly_message}",
+        ]
+        if self._tip:
+            lines.append(f"  Tip:   {self._tip}")
+        # Technical detail line (compact, for debugging)
+        if self.url:
+            # Truncate URL to avoid leaking full query strings
+            display_url = self.url.split("?")[0]
+            method_verb = (self.operation or "GET").upper()
```
qazizaahirah: method_verb is reverse-engineered from operation string instead of read from response.request.method. Fragile and indirect; the cleaner alternative is one-liner

amilandi: Fixed in 90d4c19. `APIError.__init__` now stores `self._method = response.request.method` when a response is available, and `format_for_terminal` uses `method_verb = (self._method or "GET").upper()` — the 4-branch if/elif chain is gone. Test `test_detail_uses_request_method_not_operation` pins the new behavior (`operation="read"` combined with `method="POST"` correctly displays POST).

solutions/ess-maker-skills/scripts/http_errors.py
amilandi: Added in 90d4c19 — see `tests/scripts/test_http_errors.py`. 41 tests covering: per-status-code messages and tips (400/401/403/404/429/5xx/unknown), the entity-set → friendly-name mapping, `format_for_terminal` output structure (including URL query-string stripping and graceful handling of no-response cases), `raise_api_error` no-op on 2xx and raise on 4xx/5xx, `APIError` still catchable as `requests.HTTPError` for back-compat, and the new `AuthExpiredError` subclassing contract. Pure-logic tests — no HTTP calls, no cassettes needed.

solutions/ess-maker-skills/scripts/auth.py
amilandi: Fixed in 90d4c19. `AuthExpiredError` now inherits from `APIError`, so `except APIError` in `discover.py` / `fetch_and_setup.py` also catches 401s and renders the friendly `format_for_terminal()` output. `except AuthExpiredError` in `push.py` / FlightCheck still works for re-auth-and-retry (subclass relationship preserves it). All 6 raise sites in `auth.py` now pass `response=resp` so 401s carry the same URL / method / request-id diagnostics as other API errors.

qazizaahirah (approved):

PR #138, opened 2026-06-09
Flightcheck: add WD-WF-CAT-001 Workday custom-workflow inventory checklist

apurvabanka:
## Summary
 
 Adds `WD-WF-CAT-001` — a MANUAL FlightCheck row that enumerates Workday
 scenarios wired up in customer topics that fall **outside** the 17 OOTB
 workflows the kit ships SOAP tests for. The check walks every agent topic
 under `workspace/agents/*/topics/*.mcs.yml` for two wiring patterns,
 diffs the discovered scenarios against the tenant-live OOTB catalog
 (read from Dataverse `msdyn_employeeselfservicetemplateconfigs` where
 `ismanaged=true`), and surfaces any unknowns as a bucketed MANUAL row
 with a four-item operator checklist plus a gap-discovery loop-back.
 
 A companion `WD-WF-CAT-LINK` trailer fires from inside `_check_workflows`
 so admins reading a clean 17-row SOAP-test pass don't miss the manual
 row below it.
 
 Addresses **6 incidents** rooted in this gap pattern (custom workflow
 names not in the 41-workflow catalog), notably:
 
 - `760098889` — Multiday Leave application topic
 - `783902203` — conversation flow specifics
 
 Parent feature: **AB#7392268**.
 Action owner for MANUAL row triage: **Workday admin**.
 
 ## Acceptance criteria
 
 - [x] **AC1** — Checklist enumerates: which custom Workday workflows are
       wired up, ISU account used, expected payload shape, test prompt
       *(pinned by `test_unknown_scenario_emits_manual_with_full_checklist`)*
 - [x] **AC2** — Linked from Test-WorkdayWorkflows output when an unknown
       workflow name is referenced in customer topics
       *(pinned by `test_trailer_emitted_when_unknowns_exist` /
       `test_trailer_absent_when_clean`)*
 - [x] **AC3** — Includes a "found a new pattern? log it here" loop back
       to the gap-discovery process
       *(pinned by `test_checklist_includes_gap_discovery_loopback`)*
 
 ## Failure modes covered in the remediation text
 
 | Ticket failure mode | Checklist item that addresses it |
 |---|---|
 | **Problem 1 — Workflow Contract/Payload Mismatch** (adaptive card submit fails because the request payload doesn't match the Workday SOAP/template contract) | Item 2 (`Payload shape`) — directs operator to open the template config XML and confirm field names, reference types, and required vs. optional elements match the Workday WSDL. Names "the 'Workflow Contract/Payload Mismatch' failure mode" verbatim. |
 | **Problem 2 — Intermittent Workday Authentication Failure** (expired/revoked OAuth token state on one or more connection references) | Item 4 (`Auth health`) — cross-links to `WD-CONN-*` connection token-health output and gives the Power Platform Maker → Connections re-authenticate click-path. |
 
 ## What changed
 
 | File | Purpose |
 |---|---|
 | `solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py` | New `_check_custom_workflow_inventory` (WD-WF-CAT-001) + `_append_wd_wf_cat_link_trailer` (WD-WF-CAT-LINK) + topic/workflow walkers + Dataverse catalog resolver. Wired into `run_workday_checks` after the SOAP tests. |
 | `solutions/ess-maker-skills/src/reference/ess-docs/flightcheck/validation-matrix.md` | Adds WD-WF-CAT-001 + WD-WF-CAT-LINK rows under §5 Workday Deep Validation with status-decision matrix prose. |
 | `tests/flightcheck/checks/test_workday_custom_inventory.py` | 15 tests covering every branch (gates, MANUAL, PASS, WARNING, both patterns, bucketing, caching, all three ACs, all 4 Dataverse-catalog resolver legs). |
 
 ## Design notes
 
 ### Two detection patterns
 
 - **Pattern A** — topic calls
   `WorkdaySystemGetCommonExecution` with a `scenarioName:` value. Walker
   finds the `dialog: ...WorkdaySystemGetCommonExecution` line and looks
   backward up to 30 lines for the nearest `scenarioName:`.
 - **Pattern B** — topic uses `kind: InvokeFlowAction` with a `flowId` that
   resolves to a `workflow.json` whose `connectionReferences` bind to
   `shared_workdaysoap`. Conservative qualification — non-Workday flows
   (e.g. ServiceNow-bound) are ignored so the check doesn't emit false
   positives.
 
 ### Catalog source — live Dataverse, no fallback
 
 The OOTB scenario catalog is resolved per-invocation from the customer's
 own tenant by querying `msdyn_employeeselfservicetemplateconfigs` and
 treating every row with `ismanaged=true` as OOTB. This auto-detects every
 scenario the installed Workday extension pack ships — no kit-side
 curation, no stale JSON catalog to maintain. Result is cached on
 `runner._workday_ootb_catalog_cache` so the cross-link trailer can
 re-read without re-querying.
 
 ### Status decision matrix (per AGENTS.md principles #1, #2, #3, #11)
 
 | Tenant state | Status | Why |
 |---|---|---|
 | Simplified Workday install (1 conn ref, OBO) | SKIPPED | ISU/scenario inventory doesn't apply (gated on `runner._workday_package_flavor == "simplified"` per principle #11) |
 | `workspace/agents/` missing | SKIPPED | Run `/setup` first |
 | Zero Workday refs in any topic | SKIPPED | No Workday wired into this agent |
 | Workday refs exist but no Dataverse token | SKIPPED | Cannot resolve catalog → can't claim PASS (principle #1) |
 | Dataverse query errored (500, etc.) | WARNING | Surface error verbatim (principle #3) |
 | Every discovered ref is a managed row | PASSED | — |
 | Any unknown / Pattern-B ref | MANUAL | 4-item checklist + AC3 loop-back; bucketed to ONE row per principle #7 |


nehaoss (commented):
Reviewed the WD-WF-CAT-001 implementation. Left inline comments on actionable items — mostly around a duplicated filtering logic that should be extracted, a regex that may miss hyphenated scenario names, and a fragile state-machine parser for Pattern B. No blocking issues but these would improve maintainability.

amilandi (approved):
Approving. Substantial well-tested feature — all 3 ACs explicitly pinned, decision matrix correct.

**What I liked:**
- **No-fallback Dataverse catalog** is the right call — drift between a stale JSON catalog and the customer's installed extension pack would cause silent false MANUAL/PASS verdicts. The decision matrix (SKIPPED on no-token, WARNING on query-error per AGENTS.md principle #1/#3) is correct.
- **Bucketing + cross-link trailer** — multiple unknowns collapse to one MANUAL row (principle #7); `WD-WF-CAT-LINK` fires from inside `_check_workflows` so admins reading a clean 17-row SOAP pass can't miss the manual row below. The trailer self-suppresses on no-token/no-discovery so the main check stays single-source-of-truth.
- **AC3 test cites the commit (`0fb2383`) that previously stripped the loop-back paragraph.** That's the right kind of regression-guard rationale — it tells the next refactor-er WHY this specific text matters.
- **Pattern B is conservatively qualified** — `test_invoke_flow_action_non_workday_is_ignored` ensures ServiceNow-bound flows don't leak into the Workday inventory.
- **All 4 catalog-resolver legs covered** (managed/unmanaged-excluded/query-failure/no-token).

**Non-blocking concerns to consider in follow-ups:**

1. **`_scan_topic_for_workday_refs` scenarioName regex is `[\w]+` only** (workday.py:3129). Won't match hyphens or dots. Today's `msdyn_*` names are snake_case so this is fine, but a one-character fix to `[\w.-]+` would future-proof against any naming-convention change.

2. **InvokeFlowAction block-end heuristic** (workday.py:3148-3174) assumes the next `- kind:` line closes the block. Today's topics don't nest list structures inside an InvokeFlowAction `input`/`output` binding, but if that ever changes, the walker would silently miss the `flowId`. Consider adding a test for this shape OR scoping the block by YAML indentation.

3. **workday.py is now 3700+ lines** after this +661 addition. Approaching the size where splitting into a `workday/` package (`workflows.py`, `custom_inventory.py`, `connections.py`, etc.) would help maintainability. Not in scope here.

4. **`_load_workday_ootb_catalog_from_dataverse` lazy-imports `auth`** via `sys.path.insert`. Pre-existing pattern in 3 other places in this file, so not introduced here — but the tech debt is now 4 locations. Worth a follow-up PR to move to module-level once.

5. Minor: `except (json.JSONDecodeError, OSError)` in `_is_workday_bound_workflow_json` silently returns False on a malformed `workflow.json`. A debug-level log would help an operator diagnose why a flow was silently excluded.

PR #139, opened 2026-06-10
Installer reliability: ARM64 pip, --single-branch self-heal, account-switch prompt

amilandi:
## Summary

Three installer reliability / UX fixes uncovered while debugging install failures on two different machines.

### Issue 1 — Windows ARM64 pip install failure

A user running the installer on Windows ARM64 hit:

```
error: linker `link.exe` not found
ERROR: Failed building wheel for cryptography
[warn] pip install returned exit code 1 (non-fatal, /setup will retry)
```

**Root cause:** winget installs the native ARM64 build of Python on ARM64 hosts. `msal` pulls `cryptography` transitively; `cryptography` only began shipping `cp38-abi3-win_arm64` wheels in **46.0** (Aug 2025). Anything resolving below that falls back to an sdist → Rust + maturin bootstrap → MSVC linker → fail. The installer just warned and continued, leaving the kit half-installed.

### Issue 2 — `--single-branch` clone trap (branch switching silently broken)

**What the trap actually breaks: branch switching, not main refresh.**

`git clone --branch X --single-branch` persists a narrowed refspec into `.git/config`:

```ini
[remote "origin"]
    fetch = +refs/heads/X:refs/remotes/origin/X   # ← only X, forever
```

Future `git fetch origin` only refreshes branch `X`. As long as you stay on branch `X`, everything works — `fetch + pull --ff-only` keeps it current. **The trap fires only when you try to use a *different* branch.**

How it bit us in practice:

1. The user originally ran the installer with `-Branch amilandin/pip-arm64-resilience` to test that PR. The clone's refspec was narrowed to that feature branch.
2. Later they re-ran the bootstrap with the default `-Branch main` to compare against main.
3. Installer ran `git fetch origin` (fetched only the feature branch — not main), then `git checkout main` — which **failed** with `did not match any file(s) known to git`, because `origin/main` was never in local refs.
4. Script logged `Continuing on current branch` and quietly kept running on the older feature-branch code.
5. The user saw "main is showing unknown env types again" and "browser doesn't open after FlightCheck" — both bugs already fixed in PR #137 — and reasonably assumed main had regressed. In reality they were running stale feature-branch code, not stale main.

(Note: the "stale main" *symptom* on a different occasion was a separate root cause — raw.githubusercontent.com CDN caching of bootstrap/installer files for ~5 min after a push. That's not fixed by this PR; it self-resolves with time.)

### Issue 3 — No way to switch accounts without finding the cache file

All FlightCheck clients share `.local/.token_cache.bin`. Once a user signs in, every subsequent run silently reuses those tokens. Customer engineers and partners often install under one account and then want to FlightCheck a different customer tenant — today they have to know about the cache file path and delete it manually.

### `solutions/ess-maker-skills/scripts/requirements.txt`
- Pin `cryptography>=46.0.0` on `sys_platform == "win32" and platform_machine == "ARM64"` so resolution lands on a wheel-having version. Other platforms keep msal''s wider bound.

### `setup/Install-EssAdk.ps1` — pip install resilience (Issue 1)
- New `Test-IsWindowsArm64` helper (`RuntimeInformation.OSArchitecture` with `PROCESSOR_ARCHITECTURE` env-var fallback).
- New `Install-PipRequirements` helper that:
  1. Upgrades pip first. Failures warn-and-continue so a transient pip issue can''t mask a real install error.
  2. Installs with `--prefer-binary` so resolution favors wheels over sdists.
  3. On ARM64, also adds `--only-binary cryptography` to fail fast instead of spending a minute downloading Rust.
  4. On failure, prints actionable ARM64 remediation (install x64 Python via winget `--architecture x64` or python.org and rely on Prism emulation; OR install VS Build Tools).
- Replaces the two duplicated inline pip-install blocks (pre-clone path and deferred post-clone path) with calls to the new helper.

### `setup/Install-EssAdk.ps1` — branch-switch self-heal (Issue 2)
- Drop `--single-branch` from fresh `git clone`. Repo is small; being able to switch branches later is worth the marginal clone-time cost.
- On existing clones, run `git remote set-branches origin '*'` BEFORE `git fetch`. Idempotent — broadens an old single-branch refspec to all branches, no-op if already broad. Self-heals existing trapped users on their next install run, so anyone who first installed from a feature branch can now switch to main (or any other branch) successfully.
- Make the `git checkout` failure warning loud and actionable: name the current branch, state explicitly that STALE code from a *different* branch will run, give an exact `Remove-Item -Recurse -Force` recovery command. Same treatment for `git pull` failures.

### `setup/Install-EssAdk.ps1` — switch-account prompt (Issue 3)
- In the FlightCheck-only flow, after the reconfigure-config prompt and before any auth-requiring step, check for `.local/.token_cache.bin`. If present, ask `Sign in as a different account? (y/N)`. Default N (keep cached).
- On Y, remove the cache file and print a heads-up about browser auto-SSO (cookies may silently sign the user back in to the previous account; suggests sign-out or InPrivate window).
- Placement runs once before both code paths that touch MSAL: the discover.py env/agent listing AND the FlightCheck run itself.

## Validation

- ✅ Full Python test suite: **478 passed, 8 skipped**.
- ✅ PowerShell parse-check on `Install-EssAdk.ps1` clean.
- ✅ `Install-PipRequirements` smoke-tested end-to-end on an x64 host with a tmp `requirements.txt`: pip self-upgrade runs, requirements install succeeds, exit code returns 0.
- ✅ Self-heal logic verified live against the affected user''s install: refspec broadened from `+refs/heads/amilandin/pip-arm64-resilience:refs/remotes/origin/amilandin/pip-arm64-resilience` to `+refs/heads/*:refs/remotes/origin/*`, fetch picked up all branches, `git checkout main` now succeeds and resolves to PR #137''s merge commit.

### DIFF
--- a/setup/Install-EssAdk.ps1
+++ b/setup/Install-EssAdk.ps1
@@ -156,6 +156,102 @@ function Resolve-Python {
     return $null
 }
 
+# Helper: detect Windows ARM64 host. Used to add ARM64-specific guardrails to
+# pip install (cryptography only shipped win_arm64 wheels in 46.0+; older
+# resolutions fall back to a Rust source-build that needs VS Build Tools).
+function Test-IsWindowsArm64 {
+    try {
+        return ([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture.ToString() -eq 'Arm64')
+    } catch {
+        # Older .NET Framework without RuntimeInformation. Fall back to env var.
+        return ($env:PROCESSOR_ARCHITECTURE -eq 'ARM64')
+    }
+}
+
+# Helper: install pip requirements with ARM64-aware guardrails.
+#   - Upgrades pip first (failures are warnings, not fatal — older pip can still install).
+#   - Uses --prefer-binary so resolution favors wheels over sdists when possible.
+#   - On Windows ARM64, adds --only-binary cryptography. cryptography source
+#     builds need Rust + the MSVC linker (link.exe), which most ARM64 dev
+#     boxes don't have. Failing fast with a wheel-missing message is better
+#     than a 60-second Rust download + cryptic linker error.
+#   - On failure, prints actionable remediation (especially for ARM64).
+# Returns the pip install exit code so callers can decide what to do.
+function Install-PipRequirements {
+    param(
+        [Parameter(Mandatory)] [string] $PythonExe,
+        [Parameter(Mandatory)] [string] $RequirementsFile
+    )
+
+    # Normalize launcher invocation. 'py -3.12' / 'py -3' come back as space-
+    # separated strings from Resolve-Python; everything else is an absolute path.
+    if ($PythonExe -eq 'py -3.12' -or $PythonExe -eq 'py -3') {
+        $pyExe = 'py'
+        $pyPrefix = @(($PythonExe -split ' ')[1])
+    } else {
+        $pyExe = $PythonExe
+        $pyPrefix = @()
+    }
+
+    # 1. Pip self-upgrade. Warn-and-continue on failure so a transient pip
+    # issue doesn't mask a requirements install failure (or block the install
+    # entirely when the existing pip would have been good enough).
+    Start-Spinner 'upgrading pip'
+    $upgradeOutput = Invoke-Native {
+        & $pyExe @pyPrefix -m pip install --upgrade --quiet --disable-pip-version-check pip
+    }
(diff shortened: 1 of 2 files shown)


qazizaahirah:
I dont know if we need Issue 2 as part of this PR, because after cloning is the responsibility of the user of the script to make sure they have the latest script

qazizaahirah (approved):

PR #140, opened 2026-06-10
test(flightcheck): add tests for shared connection helpers + ServiceNow

qazizaahirah:
Adds unit and integration coverage for `flightcheck/checks/connections.py` (the shared helpers behind every connector-specific check) and for the ServiceNow consumer in `flightcheck/checks/servicenow.py`. Workday was already covered by `test_workday_connections.py`; this PR fills the gap for ServiceNow and the underlying generic helper that any new connector check will route through.

New tests
---------

* `tests/flightcheck/checks/test_connections_helpers.py` (26 tests) --- pure-unit coverage of `get_connection_status` (8 cases: Connected / Error / empty list / missing properties / non-list / missing status field / verbatim passthrough) and `filter_connections_by_connector` (7 cases: string keyword, list keyword, case-insensitive, multi-alias OR, displayName vs apiId, no match, malformed records). Plus end-to-end coverage of `check_connector_connections` via `responses`-mocked PowerApps responses: SKIPPED (no env_id / no pp_admin), WARNING (stubbed `_error` dict + raised exception), NOT_CONFIGURED (no matches), PASSED (all connected), FAILED (all errored), Mixed (PASS summary + per-conn FAIL detail), non-matching connections filtered from counts, prefix/category/doc_link propagation.

* `tests/flightcheck/checks/test_servicenow_connections.py` (13 tests) --- end-to-end via `servicenow._check_connections`. Pins the ServiceNow-specific wiring (`["service-now", "servicenow"]` keyword aliases, `SN-CONN` prefix, `ServiceNow` category, `/connect servicenow` remediation, ServiceNow integration doc link). Covers good (single + HRSD+ITSM), bad (all errored, empty env, non-SN-only env), mixed-state (PASS summary + per-conn FAIL), metadata propagation, and skip paths.

* `tests/mocks/pp_admin.py` --- adds `servicenow_connection()` convenience builder mirroring `workday_connection` / `non_workday_connection`. Cites consumers per the cardinal-rule traceability convention. Defaults to `shared_service-now` apiId but exposes `api_name` so tests can pin the unhyphenated `shared_servicenow` alias.

One pinned latent behavior
--------------------------

`test_http_403_silently_degrades_to_not_configured` pins the current behavior where a real HTTP 401/403 from `/connections` is swallowed by `PPAdminClient._get_all` (lines 183-184) into an empty list and surfaces as `NotConfigured` --- indistinguishable from "tenant has no ServiceNow / Workday connections at all". This is the same silent- failure mode AUTH-006 and ENV-009 added explicit probe queries to work around. The test includes a `TODO` so when `get_connections` grows a permissions probe, the assertion can be flipped to `Warning` deliberately.

Verification
------------

* `pytest tests/` --- 517 passed, 8 skipped (no regressions; same totals as origin/main before this PR).
* `pytest tests/flightcheck/checks/test_connections_helpers.py tests/flightcheck/checks/test_servicenow_connections.py` --- 39 passed.
* `ruff check` on all three touched files --- clean.

## Related issue

Fixes #

## Type of change
- [ ] Documentation update
- [ ] Refactor / cleanup

## Checklist
- [ ] My code follows the existing style
- [ ] I have updated documentation as needed

```text
Validation
- YAML parse: PASS|FAIL|N-A
- AdaptiveDialog kind: PASS|FAIL|N-A
- XML parse: PASS|FAIL|N-A
- Filename convention (new): PASS|FAIL|N-A
- Folder convention (new, incl. README.md): PASS|FAIL|N-A
- Diff scope (samples/ only): PASS|FAIL|N-A
- Secrets / internal URLs: PASS|FAIL|N-A
```

### DIFF
--- a/tests/flightcheck/checks/test_connections_helpers.py
+++ b/tests/flightcheck/checks/test_connections_helpers.py
@@ -1,0 +1,595 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""Unit + integration tests for the shared connection-status helpers in
+``solutions/ess-maker-skills/scripts/flightcheck/checks/connections.py``.
+
+This module owns three helpers used by every connector-specific check
+(workday.py, servicenow.py, environment.py, and any future connector):
+
+* ``get_connection_status(conn)`` — pure dict navigation; pulls the
+  status string out of a BAP connection record.
+* ``filter_connections_by_connector(all_conns, keyword)`` — filters a
+  list of connections by case-insensitive substring match against
+  ``properties.apiId`` + ``properties.displayName``. Accepts a single
+  keyword or a list of keywords (OR semantics).
+* ``check_connector_connections(...)`` — the generic check used by
+  workday.py and servicenow.py. Discovers connections matching the
+  keyword(s), reports a summary plus one per-connection row, and
+  buckets the verdict per AGENTS.md design principle 7 (one summary
+  CheckResult plus per-resource detail rows).
+
+The first two are pure helpers and tested directly. The third is
+exercised end-to-end through ``responses``-mocked PowerApps API
+responses, mirroring the pattern in ``test_workday_connections.py``.
+
+ServiceNow-specific behavior (the connector_keyword routing for
+``_check_connections`` in servicenow.py) is covered separately by
+``test_servicenow_connections.py`` so this file stays focused on the
+shared infrastructure rather than any one consumer's plumbing.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any
+
+import pytest
+import responses
+
+from tests.conftest import require_validated_mock
+from tests.mocks import pp_admin as pp
+
+require_validated_mock(pp)
+
+
+# ───────────────────────────────────────────────────────────────────────
+# get_connection_status — pure dict navigation
+# ───────────────────────────────────────────────────────────────────────
+
+
+class TestGetConnectionStatus:
+    """Pure-unit tests for the status extractor.
+
+    Real BAP responses put the status in ``properties.statuses[0].status``.
+    The helper has to tolerate every shape the API has ever been observed
+    to return — including responses missing the entire ``properties``
+    block (which happens on connection records the caller doesn't have
+    permission to read).
+    """
+
+    def test_returns_status_from_first_entry(self) -> None:
+        conn = {"properties": {"statuses": [{"status": "Connected"}]}}
+        from flightcheck.checks.connections import get_connection_status
+        assert get_connection_status(conn) == "Connected"
+
+    def test_returns_first_entry_when_multiple_present(self) -> None:
+        # Real BAP responses sometimes return multiple status rows (e.g.
+        # one per region for global tenants). The check uses the first
+        # entry — pin that behaviour so it doesn't silently switch to
+        # "any errored" or "all connected" without a deliberate change.
+        from flightcheck.checks.connections import get_connection_status
+        conn = {
+            "properties": {
+                "statuses": [
+                    {"status": "Error"},
+                    {"status": "Connected"},
+                ]
+            }
+        }
+        assert get_connection_status(conn) == "Error"
+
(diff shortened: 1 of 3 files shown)


nehaoss (commented):
## Code Review Comments

Below are actionable findings placed at specific code locations. Each needs attention before merge.

---

### 1. `tests/flightcheck/checks/test_connections_helpers.py` line 129 - Mock shape diverges from production

The `_conn` helper uses plain strings like `shared_workdaysoap` for `apiId`, but production records use full paths (`/providers/Microsoft.PowerApps/scopes/admin/environments/{env_id}/apis/shared_workdaysoap`). The substring filter still works, but if the filter logic ever evolves to parse path structure, these tests won't catch the regression. Consider using the full path format to match the shape the real `pp_admin.connection()` builder produces.

---

### 2. `tests/flightcheck/checks/test_connections_helpers.py` lines 306-310 - Inline stubs don't enforce interface contracts

`StubClient` and `BoomClient` are ad-hoc classes. If `get_connections` ever adds required parameters or changes signature, these stubs won't break - masking interface drift. Consider using `unittest.mock.Mock(spec=PPAdminClient)` with `side_effect` for tighter coupling to the real client's contract.

---

### 3. `tests/flightcheck/checks/test_connections_helpers.py` line 361 - 403 mock body doesn't match real error shape

`list_connections(..., status=403)` returns a valid `{"value": []}` JSON body alongside an HTTP 403. In production, a 403 would return an error-shaped body (`{"error": {"code": "Forbidden", ...}}`). The test correctly pins the `_get_all` swallow behavior, but if `_get_all` is later enhanced to inspect the response body to distinguish "no connections" from "permission denied", this test won't exercise that path accurately. Consider adding a 403-appropriate error body.

---

### 4. `tests/mocks/pp_admin.py` lines 264-305 - `servicenow_connection` missing error params + name collision risk

**Missing error params**: Unlike `workday_connection`, this doesn't accept `error_target`/`error_code`/`error_message`. If SN-CONN checks ever inspect error details (like WD-CONN-101 does for AADSTS codes), you'll need to add them. Consider adding now for parity.

**Name collision risk**: The default `connection_name` truncates `display_name[:8]` - "ServiceNow HRSD" and "ServiceNow ITSM" both produce `servicen` slugs. While this doesn't break URL routing, two connections in the same test will have colliding `name` values. Consider using a longer slug or appending a counter.

---

### 5. `tests/flightcheck/checks/test_servicenow_connections.py` - Missing test for Unknown status edge case

There's no coverage for a ServiceNow connection with `Unknown` status (record missing `properties.statuses`). `get_connection_status` returns `Unknown` for malformed records, and downstream logic treats anything != `Connected` as errored. Add a test that builds a connection with a missing statuses array to pin that "Unknown" is correctly bucketed as a failure in the ServiceNow consumer - operators will hit this when BAP returns partial/permission-stripped records.

amilandi (approved):
Approving. Excellent test-only PR — pure additive coverage, zero production risk.

**What I liked:**
- Class-based grouping (`TestGetConnectionStatus`, `TestFilterConnectionsByConnector`, `TestCheckConnectorConnections{Skips,HandlesApiErrors,Bucketing,MetadataPassthrough}`) makes it obvious what broke when one fails.
- `test_http_403_silently_degrades_to_not_configured` is exemplary: pins a known wart with `pp_admin_client.py:183-184` line references, calls out `AUTH-006`/`ENV-009` as precedents, leaves a TODO so the assertion flip is intentional when the fix lands.
- Mixed-state coverage (`test_mixed_state_summary_passes_but_per_connection_flags_errored`) is the highest-value scenario — silent partial degradation is the production failure mode this whole helper exists for.
- New `servicenow_connection()` mock follows `workday_connection`/`non_workday_connection` shape and preserves `MOCK_STATUS = "validated"` provenance.

**Optional follow-ups (not blockers):**
1. The `from flightcheck.checks.connections import ...` line is repeated 20+ times — could move to module-level.
2. `_MinimalRunner` is duplicated between this file and `test_servicenow_connections.py` — candidate for `conftest.py`.

PR #142, opened 2026-06-10
Add FlightCheck Cloud Policy feedback checks (POL-FB-001/002)

GrahamMcMynn:
## What

Adds FlightCheck checkpoints **POL-FB-001** and **POL-FB-002** (category *Cloud Policies / Telemetry & Feedback*, priority HIGH) for the two Microsoft 365 Cloud Policies that gate end-user Copilot feedback for the ESS deployment:

- **POL-FB-001** — "Allow users to send feedback to Microsoft about Microsoft 365 apps"
- **POL-FB-002** — "Allow users to include screenshots and attachments when they submit feedback to Microsoft"

Feedback is the primary closed-loop quality signal (Trend Miner / IcM / product-quality). If the feedback policy is off for the deployment group, the thumbs control silently disappears — no error, no log entry — so it must be verified before launch.

## Why these are `Status.MANUAL` (not automated PASS/FAIL)

The original ask was an automated, effective-per-security-group check. I researched and verified that this is **not feasible**:

- No GA / publicly-documented Microsoft Graph endpoint exists for the Office Cloud Policy Service (OCPS) — the Cloud Policy overview documents a UI + Entra-role-based service only.
- The admin-center private backend (`config.office.com/policyadmin/...`) sits behind Azure App Gateway and **returns 403 to a service-acquired bearer** (confirmed live) — no documented service-principal/app-permission auth path.

Per FlightCheck's "no misleading results" rule, a check that authenticates only during a manual cassette capture but SKIP/ERRORs on every real run is worse than none. So these follow the existing `publishing.py` MANUAL pattern.

## Output framing

Because the operator is confirming a setting (not diagnosing a symptom):
- **`result`** explains *why* to verify (impact + silent-failure mode).
- **`remediation`** leads with **How to verify** (portal steps) + Scope+confidence + Still stuck?, and **POL-FB-001 emits the verbatim maker-facing data-sharing notice** for privacy documentation.

## Changes

- `scripts/flightcheck/checks/cloud_policy.py` — new module (POL-FB-001/002)
- `scripts/flightcheck/cli.py` — new `cloudpolicy` scope + full-scope wiring
- `tests/flightcheck/checks/test_cloud_policy.py` — 9 pure-logic tests
- `src/reference/ess-docs/flightcheck/{validation-matrix,remediation-guide}.md` — catalog rows + verbatim notice

## Acceptance-criteria notes

Satisfied: both checkpoints, category, HIGH priority; verbatim data-sharing notice in report + directive; named policies + portal deep link; role-aware directive; documented with verbatim notice in the reference.

Adapted: the per-group automated PASS/FAIL/WARN derivation is **not possible** (no reachable API), so it's delivered as MANUAL portal-verification guidance. The module documents how to promote to automated checks if the preview Graph Tenant Configuration Management API reaches GA.

## Testing

- 9 new tests; full flightcheck suite **304 passed**
- `ruff` clean
- `--scope cloudpolicy` renders both rows as Manual/High (overall READY)

### DIFF
--- a/solutions/ess-maker-skills/scripts/flightcheck/checks/cloud_policy.py
+++ b/solutions/ess-maker-skills/scripts/flightcheck/checks/cloud_policy.py
@@ -98,8 +98,8 @@ DOC_LINK = (
 # Maker-facing data-sharing notice — VERBATIM. Surfaced whenever feedback is
 # (or should be) enabled so the maker can lift it directly into their privacy
 # documentation. Acceptance criteria require this exact wording; do not
-# paraphrase. The same constant is mirrored verbatim in the FlightCheck
-# reference docs (remediation-guide.md / validation-matrix.md).
+# paraphrase. The same constant is reproduced verbatim in the FlightCheck
+# remediation guide (remediation-guide.md); validation-matrix.md references it.
 # --------------------------------------------------------------------------
 MAKER_NOTICE = (
     "End-user feedback collected from Copilot responses in this deployment "
@@ -123,11 +123,10 @@ def render_directive(
     These checkpoints ask the operator to confirm a setting, not to diagnose a
     symptom they're already seeing — so the directive leads with "How to
     verify" (the actionable steps), not a "probable cause." The remaining
-    role-aware sections (Shared Steps #7433818 / #7433819 / #7433820) follow:
-    Scope + confidence (IT-admin scope, since Cloud Policy is admin-controlled)
-    and Still stuck. When ``notice`` is supplied, the verbatim data-sharing
-    notice is appended as a final section so it travels with the directive
-    output, not just the console.
+    role-aware sections follow: Scope + confidence (IT-admin scope, since
+    Cloud Policy is admin-controlled) and Still stuck. When ``notice`` is
+    supplied, the verbatim data-sharing notice is appended as a final section
+    so it travels with the directive output, not just the console.
 
     The WHY of acting lives in the ``CheckResult.result`` field (the impact +
     silent-failure mode), per the result-vs-remediation contract.
--- a/solutions/ess-maker-skills/scripts/flightcheck/checks/cloud_policy.py
+++ b/solutions/ess-maker-skills/scripts/flightcheck/checks/cloud_policy.py
@@ -1,0 +1,269 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""
+ESS FlightCheck — Cloud Policies / Telemetry & Feedback (POL-FB-xxx)
+
+Validates the two Microsoft 365 Cloud Policies (Office Cloud Policy Service,
+managed in the Microsoft 365 Apps admin center → Policy Management) that gate
+end-user Copilot feedback for the ESS deployment:
+
+  POL-FB-001  "Allow users to send feedback to Microsoft about Microsoft 365
+              apps"            — must be Enabled for the ESS deployment group.
+  POL-FB-002  "Allow users to include screenshots and attachments when they
+              submit feedback to Microsoft"
+                               — should be Enabled for high-fidelity capture.
+
+When feedback is enabled, end users see the thumbs up/down control on Copilot
+responses; that signal is the primary closed-loop source for the FlightCheck
+Trend Miner, IcM correlation, and product-quality improvement. When the
+"Allow feedback" policy is Disabled / Not Configured for the deployment group,
+the control silently disappears — no error, no failed request, no log entry —
+so a maker can be unaware their tenant opted out of the quality signal.
+
+Why these are MANUAL (Status.MANUAL), not automated PASS/FAIL
+-------------------------------------------------------------
+The Office Cloud Policy Service (OCPS) has NO supported programmatic way for
+FlightCheck to read EFFECTIVE per-security-group feedback policy state:
+
+  * There is no GA / publicly-documented Microsoft Graph endpoint for OCPS
+    policy assignments (verified 2026-06 against MS Learn; the Cloud Policy
+    overview documents a UI + Entra-role-based service only, with no API).
+  * The admin center's private backend (``config.office.com/policyadmin/...``)
+    sits behind Azure Application Gateway and rejects a service-acquired
+    bearer (HTTP 403) without the browser's session cookies + Origin — there
+    is no documented service-principal / app-permission auth path.
+
+Per FlightCheck's design rule "no misleading results," a check that can only
+authenticate during a manual cassette capture but SKIP/ERROR on every real
+customer run is worse than no check. So these checkpoints follow the same
+``Status.MANUAL`` pattern as the publishing checklist (``checks/publishing.py``):
+the kit surfaces WHY the policy is worth verifying, HOW to verify it (portal
+steps + deep link), and the verbatim data-sharing notice, then the operator
+confirms the state in the Microsoft 365 Apps admin center. MANUAL items route
+to the "Needs manual verification" section of the report and never fail
+readiness.
+
+Output framing for a MANUAL verification checkpoint
+---------------------------------------------------
+Because the operator isn't looking at a symptom — they're being asked to
+confirm a setting — the rows do NOT lead with a "probable cause." Instead:
+
+  * ``result`` = WHY this is worth verifying (the impact + the silent failure
+    mode), so the operator understands the stakes before acting.
+  * ``remediation`` = HOW to verify it: the role-aware deployment directive
+    rendered as "How to verify" steps + "Scope + confidence" + "Still stuck?"
+    (+ the verbatim data-sharing notice on POL-FB-001).
+
+If Microsoft ships a GA programmatic OCPS policy API (e.g. the preview Graph
+Tenant Configuration Management API reaching GA), these can be promoted to
+automated checks — register the API in ``tests/fixtures/cassettes/INDEX.md``,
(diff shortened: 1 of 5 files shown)


solutions/ess-maker-skills/scripts/flightcheck/checks/cloud_policy.py
```diff
@@ -0,0 +1,269 @@
...
+    "https://learn.microsoft.com/en-us/microsoft-365-apps/admin-center/"
+    "overview-cloud-policy"
+)
+
+# --------------------------------------------------------------------------
+# Maker-facing data-sharing notice — VERBATIM. Surfaced whenever feedback is
+# (or should be) enabled so the maker can lift it directly into their privacy
+# documentation. Acceptance criteria require this exact wording; do not
+# paraphrase. The same constant is mirrored verbatim in the FlightCheck
+# reference docs (remediation-guide.md / validation-matrix.md).
```
GrahamMcMynn: This comment says the `MAKER_NOTICE` constant is "mirrored verbatim in the FlightCheck reference docs (remediation-guide.md / validation-matrix.md)", but that's only true for `remediation-guide.md` (which reproduces the full notice as a blockquote). `validation-matrix.md` does **not** contain the verbatim text — it only references it ("emits the verbatim maker-facing data-sharing notice (see the remediation guide)"). A future maintainer trusting this comment may look for, or try to keep in sync, verbatim text in `validation-matrix.md` that isn't there.

Fix: either drop `validation-matrix.md` from the parenthetical (leaving just `remediation-guide.md`), or add the verbatim blockquote to `validation-matrix.md` to make the comment accurate.

solutions/ess-maker-skills/scripts/flightcheck/checks/cloud_policy.py
```diff
@@ -0,0 +1,269 @@
...
+    scope_confidence: str,
+    still_stuck: str,
+    notice: str | None = None,
+) -> str:
+    """Render the role-aware deployment-directive block for a MANUAL check.
+
+    These checkpoints ask the operator to confirm a setting, not to diagnose a
+    symptom they're already seeing — so the directive leads with "How to
+    verify" (the actionable steps), not a "probable cause." The remaining
+    role-aware sections (Shared Steps #7433818 / #7433819 / #7433820) follow:
```
GrahamMcMynn: The `render_directive` docstring references internal work-item IDs ("Shared Steps #7433818 / #7433819 / #7433820"). These are Azure DevOps identifiers that are meaningless to anyone reading this code in the public GitHub repo, and they'll silently rot if those work items change. The surrounding prose already explains the three sections (How to verify / Scope + confidence / Still stuck) without needing the IDs.

Fix: drop the "(Shared Steps #...)" reference, or replace it with a stable description of where the shared directive structure is defined.

GrahamMcMynn (commented):
Code review by ESS Code Review Agent. Two low-severity documentation-accuracy notes on the new Cloud Policy check module; the implementation, CLI wiring, and tests otherwise look solid.

solutions/ess-maker-skills/scripts/flightcheck/checks/cloud_policy.py
GrahamMcMynn: Fixed in 308961b. Dropped alidation-matrix.md from the parenthetical and kept 
emediation-guide.md as the single verbatim source of truth (avoids duplicating the notice text in two files). The comment now reads: `remediation-guide.md`; validation-matrix.md references it.

solutions/ess-maker-skills/scripts/flightcheck/checks/cloud_policy.py
GrahamMcMynn: Fixed in 308961b. Removed the `(Shared Steps #7433818 / #7433819 / #7433820)` reference from the render_directive docstring; the surrounding prose already names the three sections (How to verify / Scope + confidence / Still stuck).

amilandi (dismissed):
Reviewed end-to-end. Solid PR — well-reasoned, honest about its constraints, and consistent with the existing MANUAL pattern. The decision to ship as `Status.MANUAL` instead of forcing an automated check through an undocumented private API is exactly the right call (verified: there''s no GA Graph endpoint for OCPS; the admin-center backend rejects service-acquired bearers at the WAF, per Microsoft Learn).

Highlights worth calling out:

- **Triple-source verbatim notice** (`MAKER_NOTICE` constant → test pin → remediation guide block) with explicit "do not paraphrase" warnings on each. This is the right way to enforce a compliance string.
- **Symmetric CLI wiring** (`SCOPE_MAP` + `FULL_SCOPE` both updated) — easy thing to forget, and you didn''t.
- **Module docstring documents the promotion path** ("If Microsoft ships a GA programmatic OCPS policy API ... register the API in cassettes/INDEX.md, add an OCPS client, replace the MANUAL bodies"). Future maintainers will thank you.
- **Doc additions span both `validation-matrix.md` and `remediation-guide.md`** with the "Why manual" explanation in both — readers landing in either doc get the full picture.

A few non-blocking suggestions; ship as-is if you disagree.

### 1. Add a negative assertion that POL-FB-002 does NOT carry the data-sharing notice

The intent is for POL-FB-001 alone to carry `MAKER_NOTICE` (POL-FB-002''s "Still stuck?" says "The data-sharing notice on POL-FB-001 already covers..."). But `test_attachments_checkpoint_explains_fidelity_why_and_how` doesn''t pin the absence. If someone later passes `notice=MAKER_NOTICE` to POL-FB-002''s `render_directive` call, the test won''t catch it. Suggest adding:

```python
def test_attachments_does_not_duplicate_data_sharing_notice():
    r = _results_by_id()["POL-FB-002"]
    assert "Data-sharing notice" not in r.remediation
    assert "will be shared with Microsoft" not in r.remediation
```

### 2. `result` framing is slightly inconsistent with the rest of the MANUAL family

`publishing.py`''s MANUAL `result` follows a terse "the kit can''t check X — confirm Y" pattern (1–2 sentences). POL-FB-001''s `result` is a multi-sentence WHY-it-matters narrative. The module docstring explicitly justifies this choice ("the operator isn''t looking at a symptom — they''re being asked to confirm a setting") and I think the framing is actually better — but it creates a visible inconsistency in the rendered report. If readers compare a POL-FB row to a QA row side-by-side, they''ll be styled differently.

Either:
- Add one short sentence up front noting the kit can''t read OCPS programmatically (matches publishing.py voice), then keep your WHY paragraph. E.g.: `"Cloud Policy state isn''t readable programmatically — confirm this in the admin center. Thumbs up / down feedback is the primary closed-loop quality signal..."`, OR
- Leave as-is and treat the docstring justification as sufficient.

I''d lean toward the one-sentence prepend; it''s a small change that makes the report style consistent without losing the WHY narrative.

### 3. Minor: `runner` parameter is unused

`check_feedback_enabled(runner)`, `check_feedback_attachments(runner)`, and `run_cloud_policy_checks(runner)` all take `runner` but never use it. ruff doesn''t flag this by default, and the signature is needed for interface conformance with `SCOPE_MAP`. Fine as-is — flagging only because some codebases use `_runner` to telegraph "unused-on-purpose." Not worth changing unless you''ve adopted that convention elsewhere.

### 4. `render_directive` is a public-ish helper

It''s exported (no underscore prefix) and the tests use it directly. If you''d rather it be module-private, rename to `_render_directive` and route tests through the public check functions instead. Probably not worth the churn — current shape is fine and the tests are tighter with direct access.

### 5. Module-level constants — `M365_APPS_ADMIN_URL` is good, but worth noting

The portal URL `https://config.office.com/` is hardcoded. If Microsoft ever redirects this (e.g. to the unified admin portal), the deep link goes stale silently — no test catches it. This is true of every URL in every check, not specific to this PR, but the comment "spelled out in remediation text rather than a fabricated deep URL" suggests you already considered the trade-off. Fine.

### Anything I''d outright block? No.

The acceptance-criteria coverage is complete, the tests pin everything that matters (verbatim notice, MANUAL status so it never fails readiness, portal link, doc link, section ordering), and the rationale for adapting "automated per-group derivation" → "MANUAL portal verification" is documented in three places. The "no misleading results" rule is honored.

Approving.

GrahamMcMynn:
Thanks for the thorough review, @amilandi! Dispositions:

**1. Negative assertion that POL-FB-002 doesn't carry the data-sharing notice** — ✅ Done in b366bc6. Added `test_attachments_does_not_duplicate_data_sharing_notice`, asserting POL-FB-002's remediation contains neither `"Data-sharing notice"` nor `"will be shared with Microsoft"`, so the verbatim notice stays exclusive to POL-FB-001.

**2. Prepend a "can't read OCPS programmatically" sentence to `result`** — Respectfully declining. An earlier review pass explicitly asked for the opposite: keep `result` as the *why-it-matters* narrative and not lead with the "can't read the policy" framing. That fact still lives in the remediation's **Scope + confidence** section, so it isn't lost — it's just not leading the `result`. Keeping the current framing per that direction.

**3. Rename unused `runner` → `_runner`** — Declining. All 11 `run_*_checks` functions take `runner` (none use `_runner`); renaming would break that convention for a cosmetic gain. Signature is required for `SCOPE_MAP` conformance.

**4. Make `render_directive` private** — Declining (per your "not worth the churn"). Tests are tighter with direct access and the current shape is fine.

**5. Hardcoded `config.office.com` URL staleness** — No action (your "Fine"); the no-fabricated-deep-link trade-off was deliberate and this applies to every check's URLs.

Thanks again!

amilandi (approved):

PR #144, opened 2026-06-11
Installer: auto-launch /setup in Copilot Chat via 'code chat'

amilandi:
## Summary

Auto-launches `/setup` in Copilot Chat at the end of the one-shot installer instead of leaving the user with "Next: in VS Code, open Copilot Chat and run /setup."

## Approach

Uses the **`code chat <prompt>`** CLI subcommand built into VS Code 1.102+ (June 2025), which opens the chat panel in the workspace at the current working directory and submits the prompt (calls `chatWidget.acceptInput()` internally — not just pre-fill). Discovered via reading `src/vs/workbench/contrib/chat/electron-browser/chat.contribution.ts` and `src/vs/platform/environment/node/argv.ts`.

`code chat` does a workspace-trust gate first; the user is prompted to trust the workspace, then `/setup` runs in chat.

### `setup/Install-EssAdk.ps1` (Windows)
- Section 7 ("Launch") now does `Push-Location $workspace; & code chat ''/setup''; Pop-Location` (wrapped in `Invoke-Native` to match the rest of the script''s pattern), instead of `Start-Process code <workspace>`.
- Default window-targeting behavior (not `--new-window`) so re-runs reuse an existing window at the same workspace.
- **Falls back** to the old "open workspace only" behavior on any non-zero exit (e.g. pre-1.102 VS Code where `code chat` doesn''t exist).
- **Messaging is honest:** success message says `Requested /setup in Copilot Chat` (not "ran"), because exit 0 only means "VS Code accepted the chat request" — the user may still need to grant trust + sign in before `/setup` actually executes. Yellow follow-up lines tell them what to expect and how to manually run `/setup` if it doesn''t auto-fire.

### `setup/install-ess-adk.sh` (macOS)
- Mirrors the Windows behavior.
- Wraps the `code chat` call in `if (cd && code chat); then` so `set -euo pipefail` doesn''t abort the script on a non-zero exit — the fallback branch handles it cleanly.

### `solutions/ess-maker-skills/.vscode/settings.json`
- Adds `"workbench.startupEditor": "none"` so the Welcome editor doesn''t compete with the Copilot Chat panel on subsequent workspace opens.
- **Caveat:** This does NOT suppress VS Code''s one-time first-run experience (telemetry/walkthrough). On a brand-new VS Code install the user still sees that gate once, closes it, and `/setup` then runs. It''s an application-level gate that ignores workspace settings.

### `setup/README.md` + script doc-comments
- Updated to reflect the new auto-`/setup` flow.

## Validation (live)

Tested on Windows 11, uninstalled VS Code, wiped all leftover user data (`%APPDATA%\Code`, `~\.vscode`, `~\.vscode-shared`), reinstalled via winget (got 1.124.0, well past 1.102), then ran:

```powershell
Push-Location <kit-workspace>
& code chat "/setup"
```

Observed:
- ✅ `code chat --help` confirms the subcommand exists with documented options.
- ✅ Exit code `0`.
- ✅ VS Code launched with workspace title `ess-maker-skills`.
- ✅ **GitHub Copilot sign-in pop-up appeared** — this is the proof that the chat handler ran end-to-end (Copilot Chat needed auth to handle `/setup`).
- ✅ On subsequent opens (with `workbench.startupEditor: "none"` applied), no Welcome tab in the editor.
- ⚠️ On the very first launch of a brand-new VS Code install, the one-time first-run walkthrough still appeared (closing it once allowed `/setup` to proceed).

PS parse-check and `bash -n` both clean.

## Trade-offs considered (and rejected)

- **`--new-window`**: would create a duplicate window when the user already has the workspace open (e.g., re-running the installer). Default behavior — reuse a window at the same workspace, else open a new one — is friendlier.
- **`code chat --help` feature-probe before invocation**: adds a round-trip; the failure path already handles unsupported VS Code versions cleanly.
- **`.vscode/tasks.json` with `runOn: folderOpen`**: tasks can only run shell commands, not VS Code internal commands like `workbench.action.chat.open`.
- **Markdown welcome file with `command:workbench.action.chat.open?...` link**: requires a click; not actually more automatic.
- **Pre-granting workspace trust via repo settings**: defeats VS Code''s security model and isn''t honored from a workspace-level config anyway.

### DIFF
--- a/setup/Install-EssAdk.ps1
+++ b/setup/Install-EssAdk.ps1
@@ -16,7 +16,8 @@
          (GitHub.copilot, GitHub.copilot-chat, ms-python.python).
       5. Clones the Employee-Self-Service-Agent-Developer-Kit repo to a known
          location (default: $env:USERPROFILE\source\Employee-Self-Service-Agent-Developer-Kit).
-      6. Opens the ess-maker-skills workspace in VS Code.
+      6. Opens the ess-maker-skills workspace in VS Code and automatically
+         requests `/setup` in Copilot Chat (requires VS Code 1.102+).
 
     The script is idempotent: re-run to repair a partial install.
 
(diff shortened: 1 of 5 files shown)


qazizaahirah:
We should add tests to this

qazizaahirah (dismissed):

amilandi:
Added in b39cf61: `tests/scripts/test_installer_launch.py` — 12 static-text regression tests covering the four behavioral guarantees of this PR:

1. Both installers invoke `code chat /setup` (not just `code <workspace>`)
2. The chat call runs from inside the kit workspace (`Push-Location ` on PS1, `(cd "" && …)` on bash) so chat opens the right workspace
3. A fallback opens VS Code + prints manual `/setup` instructions when `code chat` exits non-zero (old-VS-Code escape hatch)
4. `.vscode/settings.json` sets `workbench.startupEditor: "none"` so the Welcome tab does not sit on top of the chat prompt on subsequent opens

I went with static-text assertions over subprocess integration tests because actually running the installers requires network installs of Homebrew / Python / git / VS Code / pip deps (out of scope for the suite), and the live behavior was validated by hand on a freshly-reinstalled VS Code as documented in the PR conversation. The tests are scoped to the relevant section of each installer (a regex slice of `# 7. Launch` for PS1, `# 8. Launch …` for bash) so a coincidental string elsewhere in the file can't satisfy them.

Mutation-tested locally — flipping each guarantee (commenting out `Push-Location`, removing `chat /setup`, deleting the fallback branch, etc.) causes the corresponding test to fail. 6/6 mutations caught. All 554 tests in the suite still pass.

qazizaahirah (approved):

PR #148, opened 2026-06-12
Add per-check Role (next-step owner) to FlightCheck

GrahamMcMynn:
## Summary

Adds a **Role** to every actionable FlightCheck result — the admin persona(s) who must take the next action to fix a failure or perform the manual validation — and surfaces it as a new column in the report output.

## What changed

- **`runner.py`**: new `Role` enum (Entra Admin, Microsoft 365 Admin, Power Platform Admin, Workday Admin, ServiceNow Admin, SAP Admin, ESS Maker / Agent Developer) and a `roles: list[str]` field on `CheckResult`.
- **Per-check assignment**: `roles` set on all 229 `CheckResult` constructors across the check modules, chosen by **who performs the remediation**. Multi-role where the fix spans systems (e.g. `WD-CONN-010` / `WD-CONN-102` → Entra + Workday Admin; ServiceNow template configs → ESS Maker + Power Platform Admin).
- **Output surfaces**:
  - HTML report — new **Role** column, blanked (`—`) on Passed/Skipped rows (no next step).
  - Terminal — role shown on the ACTION REQUIRED and NEEDS MANUAL VERIFICATION rows.
  - `results.json` — `roles` included automatically.
- **Docs**: "Assigning roles" section added to `flightcheck/AGENTS.md` (incl. updated minimal example) and a Roles legend added to `validation-matrix.md`.
- **Tests**: new `tests/flightcheck/test_check_roles.py` covering the enum, the field default, status-gated rendering, JSON persistence, and terminal output.

## Design note

Roles are only meaningful for actionable results (Failed / Error / Warning / Manual / NotConfigured). Passed and Skipped rows have no next step, so the Role column is blanked for them at render time.

## Verification

- `ruff check solutions/ess-maker-skills/scripts` — clean
- `pytest tests` — **602 passed, 8 skipped** (8 pre-existing skips)

### DIFF
--- a/solutions/ess-maker-skills/scripts/flightcheck/checks/servicenow.py
+++ b/solutions/ess-maker-skills/scripts/flightcheck/checks/servicenow.py
@@ -12,7 +12,7 @@ import os
 import sys
 from pathlib import Path
 
-from ..runner import CheckResult, Status, Priority
+from ..runner import CheckResult, Priority, Role, Status
 from .connections import check_connector_connections
 from .external_systems import _categorize_servicenow_flows
 
@@ -118,7 +118,7 @@ def _check_flow_status(runner, sn_flows: list) -> list[CheckResult]:
         elif f in itsm:
             pack_label = "ITSM"
 
-        results.append(CheckResult(
+        results.append(CheckResult(roles=[Role.POWER_PLATFORM_ADMIN.value],
             checkpoint_id=cid, category="ServiceNow",
             priority=Priority.HIGH.value,
             status=Status.PASSED.value if is_on else Status.FAILED.value,
@@ -135,7 +135,7 @@ def _check_flow_status(runner, sn_flows: list) -> list[CheckResult]:
         other_detail = f"{len(other)} other" if other else ""
         breakdown = ", ".join(filter(None, [hrsd_detail, itsm_detail, other_detail]))
 
-        results.insert(0, CheckResult(
+        results.insert(0, CheckResult(roles=[Role.POWER_PLATFORM_ADMIN.value],
             checkpoint_id="SN-FLOW-000", category="ServiceNow",
             priority=Priority.HIGH.value,
             status=Status.PASSED.value if disabled == 0 else Status.WARNING.value,
@@ -154,7 +154,7 @@ def _check_template_configs(runner) -> list[CheckResult]:
     dv_token = runner.dv_token
 
     if not env_url or not dv_token:
-        results.append(CheckResult(
+        results.append(CheckResult(roles=[Role.POWER_PLATFORM_ADMIN.value],
             checkpoint_id="SN-CFG-001", category="ServiceNow",
             priority=Priority.HIGH.value, status=Status.SKIPPED.value,
             description="ServiceNow template configurations",
@@ -175,7 +175,7 @@ def _check_template_configs(runner) -> list[CheckResult]:
         )
 
         if configs:
-            results.append(CheckResult(
+            results.append(CheckResult(roles=[Role.POWER_PLATFORM_ADMIN.value],
                 checkpoint_id="SN-CFG-001", category="ServiceNow",
                 priority=Priority.HIGH.value, status=Status.PASSED.value,
                 description="ServiceNow template configurations",
@@ -188,7 +188,7 @@ def _check_template_configs(runner) -> list[CheckResult]:
             _validate_expected_configs(results, config_names, "hrsd", "SN-CFG-01")
             _validate_expected_configs(results, config_names, "itsm", "SN-CFG-02")
         else:
-            results.append(CheckResult(
+            results.append(CheckResult(roles=[Role.POWER_PLATFORM_ADMIN.value],
                 checkpoint_id="SN-CFG-001", category="ServiceNow",
                 priority=Priority.HIGH.value, status=Status.NOT_CONFIGURED.value,
                 description="ServiceNow template configurations",
@@ -201,7 +201,7 @@ def _check_template_configs(runner) -> list[CheckResult]:
             ))
 
     except Exception as e:
-        results.append(CheckResult(
+        results.append(CheckResult(roles=[Role.POWER_PLATFORM_ADMIN.value],
             checkpoint_id="SN-CFG-001", category="ServiceNow",
             priority=Priority.HIGH.value, status=Status.WARNING.value,
             description="ServiceNow template configurations",
@@ -235,6 +235,7 @@ def _validate_expected_configs(
             priority=Priority.MEDIUM.value, status=Status.PASSED.value,
             description=f"ServiceNow {pack_label} template configs",
             result=f"All {len(expected)} expected {pack_label} configs present",
+            roles=[Role.ESS_MAKER.value, Role.POWER_PLATFORM_ADMIN.value],
         ))
     elif found:
         results.append(CheckResult(
@@ -243,6 +244,7 @@ def _validate_expected_configs(
             description=f"ServiceNow {pack_label} template configs",
             result=f"{len(found)}/{len(expected)} configs found — missing: {', '.join(missing)}",
             remediation=f"Reinstall the ServiceNow {pack_label} extension pack or create missing configs manually.",
+            roles=[Role.ESS_MAKER.value, Role.POWER_PLATFORM_ADMIN.value],
         ))
     # If none found, the pack likely isn't installed — don't flag as error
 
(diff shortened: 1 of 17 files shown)


GrahamMcMynn (commented):

solutions/ess-maker-skills/scripts/flightcheck/checks/servicenow.py
```diff
@@ -188,7 +188,7 @@ def _check_template_configs(runner) -> list[CheckResult]:
             _validate_expected_configs(results, config_names, "hrsd", "SN-CFG-01")
             _validate_expected_configs(results, config_names, "itsm", "SN-CFG-02")
         else:
-            results.append(CheckResult(
+            results.append(CheckResult(roles=[Role.POWER_PLATFORM_ADMIN.value],
```
GrahamMcMynn: SN-CFG-001's NotConfigured result is tagged `roles=[Role.POWER_PLATFORM_ADMIN.value]` only, but its remediation ("Install the ServiceNow extension pack (HRSD/ITSM) in Copilot Studio") is a maker action. The sibling SN-CFG-01/02 results in `_validate_expected_configs` — same "(re)install the pack" concept — are tagged `[Role.ESS_MAKER.value, Role.POWER_PLATFORM_ADMIN.value]`.

`AGENTS.md` says to pick roles from "WHO performs the fix described in `remediation`, not from the category alone," which suggests SN-CFG-001 (and arguably the WARNING branch below) should also include `Role.ESS_MAKER`.

GrahamMcMynn: Good catch — fixed in 660a2c7. All four SN-CFG-001 branches are now tagged `[ESS Maker / Agent Developer, Power Platform Admin]`, matching the sibling SN-CFG-01/02 results, since the remediation (install the extension pack in Copilot Studio) is a maker action.

amilandi (commented):
Reviewed locally — checked out the branch, ran `pytest tests` (602 passed, 8 pre-existing skips) and `ruff check` (clean), AST-scanned the checks tree and confirmed **all 229 `CheckResult(...)` constructors carry `roles=`**. Both the design and the execution are solid. A few non-blocking observations below.

## Strengths

- `Role` enum is clean, well-documented, and the seven values map cleanly onto the actual admin personas the kit&apos;s remediations point at.
- `CheckResult.roles: list[str] = field(default_factory=list)` is backwards-compat — unit tests that construct minimal CheckResults won&apos;t break, and the runner&apos;s ERROR-fallback path still builds.
- The HTML renderer correctly blanks the Role cell for Passed/Skipped via `actionable = res.status not in (Status.PASSED.value, Status.SKIPPED.value)` — matches the docs&apos; &quot;a row that only ever Passes or is Skipped has no next step&quot; rule, and the test pins that even a *populated* `roles` list gets blanked on Passed/Skipped so a conditional-status check (`Status.PASSED.value if connected else Status.FAILED.value`) renders correctly on either branch.
- Multi-role assignment is used where the fix genuinely spans systems: `WD-CONN-010 / -102 → [ENTRA_ADMIN, WORKDAY_ADMIN]` for SAML cert validation (cert lives on the Entra app, compared in Workday), `SN-CFG-* → [ESS_MAKER, POWER_PLATFORM_ADMIN]` for template configs (maker authors, PP Admin imports the solution). These match the AGENTS.md guidance.
- AGENTS.md and `validation-matrix.md` both updated with the new convention — `Role` is now part of the documented contract, not just an implementation detail.
- Test coverage hits the contract points: enum membership, default-empty, HTML header+cell, status-gated rendering on every actionable status, JSON persistence, terminal output, and one production check module end-to-end.

## Non-blocking suggestions

**1. Coverage hole: `test_publishing_checks_all_carry_roles` is the only enforcement that real check modules actually set `roles`.** I verified via AST scan that all 229 constructors across 12 check modules currently set `roles=`, so the PR is complete — but the suite only enforces it for `publishing.py`. A future PR that adds a check without `roles=` would land green except in publishing.py. One-line fix that catches all 12 modules:

```python
def test_all_check_modules_carry_roles_on_every_constructor():
    """AST-scan checks/*.py — every CheckResult(...) call must pass roles=."""
    import ast
    from pathlib import Path
    checks_dir = (
        Path(__file__).resolve().parents[1]
        / &quot;solutions&quot; / &quot;ess-maker-skills&quot; / &quot;scripts&quot;
        / &quot;flightcheck&quot; / &quot;checks&quot;
    )
    missing = []
    for f in sorted(checks_dir.glob(&quot;*.py&quot;)):
        if f.name == &quot;__init__.py&quot;:
            continue
        tree = ast.parse(f.read_text(encoding=&quot;utf-8&quot;))
        for node in ast.walk(tree):
            if (isinstance(node, ast.Call)
                and isinstance(node.func, ast.Name)
                and node.func.id == &quot;CheckResult&quot;
                and not any(k.arg == &quot;roles&quot; for k in node.keywords)):
                missing.append(f&quot;{f.name}:{node.lineno}&quot;)
    assert not missing, f&quot;CheckResult() missing roles=: {missing}&quot;
```

This is the structural enforcement that backs the &quot;every actionable result has a role&quot; promise repo-wide. Without it the AGENTS.md rule is only conventionally enforced.

**2. SKIPPED branches still assign roles even though the renderer blanks them.** Examples: `connections.py:107`, `workday.py:765, 785` (WD-PKG-001 SKIPPED). Harmless (the renderer blanks Role on SKIPPED anyway), but it&apos;s wasted bytes and the AGENTS.md guidance suggests roles isn&apos;t needed there. Two ways to reconcile: (a) drop `roles=` from the pure-SKIPPED branches, or (b) add a sentence to AGENTS.md saying &quot;we set `roles` even on SKIPPED rows for uniformity; the renderer hides it.&quot; Either works; the current state is just slightly inconsistent with the doc.

**3. WD-ENV-101 role attribution may be too narrow.** The whole check is gated on `runner._workday_package_flavor == "simplified"` (skip), and the `"@" not in isu_value` warning branch (line 1342) directs the fix to `EmployeeContextRequestAccountName` in Power Platform admin center → so POWER_PLATFORM_ADMIN is right per the &quot;follow the remediation&quot; rule. But the remediation text *also* says &quot;If the tenant federates identity, set the Workday ISU username to the Entra UPN format&quot; — that side of the fix happens in the Workday tenant, so a strict reading of multi-role suggests `[POWER_PLATFORM_ADMIN, WORKDAY_ADMIN]`. Same pattern at the &quot;domain not in verified&quot; branch (line 1436). Defensible either way; flagging for the author&apos;s call.

**4. Terminal output format `[Critical | ESS Maker / Agent Developer, Power Platform Admin]` is long.** On the multi-role PUB-002 / PUB-006 / SN-CFG-* rows this approaches 80 chars before the result text even starts. Optional polish: drop the priority+role bracket onto its own line, or abbreviate roles (e.g. `ESS Maker` instead of `ESS Maker / Agent Developer`). Not blocking — just something the operator will notice on a wide-terminal report.

## CI

Lint Python ✅, CodeQL ✅, CLA ✅, mergeable. Local `pytest tests`: 602 passed / 8 pre-existing skipped.

LGTM modulo the test enforcement (#1) — would recommend adding before merge so the contract is structurally pinned. #2–4 are polish that can land in a follow-up.

amilandi (dismissed):

GrahamMcMynn:
Thanks for the thorough review @amilandi — addressed in fb38554:

- **#1 (test enforcement):** Added `test_all_check_modules_carry_roles_on_every_constructor` — AST-scans every `checks/*.py` and fails if any `CheckResult(...)` omits `roles=`. The convention is now structurally pinned repo-wide, not just in `publishing.py`.
- **#3 (WD-ENV-101):** Retagged all branches `[Power Platform Admin, Workday Admin]` — the remediation spans the Dataverse env var (Power Platform) and the Workday ISU username format (Workday tenant), matching the multi-role pattern used by WD-CONN-010/-102.
- **#2 (SKIPPED rows):** Went with the documentation option — added a note to AGENTS.md that `roles` is set on every constructor (including SKIPPED) for uniformity and is enforced by the new test, while the renderer blanks the Role cell on Passed/Skipped. This keeps it consistent with #1 (dropping `roles=` from SKIPPED branches would fail the new enforcement test).
- **#4 (terminal width):** Skipping for now — abbreviating roles in the terminal would desync the labels from the HTML/JSON output. Happy to revisit as a follow-up if it's bothersome in practice.

`pytest tests` 430 passed (flightcheck) and `ruff check` clean after the changes.

amilandi (approved):

PR #157, opened 2026-06-22
Add WD-RUN-001: Workday flow run-health check

GrahamMcMynn:
## What

Adds **`WD-RUN-001` — Workday flow run-health**, a FlightCheck check that reads the Workday shared-flow **run history** to detect runtime failures that the existing connection-status checks (`WD-CONN-001+`) cannot see.

## Why

A Power Platform connection can show **Connected** while Workday calls still fail at runtime (revoked Workday security domain/permission, a misconfigured scenario template config, expired token, Workday-side outage). Connection status can't see that — the only evidence is in the shared flow's run history.

Confirmed live against a real ESS Workday tenant: **the ESS flow catches Workday faults and responds gracefully, so a failed Workday call leaves the run `status=Succeeded`.** The real signal is the flow's **Response action name** (`Respond_to_Copilot_with_Success` vs failure branches like `Respond_to_Copilot_with_failure_errorMessage`). Some failures (template/transform) also surface as `status=Failed`. The check uses both signals.

## Behaviour

- **Deterministic-break litmus test:** evaluates only the most recent window of runs and **FAILs only when none of them succeeded**. A couple of scattered failures among recent successes do **not** fail readiness (the integration is demonstrably working); a single failed run with no successes **does** fail.
- **Suppresses redundant manual checks when healthy:** when `WD-RUN-001` passes, the MANUAL Workday connection/security rows (`WD-CONN-010` federation, `WD-CONN-102` SAML cert, `WD-SEC-003` personal-data write) are hidden — runtime traffic already proves that chain works. They stay visible when it fails or can't confirm health (no traffic / unavailable), e.g. a fresh pre-deployment env.

### Known limitations (documented in code)
- A fully-broken/unconfigured connection makes Copilot Studio prompt "connect to continue" and never invokes the flow, so **no run is created** — covered by `WD-CONN-001+` (connection status), not here.
- Agent-side timeouts (`flowActionTimedOut`) leave the run `Succeeded` and aren't detectable from run history.

## Changes

- `pp_admin_client.py`: `get_flow_runs()` (runtime runs endpoint, reuses the existing flow-token auth).
- `checks/workday.py`: `_check_workday_run_health` (WD-RUN-001), `_classify_run`, `_suppress_manual_conn_sec_when_runs_healthy`; registered in `run_workday_checks`.
- `tests/mocks/pp_admin.py`: `flow_run` / `list_flow_runs` builders (+ a `flow_runs` 403 branch).
- `tests/fixtures/cassettes/INDEX.md`: API tier registry + confirmed-endpoint rows.
- `tests/flightcheck/checks/test_workday_run_health.py`: 12 tests (all-success, recent-successes-with-failures, deterministic break, single failed run, no-traffic, running-ignored, 403, no-flows, no-pp-admin, and the manual-suppression matrix).
- `tests/captures/record_flightcheck_workday_runs.py` + `tests/fixtures/cassettes/flightcheck_workday_runs.yaml` (captured from a real tenant — includes success and both failure signatures — scrubbed of SAS tokens / env id).

## Testing

- 12 new tests pass · **473/473** FlightCheck tests pass · `ruff check` clean.
- Cassette captured all three real run signatures (success, caught Workday fault, template error) and verified clean (0 SAS tokens, env id redacted, valid YAML).


solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py
```diff
@@ -2686,6 +2720,223 @@ def _check_flow_status(runner, wd_flows: list) -> list[CheckResult]:
...
+    props = run.get("properties", {}) or {}
+    status = props.get("status")
+    resp_name = ((props.get("response") or {}).get("name")) or ""
+    if status in (None, "Running", "Waiting", "Paused"):
+        return "pending"
+    if status == "Failed":
+        return "hard_failure"
+    if status == "Succeeded" and resp_name and resp_name != _WD_SUCCESS_RESPONSE_ACTION:
+        return "caught_failure"
+    return "success"
```
GrahamMcMynn: `_classify_run` treats any terminal status that isn't `"Failed"` (or a `"Succeeded"` run on a failure response branch) as `"success"` via the final `return "success"`. But `get_flow_runs`'s docstring lists `"Cancelled"` as a possible status (and the runs API can also emit `Aborted`/`TimedOut`/`Skipped`) — all of which fall through to `"success"` here.

**Impact:** if the most recent window contains only `Cancelled` runs, `_check_workday_run_health` counts them as successes, so WD-RUN-001 PASSES and reports "All N most recent Workday flow run(s) succeeded", and `_suppress_manual_conn_sec_when_runs_healthy` then hides WD-CONN-010, WD-CONN-102 and WD-SEC-003 — a misleading green from runs that never actually succeeded. There is no test covering a `Cancelled` run.

**Suggested fix:** gate the success branch on the status explicitly, e.g.:

```python
if status == "Succeeded":
    if resp_name and resp_name != _WD_SUCCESS_RESPONSE_ACTION:
        return "caught_failure"
    return "success"
return "hard_failure"  # Cancelled/Aborted/TimedOut/etc. are not success
```

(or classify non-`Succeeded` terminal states as a distinct non-scoring kind if you'd rather not count a user-cancelled run as a failure). Worth adding a test for `status="Cancelled"`.

GrahamMcMynn (commented):

qazizaahirah (approved):

PR #158, opened 2026-06-23
Add WD-REF-001: Workday write-scenario reference-data availability

GrahamMcMynn:
## What

Adds **`WD-REF-001` — Workday write-scenario reference-data availability**: verifies that every reference-data picklist a Workday topic requests from the shared `GetReferenceData` system topic is actually supported, so the agent can validate user input for Add/Update scenarios (phone, email, emergency contact, dependent, address, government IDs, …).

## Why

Many Workday write scenarios validate user-supplied free text against Workday reference data (ISO country codes, phone device types, relationship types, government-ID types, …). If the reference set for a field can't be loaded, the agent **rejects valid inputs, accepts invalid ones (causing a downstream Workday SOAP fault), or hallucinates allowed values**.

How ESS loads reference data (confirmed on a live tenant): a write topic lazily calls the shared `GetReferenceData` topic with a `referenceDataKey`, which loads the picklist at runtime (via the `msdyn_HRWorkdayHCMEmployeeGetReferenceData` read scenario) for a **fixed set of supported keys**. A topic that requests a key GetReferenceData doesn't support gets an empty picklist.

## How it works

Pure **Dataverse** reconciliation (`documented` tier — no external API, **no cassette**): reads all topic `botcomponents.data` via `query_all` and checks that every reference key a topic **requests** (`referenceDataKey: KEY`) is in `GetReferenceData`'s **supported** switch (`referenceDataKey = "KEY"`).

- **PASS** — every requesting topic requests only supported keys.
- **FAIL** — a topic requests an unsupported key (names the topic + key), or `GetReferenceData` is missing entirely.
- **NotConfigured** — no topic requests a reference picklist.
- Failure remediation includes a **clickable Copilot Studio deep-link** (reusing the kit's verified `_studio_link_md` helper) to the agent → Topics, where the `GetReferenceData` call is edited.

### Key design point (investigated against a live tenant)

The check reconciles the topics' **actual `GetReferenceData` requests**, NOT the write request-template ID types. Those do **not** map 1:1 to GetReferenceData keys — e.g. a topic loads countries via `ISO_3166-1_Alpha-2_Code` while the request template pulls the alpha-3 / region value from that same table, and some fields (gender) are static. Reconciling request-template types would **false-positive on OOTB scenarios** (AddEmergencyContact, AddDependent). The request-based reconciliation is correct and false-positive-free (validated: phone, dependent topics → PASS).

## Testing

- 8 new tests (extractors + pass / unsupported-key gap / GetReferenceData-missing / no-requests / no-token / query-error, incl. the remediation deep-link). **469/469** FlightCheck tests pass · `ruff` clean.
- Validated live: PASS on the OOTB agent; and FAIL correctly surfaced (with the clickable deep-link) when a topic was temporarily pointed at an unsupported key.

## Notes

- This catches *"a topic requests an unsupported picklist"* (the high-confidence, common gap). It does **not** attempt the far-less-reliable *"a scenario should present a picklist but the topic never loads one"* (which needs fragile semantic set-mapping) — that could be a follow-up.


solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py
```diff
@@ -322,6 +322,240 @@
...
+
+def _wd_studio_link(runner) -> str:
+    """Markdown deep-link to the active agent in Copilot Studio (its **Topics**
+    page is where the GetReferenceData call is edited). Falls back to the
+    Copilot Studio homepage when the env/bot id can't be resolved. Reuses the
+    verified URL helper from local_files."""
+    try:
+        from .local_files import _studio_link_md
+        config = getattr(runner, "config", {}) or {}
+        slug = config.get("activeAgent") or (config.get("agent") or {}).get("slug") or ""
```
GrahamMcMynn: **[bug]** `_wd_studio_link` reads `config.get("activeAgent")`, but `activeAgent` is set nowhere in the kit (0 occurrences outside this function). The rest of FlightCheck identifies the active agent by folder name matched against `config["agents"][].slug` (see `local_files.py` -> `agent_path.name`). Since `activeAgent` is always `None`, `slug` falls back to `config["agent"].get("slug")` (the single-agent key). In a multi-agent config that key is absent, so `slug == ""` -> `_studio_agent_url` returns `None` -> the remediation silently links to the generic Copilot Studio homepage instead of the agent -> Topics deep-link this PR advertises.

Consider resolving the slug from `config["agents"]` (the active agent), consistent with how the rest of the kit derives it.

GrahamMcMynn (commented):
Code review by ESS Code Review Agent. 2 comments on the new Copilot Studio deep-link helper (`_wd_studio_link`).

tests/flightcheck/checks/test_workday_reference_data.py
```diff
@@ -0,0 +1,156 @@
...
+    r = _run()
+    assert r.status == "Failed"
+    assert "1 of 2 Workday topic(s)" in r.result
+    assert "Workday Custom Marital Status" in r.result
+    assert "Marital_Status_ID" in r.result
+    # The other (valid) topic must NOT be reported as a gap.
+    assert "Update PhoneNumber" not in r.result
+    assert "GetReferenceData" in r.remediation
+    # A clickable fix-link to Copilot Studio is present in the remediation.
+    assert "copilotstudio.microsoft.com" in r.remediation
```
GrahamMcMynn, line 117: **[testing]** This is the only test that touches the Studio deep-link, but `_runner()` exposes no `env_id` or `config`, so `_wd_studio_link` only ever exercises the homepage fallback (`_studio_agent_url` returns `None` because both `env_id` and the slug are falsy). The assertion `"copilotstudio.microsoft.com" in r.remediation` passes for **both** the real deep-link and the fallback, so it can't catch a broken deep-link (e.g. the `activeAgent` slug resolution issue above).

Add a case with a runner exposing `env_id` and `config={'agents': [{'slug': ..., 'botId': ...}]}` and assert the `/environments/{env}/bots/{bot}/overview` deep-link shape appears in the remediation.

solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py
GrahamMcMynn: Thanks for flagging this. After investigating, I believe the premise here is incorrect: `activeAgent` **is** written and is the canonical key.

- `setup.py` writes it: `config["activeAgent"] = slug` (setup.py:549).
- `cli.py` reads it with the exact same precedence chain this helper mirrors (cli.py:160): `config.get("activeAgent") or (config.get("agent") or {}).get("slug")`.
- Live config on the test tenant has `activeAgent`, `agent.slug` and `agents[].slug` all populated (= `esshrwdayonlyoauth`), and a live FlightCheck run rendered the resolved agent overview URL — not the homepage fallback.

So the silent-homepage regression doesn't occur in practice. That said, your underlying point about robustness is fair, so in 939dc41 I added a defensive fallback to `config["agents"][0].slug` (after `activeAgent` / `agent.slug`) and added two tests that assert the remediation contains the resolved `/environments/{env}/bots/{bot}/overview` URL — including a case where `activeAgent` is absent and the link resolves purely from `agents[]`.

tests/flightcheck/checks/test_workday_reference_data.py
GrahamMcMynn: Good catch — fixed in 939dc41. Added `test_failure_remediation_contains_resolved_studio_deeplink` (runner with `env_id` + `agents[]` and `activeAgent`) and `test_deeplink_resolves_from_agents_when_activeagent_absent` (no `activeAgent`, link resolved from `agents[]` only). Both assert the remediation contains the resolved `/environments/ENV-123/bots/BOT-456/overview` URL, so the deep-link shape is now pinned rather than only the homepage fallback.

qazizaahirah (approved):

PR #165, opened 2026-06-24
Add Aria/1DS telemetry for ESS ADK & FlightCheck

amilandi:
## Summary

Adds Aria/1DS telemetry for the **ESS ADK** so leadership dashboards can track
adoption (active installs, sessions, capability usage), performance, and
**FlightCheck** outcomes (success rate, per-checkpoint pass rates). Two event
families ship together:

- **FlightCheck outcomes** — emitted from the Python FlightCheck runner, where
  the pass/fail verdict is actually produced (the VS Code extension only knows a
  button was clicked).
- **`adk.*` platform events** — session/build/deploy/api/capability/flightcheck
  events emitted across the ADK CLI skills.

Powers the **"ESS ADK & FlightCheck"** Aria dashboards (dev + prod).

## Dashboards

- **Dev**: https://portal.aria.ms/dashboard/gJ3tJywEt63-COOXssbCQ-6us0HhEcNhZ5-6PfS03EJesZkuANYXPHs?projectId=08e397b2c6c243eeaeb341e111c36167&isEU=false
- **Prod**: https://portal.aria.ms/dashboard/gJ3tJywEt63-MRJUJXu8QX6GDHZ4HUhjyAUUbnK5v0MHk2vIoQqvUyo?projectId=311254257bbc417e860c76781d4863c8&isEU=false

### FlightCheck outcome telemetry
- **`scripts/flightcheck/telemetry.py`** (new): dependency-light NDJSON POST to the
  verified 1DS OneCollector contract.
  - Config-driven **dev/prod iKey split** (default **prod**) via
    `ESS_FLIGHTCHECK_ARIA_ENV` / `ESS_FLIGHTCHECK_ARIA_IKEY`; disable with
    `ESS_FLIGHTCHECK_TELEMETRY=off`. Every event carries an `env` dimension as a backstop.
  - **Fail-open**: never raises, never changes the FlightCheck exit code; short timeout.
  - Persisted client `instanceId` (`.local/.instance_id`, not user-linked); `--selftest` entry point.
- **`flightcheck/cli.py`**: emit after `save_results`; `--no-telemetry` opt-out. Also emits the
  additive `adk.flightcheck.*` events.

### ADK `adk.*` event family
- **`scripts/adk_telemetry.py`** (new): shared emitter reusing the FlightCheck OneCollector
  transport. Event catalog: `adk.session.start/end`, `adk.agent.create/deploy`,
  `adk.build.start/complete`, `adk.api.call`, `adk.capability.use`,
  `adk.flightcheck.run/result/error`. Session windowing (30-min), on-failure NDJSON buffering,
  one-time consent notice, `on/off` opt-out, and a `synth` seeder for populating dashboards.
- **`invocation_source`** dimension to distinguish ADK / installer / CLI runs.
- Wired into the CLI skills: `auth.py`, `discover.py`, `push.py`, `setup.py`,
  `evaluate_evals.py`, `list_environments.py` (best-effort; telemetry never breaks a skill).
- Docs/config: `SKILL.md`, onboarding `step3-flightcheck.md`, `.vscode/mcp.json`, setup scripts.

### Privacy model (aligned to the approved Data Profile)
- **No developer/user identity is collected — not even hashed.** Active-install / DAU-WAU-MAU
  counts dedupe on **`instance_id`**, a random per-install GUID that is not linkable to any AAD
  user. (Earlier developer-id hashing was removed entirely.)
- **`tenant_id` is emitted RAW** — the approved Data Scout Data Profile (privacy review
  **COMPLETED**) classifies it Organizational Identifiable Information (OII) with
  **"No Data Transformation"** (it identifies the enterprise tenant, not an individual),
  retained <= 30 days.
- Enums + identifiers only; error fields scrubbed of paths/URLs and truncated. Deliberately
  **no `result` / `remediation` free text**. No user content is ever emitted.

### Consent, opt-out & docs (spec Consent & Notice)
- Telemetry is **enabled by default** with a one-time notice on first CLI use.
- Opt-out via `python scripts/adk_telemetry.py off` / `on` / `status` (preference in
  `~/.adk/config`, delete to reset), or `ESS_ADK_TELEMETRY=off`. The notice now points to
  this real invocation (there is no `adk` binary in this repo).
- Added a **Telemetry & Privacy** section to `solutions/ess-maker-skills/README.md` (what is/
  isn't collected, opt-out, storage, 30-day retention, https://aka.ms/adk-telemetry) plus a
  pointer in the root README, per the spec's README requirement.

## Tests
- **`tests/flightcheck/test_telemetry.py`** (new): envelope/iKey mapping, NDJSON shape, no
  free-text leak, env/key resolution, instance_id persistence, fail-open, raw-tenant emission.
- **`tests/flightcheck/test_adk_telemetry.py`** (new): identity (`instance_id` dedup, no
  developer id), common-dimensions shape, event envelopes, session windowing, buffering.
- All pass (33 across both files); existing `test_cli.py` still passes (no regression).

## Validation
- `python scripts/flightcheck/telemetry.py --selftest` -> **HTTP 204 (accepted)** by the dev collector.
- Dashboards verified live: the "ADK: Active Users" tiles now dedupe on the `instance Id` cube
  measure; RTA returns 220 distinct instances in dev, filterable by raw tenant GUID.

## Notes
- Emits to **prod** by default (real maker runs). Set `ESS_FLIGHTCHECK_ARIA_ENV=dev`
  `ESS_ADK_ARIA_ENV=dev` for local testing / dashboard seeding so those runs don't pollute prod.
- A single opt-out (`python scripts/adk_telemetry.py off`, or `ESS_ADK_TELEMETRY=off`) covers **both** the `adk.*` and legacy `ESSMakerKit.FlightCheck.*` event families.
- iKeys are 1DS write-only ingestion keys (embedded in clients by design).

ADO: https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7519625


apurvabanka:
@amilandi Please review these PR comments.

🔴 High — functional / data-quality bugs
1. Async capability.use events are silently dropped in 3 skills.
discover.py, evaluate_evals.py, and list_environments.py call emit_capability_use(...), which dispatches on a daemon thread (_emit → threading.Thread(daemon=True)). None of these three call flush() before returning, and they're short-lived CLI processes — so the interpreter exits and kills the daemon thread mid-POST. The event never reaches the collector and isn't buffered (buffering only happens inside _emit_sync after a completed-but-failed POST). Net effect: the "onboarding/evaluations capability usage" metric these were added for will be badly undercounted. push.py, setup.py, and cli.py do flush; these three should too (or emit with block=True).

2. Committed .local/.instance_id with a real GUID.
solutions/ess-maker-skills/scripts/.local/.instance_id = 6f7c8f9c-… is checked in. It sits under scripts/.local/, which is not covered by the solutions/ess-maker-skills/.gitignore .local/** rule (that rule is relative to that dir, i.e. …/ess-maker-skills/.local/). This looks like leftover from the author running --selftest from scripts/. Two problems: (a) it ships a canned instance_id that pollutes/undercounts the very active-install dedup metric the PR is built around, and (b) it's committed state that shouldn't be. Delete it and add scripts/.local/ to .gitignore.

3. .vscode/mcp.json force-committed with a hardcoded tenant org URL.
The root .gitignore line 40 ignores .vscode/mcp.json (matches at any depth), yet this file was force-added and hardcodes https://orgc43de63f.crm.dynamics.com/api/mcp — the author's specific Dataverse org. This leaks a dev org endpoint and re-commits an intentionally-ignored file. It's also unrelated to telemetry — looks accidental. Should be dropped from the PR.

🟠 Medium — privacy & correctness
4. error_message scrubbing doesn't cover emails / UPNs / GUIDs.
_scrub() strips paths and URLs but not email addresses, user principal names, or GUIDs. In auth.py, error_message=str(error) comes straight from Dataverse exceptions, which routinely contain a UPN or user@tenant. Given the PR's strong "no developer/user identity — not even hashed" claim, a raw UPN slipping through in error_message contradicts the stated Data Profile. Either drop error_message free text entirely (as FlightCheck deliberately does) or extend _scrub to redact emails/GUIDs.

5. Prod ingestion key is now public + prod-by-default.
The prod iKey is embedded in a public repo (acknowledged as write-only by design), and DEFAULT_ENV is now prod. Combined, anyone can POST arbitrary ESSMakerKit.*/adk.* envelopes to the production dashboards, and every maker running from the public sample emits to prod by default. That's an accepted 1DS pattern, but worth an explicit callout: the "success rate" dashboards are spoofable, and there's no shared secret gating them. Consider server-side filtering on a signed dimension if the numbers feed leadership decisions.

6. query_all doesn't emit on error statuses.
In auth.py, query_all emits api.call only on 401 and on final 200. For 403/500 it falls through to raise_api_error(...) and raises without emitting — so failed reads are missing from the api-error metrics, while create/update/delete/get do emit their real status. Inconsistent coverage skews the outcome distribution.

🟡 Low — robustness / style
7. cli.py cross-block variable coupling. The adk.* block reuses active_agent, which is assigned inside the previous try. If that first try raises before the assignment, the second block hits NameError (caught, but the adk events are lost for a non-telemetry reason). Compute active_agent/_agent_id once, before both blocks.

8. get_session read-modify-write is unlocked (unlike the buffer, which got a lock). Concurrent skills could clobber session.json. Low risk for sequential CLI use, but inconsistent with the care taken elsewhere.

9. flush() joins each thread with the full timeout, so worst case is timeout × live-threads at exit. Usually 1–2 threads, but on a busy push it could add noticeable exit latency. A single deadline across all joins would be safer.

10. Test placement. tests/flightcheck/test_adk_telemetry.py lives under flightcheck/ though adk_telemetry is the general SDK, not FlightCheck. Minor, but future readers will look in the wrong place.

amilandi:
Thanks @apurvabanka — addressed in `1a10f13`. Summary per comment:

**🔴 High**
1. **Async `capability.use` dropped in 3 skills** — Fixed. `discover.py`, `evaluate_evals.py`, and `list_environments.py` now call `emit_capability_use(..., block=True)`, so the event is emitted synchronously on the calling thread instead of a daemon thread the interpreter kills on exit.
2. **Committed `.local/.instance_id`** — Removed from the repo and added `scripts/.local/` to `solutions/ess-maker-skills/.gitignore` (the existing `.local/**` rule is relative to that dir and didn't cover `scripts/.local/`). It was leftover `--selftest` state.
3. **Force-committed `.vscode/mcp.json` with a hardcoded org URL** — Removed. It was accidental, unrelated to telemetry, and re-committed an intentionally-ignored file.

**🟠 Medium**
4. **`_scrub` didn't cover emails/UPNs/GUIDs** — Fixed. `_scrub()` now also redacts email/UPN patterns and GUIDs, so a UPN or object id echoed by a Dataverse exception can't leak into `error_message`. Added a test (`test_scrub_redacts_emails_upns_and_guids`).
5. **Prod iKey public + prod-by-default (spoofable dashboards)** — Acknowledged; this is the accepted 1DS write-only-key pattern and there's no shared secret gating ingestion. Every event already carries an `env` dimension, and instance/tenant are raw dimensions, so we can add server-side/dashboard-side filtering (e.g. drop envelopes with unexpected `adk_version`/`surface` combinations) if the success-rate numbers start feeding leadership decisions. Flagging as a known limitation rather than a code change in this PR — happy to open a follow-up if you'd prefer a signed dimension.
6. **`query_all` didn't emit on error statuses** — Fixed for the reachable case: it now emits `adk.api.call` with the real status for any `>= 400` before `raise_api_error`. Note 5xx (429/500/502/503/504) is retried by the shared `_SESSION` and surfaces as `RetryError` before reaching this path; 403 now emits. Added `test_emits_api_call_on_client_error_before_raising`.

**🟡 Low**
7. **`cli.py` `active_agent` cross-block coupling** — Fixed. `active_agent` is now computed once before both telemetry blocks, so the `adk.*` block can't hit `NameError` if the first block raises early.
8. **Unlocked `get_session` read-modify-write** — Fixed. Added `_SESSION_LOCK` around the `~/.adk/session.json` read-modify-write, matching the buffer lock.
9. **`flush()` full-timeout-per-thread** — Fixed. `flush()` now uses a single overall deadline across all joins, so worst-case exit latency is bounded by `timeout` regardless of live-thread count.
10. **Test placement** — Moved `tests/flightcheck/test_adk_telemetry.py` → `tests/test_adk_telemetry.py` (it's the general SDK, not FlightCheck-specific). The autouse isolation fixture is self-contained in the file and `pythonpath` is global, so it runs unchanged.

Full suite green (688 passed, unrelated pre-existing `test_installer_*` failures excepted).

daeunJe0ng:
Took a pass through this, looks solid.
Design is careful: fail-open everywhere, consent notice is on the auth path, and it only emits IDs and enums, not the result/remediation text. 
Earlier review points look handled too.

Couple small things, neither blocking:

A. The docstrings in  flightcheck/telemetry.py  ( _consent_disabled ,  resolve_ikey ) still say  adk telemetry off , but that command doesn't exist. The actual one is  python scripts/adk_telemetry.py off , which is what the notice text already prints. Just needs the docstrings updated to match.

B. Every run prints  [telemetry] env=prod sent=... status=... . That's internal detail makers shouldn't really see. Could hide it behind a debug/verbose flag, or trim it to one plain line.

One thing I do want tracked before merge: the prod ingestion key is public and we default to prod, so anyone can push events to the dashboards. 
I know that's the standard write-only-key setup, but since these numbers go up to leadership, can we open a follow-up for some server-side filtering? 
The env dimension by itself doesn't stop spoofing.

A and B are optional cleanups.

apurvabanka (dismissed):

amilandi:
Thanks @daeunJe0ng — all three addressed in `8bd799c`:

**A. Stale `adk telemetry off` docstrings** — Fixed. Updated the `_consent_disabled` and `resolve_ikey` docstrings in `flightcheck/telemetry.py` to reference the real command `python scripts/adk_telemetry.py off` (matching the notice text and README).

**B. Noisy `[telemetry] ...` output** — Fixed. Those lines are internal detail makers shouldn't see, so all telemetry status prints (success, skipped, and the `--no-telemetry` line) are now gated behind `ESS_FLIGHTCHECK_TELEMETRY_DEBUG`. Normal runs are silent; set that env var when debugging emission.

**C. Public prod key spoofing (tracked before merge)** — Opened #179 to track server-side/dashboard-side filtering (signed dimension / implausible-combo drops / gateway with shared secret) so the leadership-facing numbers are resilient to spoofed ingestion. Agreed the `env` dimension alone doesn't stop it; it's a known limitation of the write-only-key model.

apurvabanka (approved):

PR #166, opened 2026-06-24
Fix /setup environment selection (render Markdown table instead of scraping terminal)

amilandi:
## Problem

Running `/setup` and choosing "Yes, list my environments" frequently shows **no environment table** (or a truncated one), and Copilot Chat often doesn't know how to proceed.

**Root cause:** `step1.md` ran `discover.py --list-environments` (which prints a plain-text ASCII table to the terminal) and then asked the model to *read that terminal output, parse every row, and rebuild each environment into a `vscode_askQuestions` picker*. Copilot Chat truncates terminal scrollback and never surfaces the ASCII table to the user, so the reconstruction lost rows or failed entirely.

## Fix

Stop relying on the model scraping the terminal. `discover.py --list-environments` now also writes a complete, numbered **Markdown** table to `workspace/onboarding/environments.md` and prints an `ENV_TABLE_MARKDOWN:<path>` marker. `step1.md` reads that file, shows the full table to the user verbatim, asks them to type the row number, then runs `--select N`.

Benefits: no terminal scraping, no picker reconstruction, the user always sees the full table, and it scales to any number of environments.

## Changes

- **`list_environments.py`**: new `write_environment_markdown()` (pipe-escaped names, `(no Dataverse linked)` placeholder).
- **`discover.py`**: write the Markdown file and emit the `ENV_TABLE_MARKDOWN:` marker on the list (non-select) path.
- **`src/skills/onboarding/step1.md`**: §1.1/§1.1a/§1.1b now render the table from the file and use number-entry selection with an invalid-number retry.
- **`tests/scripts/test_discover.py`**: cover the Markdown writer (numbering, placeholder, pipe escaping).

## Validation

- New + existing `tests/scripts/test_discover.py` pass (16).
- Pre-existing failures in `test_installer_maker_profile.py` are unrelated (they also fail on `main`).

### DIFF
--- a/solutions/ess-maker-skills/scripts/discover.py
+++ b/solutions/ess-maker-skills/scripts/discover.py
@@ -87,6 +87,7 @@ def main():
         from list_environments import (
             get_dataverse_environments,
             print_environment_table,
+            write_environment_markdown,
         )
 
         dv_environments, excluded = get_dataverse_environments()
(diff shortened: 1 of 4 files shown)


nehaoss:

amilandi:
Abandoning — no longer pursuing this fix.

PR #169, opened 2026-06-25
flightcheck: add SN-CFG-002 — ServiceNow portal base URL populated in template config

### LINKED ISSUE #168: FlightCheck: Add check for ServiceNow portal base URL populated in template config

**Frequency in analyzed window:** 1 of 36 analyzed customer incident(s)
**Estimated applicability:** Conditional on ServiceNow HRSD/ITSM extension pack installed
**Affected feature:** ServiceNow integration (HRSD / ITSM extension packs)

## What should be checked

When a ServiceNow extension pack is installed, the agent must have the **ServiceNow portal base URL populated** in its template configuration so that returned tickets/cases render as resolvable hyperlinks. Today `SN-CFG-001` validates only that the expected ServiceNow template *config records exist* (by scenario name) in Dataverse — it does not validate that the **portal base URL value inside those configs is non-empty / well-formed**.

When the base URL value is blank, "read all tickets" / "read all cases" responses omit hyperlinks (links only appear when the user explicitly asks for them), degrading the experience without any error surfaced at deploy time.

## Suggested implementation

- **Checkpoint ID:** SN-CFG-002
- **Category:** ServiceNow
- **Priority:** MEDIUM
- **Conditional?** Yes, only runs when ServiceNow flows are detected (same gate as the existing ServiceNow deep validation, which returns early when `runner._servicenow_flows` is empty).
- **Validation logic:** For each ServiceNow template config found by `SN-CFG-001`, read the portal/base-URL field and assert it is set to a non-empty, http(s) absolute URL. WARN (with remediation pointing to the HRSD/ITSM extension-pack template config) when missing or empty.
- **Gating logic (if Conditional):** Only execute when a ServiceNow HRSD or ITSM solution/flow is detected (mirror `run_servicenow_checks` early-return on no ServiceNow flows).

## Verification

- **OOTB requirement?** No, conditional on the ServiceNow extension pack being installed.
- **ESS docs reference:** ServiceNow integration setup (extension-pack template configuration / portal base URL).
- **ESSVivaCopilot code path:** ServiceNow HRSD/ITSM extension-pack template configuration.
- **ESS Toolkit sample reference:** ServiceNow extension packs (HRSD + ITSM).
- **Existing FlightCheck pattern to mirror:** `SN-CFG-001` in `solutions/ess-maker-skills/scripts/flightcheck/checks/servicenow.py` (template-config presence, gated on detected ServiceNow flows). This new check extends it from *presence* to *value populated*.
- **Why a FlightCheck check is appropriate (vs. docs-only fix or no action):** The failure is silent — no runtime error — and only manifests as missing hyperlinks in production. A pre-flight check on the base-URL value catches it before rollout, and the gating/precedent already exist in SN-CFG-001.

## Context

This gap was identified by analyzing resolved customer incidents for the Employee Self-Service Agent service. Customers hit this issue because the FlightCheck tool did not warn them that the ServiceNow portal base URL was unset before deployment.

GrahamMcMynn:
Implements the FlightCheck check requested in #168, scoped and validated against a real PROD tenant.

**Checkpoint:** `SN-CFG-002`
**Category:** ServiceNow
**Priority:** MEDIUM
**Conditional?** Yes — only runs when ServiceNow flows are detected (same gate as `run_servicenow_checks`).

**What it validates**
Extends SN-CFG-001 from config *presence* to *portal base URL value populated*. The portal base URL is what renders clickable ticket/case hyperlinks; an empty value is the root cause of ICM 820635151 (links silently omitted).

**Scoping (validated against PROD)**
Live validation against a production tenant showed that of 21 ServiceNow-named template configs, only 2 carry the portal base URL: `msdyn_ServiceNowHRSD` and `msdyn_ServiceNowITSM`, each holding a `{ "ServiceNowPortalBaseURI": "<url>" }` object. The other 19 are scenario/field-mapping configs that never contain a URL. The check therefore evaluates only configs that carry a `ServiceNowPortalBaseURI` field, and validates that field's value specifically. This eliminates the 19 false positives the original name-only filter produced.

Detection rules (unchanged): WARNs (naming the offending configs) when the base URL value is blank, carries an unsubstituted placeholder (e.g. `{{ServiceNowBaseUrl}}`), or is only a relative path.

**Deference to the update-safe env var (pairs with #174)**
US 7535608 (PR #174) moves the base URL to update-safe environment variables (`ServiceNow{HRSD,ITSM}PortalBaseURI`, checked by SN-URL-001/002). Once a tenant migrates, the template config is intentionally left empty. To avoid SN-CFG-002 and SN-URL-001/002 double-firing, SN-CFG-002 defers per-pack: when the corresponding env var holds a valid URL, the empty template config is treated as expected (not a warning). Configs with no matching env var are still validated normally.

**Tests**
`tests/flightcheck/checks/test_servicenow_deep.py` — rewritten to the real config shape (`msdyn_ServiceNow{HRSD,ITSM}` with `ServiceNowPortalBaseURI`). Covers: all-populated → Passed; blank/placeholder/relative-path → Warning; scenario-configs-ignored (the 19-false-positive regression); root-config-only-flagged-amongst-scenarios (mirrors the PROD tenant); full and partial env-var deference; env-var-only; plus NotConfigured, Skipped, and query-error branches. 20 passing in the file, 76 passing across the ServiceNow suites.

Fixes #168


solutions/ess-maker-skills/scripts/flightcheck/checks/servicenow.py
```diff
@@ -249,6 +254,168 @@ def _validate_expected_configs(
...
+    unsubstituted base-URL placeholder token (e.g. ``{{baseUrl}}`` /
+    ``<ServiceNow Base URL>``) and no real URL alongside it, or when it
+    contains no absolute http(s) URL at all.
+    """
+    if not value or not value.strip():
+        return False
+
+    candidates = _ABSOLUTE_URL_RE.findall(value)
+    has_real_url = any(_is_absolute_http_url(c) for c in candidates)
+    if has_real_url:
```
GrahamMcMynn, line 623: SN-CFG-002 can still pass an unsubstituted placeholder when the placeholder appears in the host. For example, `https://{{ServiceNowBaseUrl}}/api/now/table/incident` is matched as `https://{{ServiceNowBaseUrl`, and `_is_absolute_http_url()` accepts it because it has an http(s) scheme and non-empty `netloc`, so `has_real_url` becomes true. That is one of the bad states this check is meant to catch. Please reject placeholder tokens before accepting URL candidates (or validate the parsed host excludes template tokens) and add a test for this scheme-prefixed placeholder case.

tests/flightcheck/checks/test_servicenow_deep.py
```diff
@@ -131,6 +135,123 @@ def test_template_configs_skipped_without_token():
...
+                name="ServiceNowHRSDGetCasesList",
+                value="{{ServiceNowBaseUrl}}/api/now/table/sn_hr_core_case",
+            ),
+        ],
+    )
+    from flightcheck.checks.servicenow import _check_template_config_base_urls
+    runner = SimpleNamespace(env_url="https://org.crm.dynamics.com", dv_token="t")
+    cfg = _by_id(_check_template_config_base_urls(runner), "SN-CFG-002")
+
+    assert cfg.status == "Warning"
```
GrahamMcMynn, line 237: `tests/AGENTS.md` requires every WARNING test to assert specific phrases from both `result` and `remediation`, including the impact and concrete fix path. This warning test only pins `result`; `test_base_url_relative_path_only_warns` and `test_base_url_query_error_warns` below have the same gap. As written, the remediation text for these actionable warnings could regress or disappear without failing tests. Please add remediation assertions for each SN-CFG-002 Warning branch.

GrahamMcMynn (commented):
Code review by ESS Code Review Agent (Auto mode)

tests/mocks/dataverse.py
```diff
@@ -240,6 +240,45 @@ def workday_connection_refs_full() -> list[dict[str, Any]]:
...
+) -> dict[str, Any]:
+    """Build a single msdyn_employeeselfservicetemplateconfigs record.
+
+    Cited consumers:
+      - solutions/ess-maker-skills/scripts/flightcheck/checks/servicenow.py
+        — SN-CFG-001 reads msdyn_name; SN-CFG-002 reads msdyn_value to
+        verify the ServiceNow portal base URL inside the value is a
+        populated, well-formed absolute http(s) URL.
+
+    Source (documented):
```
GrahamMcMynn, line 258: The documented-tier mock requirements in `tests/AGENTS.md` require the builder docstring to cite an anchored vendor doc and include the documented example response verbatim. This builder cites generic Dataverse query docs plus `backup_template_configs.py`, but it does not include a documented example for the `msdyn_employeeselfservicetemplateconfigs` shape/fields it returns. That leaves tests depending on a shape that is not verified by one of the permitted tiers. Please add the required documented example evidence, or promote this surface with cassette/schema evidence before using the builder in FlightCheck tests.

solutions/ess-maker-skills/scripts/flightcheck/checks/servicenow.py
GrahamMcMynn: Fixed in d240a6c. _is_absolute_http_url now rejects any netloc containing placeholder characters (`{}<>%$ `), so https://{{ServiceNowBaseUrl}}/... no longer passes. Added a regression test pinning the scheme-prefixed form (plus a bare-token case).

tests/flightcheck/checks/test_servicenow_deep.py
GrahamMcMynn: Fixed in d240a6c. Both warning tests now assert remediation phrases (https://<instance>.service-now.com and the omit/hyperlinks impact), and a new bare-placeholder warning test was added.

tests/mocks/dataverse.py
GrahamMcMynn: Fixed in d240a6c. The 	emplate_config builder docstring now carries the verbatim field contract (cited to backup_template_configs.py \ + record mapping) wrapped in a documented OData v4 collection-envelope example.

nehaoss:

daeunJe0ng:
Closing this in favor of #174.

Both checks validate the same ServiceNow portal base URL. #174 validates the update-safe environment variables (SN-URL-001/002), which are the real source of truth after the URL moved out of template config under US 7535608 to fix ICM 820635151. The template-config value ships empty and resets on managed-solution updates, so the env-var check supersedes this one.

#174 has been reviewed, fixed, and validated end to end against a live tenant. Thanks Graham for the original work here.

PR #171, opened 2026-06-26
WD-RUN-001: error-aware suppression of MANUAL Workday verification checks

GrahamMcMynn:
## What & why

The MANUAL Workday verification checks — WD-CONN-010 (Entra↔Workday SAML federation), WD-CONN-102 (SAML signing-cert health), and WD-SEC-003 (Personal Data write permission) — ask the operator to hand-verify config that FlightCheck can't probe. Previously they were suppressed all-or-nothing: hidden when WD-RUN-001 (run-health litmus) passed, shown otherwise.

This PR makes the suppression **error-aware**: when runs are failing, it classifies the failure and shows only the manual verification(s) the observed error cannot rule out, so the operator is pointed at the likely culprit instead of all three.

## How it works (no new API / cassette)

The only run-history-visible signal of *what kind* of failure occurred is the flow's **Response-action branch** (the Workday faultstring lives behind a separate SAS-signed `outputsLink` we deliberately do not fetch). `_compute_run_failure_signal` summarises the recent window into:

- **`auth_proven`** — a success, a caught Workday fault, or a transform-failed branch ⟹ Workday accepted the OAuthUser SSO token and answered ⟹ the SSO / federation / signing-cert path works.
- **`workday_fault`** — a caught `failure_errorMessage` branch ⟹ Workday processed and rejected ⟹ permission/credential signal.
- **`hard_failure`** — a no-Response terminal failure (informational).

### Suppression rules (when WD-RUN-001 is not passing)

A manual check is removed **only on positive rule-out evidence**:

| Evidence | Suppresses |
|---|---|
| `auth_proven` | WD-CONN-010 + WD-CONN-102 (SSO domain proven OK) |
| `auth_proven` **and** no caught fault | also WD-SEC-003 (Workday answered, no permission denial) |

Everything else stays visible:
- **caught Workday fault** ⟹ only WD-SEC-003 shown (SSO ruled out).
- **no-response / uninformative / no signal** ⟹ all shown (nothing ruled out).
- **safety net** — if the rule-outs would hide *every* manual check while runs are unhealthy, that's contradictory (the break is outside the modelled domains), so show all.

WD-RUN-001 **PASSED** is unchanged: a clean litmus pass suppresses all three.

## Honest limit

Run history can narrow to a *domain* but cannot split WD-CONN-010 from WD-CONN-102 — both are the same OAuthUser SSO path and run history never says "cert expired" vs "federation misconfigured" — so an SSO-family break surfaces both. Permission (WD-SEC-003) is independently isolable.

## Tests

`tests/flightcheck/checks/test_workday_run_health.py`: signal-helper coverage across every branch, the full rule-out matrix, the safety net, the PASS short-circuit, and a wiring test pinning that WD-RUN-001 stashes the signal. The 3 pre-existing suppression tests are preserved unchanged (the new `runner` arg is optional). Full `tests/flightcheck` suite: **524 passed**; ruff clean.

### DIFF
--- a/solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py
+++ b/solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py
@@ -747,6 +747,12 @@ def _suppress_manual_conn_sec_when_runs_healthy(
 
     if run_health == Status.PASSED.value:
         # Clean litmus pass — hide every MANUAL Workday connection/security row.
+        # NOTE: deliberately BROAD (prefix match) — a clean pass should silence
+        # ALL manual conn/sec asks, including any added later. The not-passed
+        # branch below is intentionally NARROW (exact ids in _RUNHEALTH_*),
+        # because failure suppression must be evidence-based. Do not "align"
+        # the two: broadening the failure branch would hide manual checks the
+        # error signal hasn't actually ruled out.
         return [
             r for r in results
             if not (
@@ -777,9 +783,11 @@ def _suppress_manual_conn_sec_when_runs_healthy(
 
     if not suppress:
         return results
-    if suppress >= _RUNHEALTH_ALL_SUPPRESSIBLE:
+    if suppress == _RUNHEALTH_ALL_SUPPRESSIBLE:
         # Ruled out every manual check while runs are unhealthy → the break is
         # outside the model's domains (or we misclassified). Fail safe: show all.
+        # (suppress is only ever drawn from _RUNHEALTH_ALL_SUPPRESSIBLE, so this
+        # equality means "every suppressible domain was ruled out".)
         return results
     return [
         r for r in results
@@ -3144,7 +3152,7 @@ def _classify_run(run: dict) -> str:
     return "pending"
 
 
-def _compute_run_failure_signal(window: list[dict]) -> dict:
+def _compute_run_failure_signal(window: list[dict]) -> dict[str, bool]:
     """Summarise which failure-branch categories appear in the recent run window
     so run-health can suppress only the MANUAL Workday verification checks whose
     failure DOMAIN the observed errors positively rule out (see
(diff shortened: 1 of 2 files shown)


apurvabanka:
Overall: Well-designed and well-tested. The approach of using positive evidence to rule out failure domains (rather than guessing from absence) is sound, and the safety net is a smart fallback.

3 minor suggestions:

Return type annotation (_compute_run_failure_signal): -> dict could be -> dict[str, bool] for a self-documenting contract.

Set comparison (line ~770): suppress >= _RUNHEALTH_ALL_SUPPRESSIBLE can only ever be == since suppress is populated exclusively from that same frozenset. Using == communicates intent more precisely.

PASSED branch breadth (lines ~735–741): The PASSED branch uses broad startswith("WD-CONN")/startswith("WD-SEC") matching, while the error-aware branch uses exact checkpoint_id in suppress. This is likely intentional but a brief comment explaining the deliberate breadth would prevent future accidental "alignment" that could break suppression for new checks.

CC: @GrahamMcMynn

GrahamMcMynn:
Thanks @apurvabanka — applied all three in d37d525:

1. `_compute_run_failure_signal` now annotated `-> dict[str, bool]`.
2. Safety-net guard changed to `suppress == _RUNHEALTH_ALL_SUPPRESSIBLE` — you're right that `suppress` is only ever drawn from that frozenset, so `>=` could only be `==`; the equality states the intent ("every suppressible domain ruled out").
3. Added a NOTE on the PASSED branch documenting the deliberate breadth (broad prefix match silences all manual conn/sec on a clean pass, including future checks) vs. the intentionally narrow exact-id failure branch, with an explicit "do not align the two" warning.

No behavior change; 28 run-health tests + full flightcheck suite (563) still green, ruff clean.

apurvabanka (approved):

PR #173, opened 2026-06-26
WD-CONN-013: Agent connection OBO parameter-sharing check

GrahamMcMynn:
## What & why

Adds **WD-CONN-013 — Agent connection OBO parameter sharing**. It verifies that every connection the agent uses has *Allow permission to share parameters* (On-Behalf-Of) enabled, so end users invoke the backing services (Workday, etc.) without being prompted to establish their own connection at first use — the requirement in Microsoft Learn → Copilot Studio → *Share connection parameters for On-Behalf-Of (OBO) authentication* and the Workday Step 2/3 setup docs.

## How it works (no new endpoint, no cassette)

The OBO `Allow permission to share parameters` toggle is persisted on the agent's own Dataverse `connectionreference` rows in column **`connectionparametersetconfig`** (populated = shared, null = not shared) — confirmed empirically by toggling it live on a tenant and diffing the row. The check reads the **same documented-tier `connectionreferences` query WD-PKG-001 already makes**, with two extra columns (`connectionparametersetconfig` / `connectionparametersconfig`).

Two things were ruled out during investigation:
- `promptingbehavior` — unrelated (it's the solution-import *Prompt on import / Skip* choice, per the entity reference).
- The connection `/permissions` role-assignments endpoint — a different, unrelated sharing surface (and its connection objects leak Workday tenant/clientId, so it was abandoned).

### Scoping to the agent's connections

Copilot Studio names the agent's own connection references `{schema}.{guid}.{connector}` (a GUID delimited by dots). The check matches that **structural format** (any connector) and excludes the unused solution-template refs (`new_sharedworkdaysoap_ff0df`, `msdyn_sharedcommondataserviceforapps_92b66` — underscore-named, no `.{guid}.` segment). We deliberately match on the structural format rather than `config.agent.schemaName`: the reference prefix is the *product* bot schema, which does not necessarily equal the published agent's schema (an earlier schema-prefix attempt produced a false NOT_CONFIGURED).

### Verdicts

- **PASSED** — every agent connection has parameter sharing enabled.
- **FAILED** — lists the unshared connections, with the *Copilot Studio → Settings → Connection Settings → See details → Connection parameters → Allow permission to share parameters* remediation.
- **NOT_CONFIGURED** — no agent connection references found.
- **SKIPPED** — no Dataverse token / read error.

## Known limitation

`connectionparametersetconfig` being null is treated as "not shared" for every agent connector. A connector that does not support parameter sharing could therefore surface as a finding. This false-positive risk is accepted in favour of covering **every** connection the agent references (the requested scope).

## Tests

`tests/flightcheck/checks/test_workday_connection_obo_sharing.py` — 9 tests: PASS/FAIL, multi-connector coverage (a non-Workday agent connection), solution-ref exclusion, the alternate `connectionparametersconfig` column, NOT_CONFIGURED/SKIPPED, and wiring. Full `tests/flightcheck` suite: **559 passed**; ruff clean.

## Live validation

Verified end-to-end on a real simplified-install tenant: with both Workday connections' *Allow permission to share parameters* turned on, WD-CONN-013 reports **Passed**; with one off it correctly **FAILs** and names the connection.

### DIFF
--- a/tests/flightcheck/checks/test_workday_connection_obo_sharing.py
+++ b/tests/flightcheck/checks/test_workday_connection_obo_sharing.py
@@ -23,18 +23,24 @@ from typing import Any
 import pytest
 
 SCHEMA = "msdyn_copilotforemployeeselfservicehr"
-_WD_CONNECTOR = "/providers/Microsoft.PowerApps/apis/shared_workdaysoap"
 _SHARED_CFG = '{"name":"oauth","values":{}}'
 
+# Real GUID-format middle segments — the agent connection-ref detector requires
+# a ``.{guid}.`` segment.
+G1 = "9f1b2c3d-4e5f-6789-abcd-1234567890ab"
+G2 = "aab52569-483a-f111-8e38-0022480875be"
+G3 = "11111111-2222-3333-4444-555555555555"
 
-def _agent_ref(guid: str, *, shared: bool = True, display: str = "ESS HR Workday",
+
+def _agent_ref(guid: str, *, connector: str = "shared_workdaysoap",
+               shared: bool = True, display: str = "ESS HR Workday",
                via_params_config: bool = False):
     setcfg = None if (via_params_config or not shared) else _SHARED_CFG
     paramscfg = _SHARED_CFG if (via_params_config and shared) else None
     return {
-        "connectionreferencelogicalname": f"{SCHEMA}.{guid}.shared_workdaysoap",
+        "connectionreferencelogicalname": f"{SCHEMA}.{guid}.{connector}",
         "connectionreferencedisplayname": display,
-        "connectorid": _WD_CONNECTOR,
+        "connectorid": f"/providers/Microsoft.PowerApps/apis/{connector}",
         "connectionid": f"conn-{guid}",
         "connectionparametersetconfig": setcfg,
         "connectionparametersconfig": paramscfg,
@@ -42,11 +48,11 @@ def _agent_ref(guid: str, *, shared: bool = True, display: str = "ESS HR Workday
 
 
 def _solution_ref(name: str = "new_sharedworkdaysoap_ff0df"):
-    # A bound Workday SOAP ref that is NOT the agent's (no schema prefix), unshared.
+    # A bound ref that is NOT the agent's (no ``.{guid}.`` segment), unshared.
     return {
         "connectionreferencelogicalname": name,
         "connectionreferencedisplayname": "OAuthUser",
-        "connectorid": _WD_CONNECTOR,
+        "connectorid": "/providers/Microsoft.PowerApps/apis/shared_workdaysoap",
         "connectionid": "conn-ff0df",
         "connectionparametersetconfig": None,
         "connectionparametersconfig": None,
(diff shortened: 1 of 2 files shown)


tests/flightcheck/checks/test_workday_connection_obo_sharing.py
```diff
@@ -0,0 +1,202 @@
...
+
+"""Tests for WD-CONN-013 (Workday connection OBO parameter sharing).
+
+The check verifies that every Workday connection the *agent* uses has
+"Allow permission to share parameters" enabled. That setting is persisted on the
+agent's own ``connectionreference`` rows (logical name
+``{agentSchemaName}.{guid}.{connector}``) in column ``connectionparametersetconfig``
+(populated = shared, null = not shared) — confirmed empirically by toggling it
+live. Solution-template refs (``new_sharedworkdaysoap_ff0df`` etc.) are bound but
+are NOT the agent's connections, so they're excluded by the schema-name prefix.
```
GrahamMcMynn: The module docstring says solution-template refs "are excluded by the schema-name prefix," but the implementation excludes them structurally via the `.{guid}.` regex (`_AGENT_CONNECTION_REF_RE`), not by a schema-name prefix. The PR description explicitly notes schema-name matching was *deliberately rejected* (it produced a false NOT_CONFIGURED). This docstring contradicts both the code and the stated design rationale, and could mislead a future maintainer into "fixing" the detector to use the schema prefix.

Suggested reword: "...excluded because their underscore-only logical names carry no `.{guid}.` segment (the agent-connection structural format)."

GrahamMcMynn (commented):

tests/flightcheck/checks/test_workday_connection_obo_sharing.py
GrahamMcMynn: Good catch — fixed in 11ee63f. Reworded to your suggestion (excluded because the underscore-only logical names carry no `.{guid}.` segment), and also generalized the leftover `Workday connection` wording in the same docstring to `every connection the agent uses` to match the broadened check. The test docstring now agrees with the `_AGENT_CONNECTION_REF_RE` detector and the module comment in workday.py.

daeunJe0ng:
Overall this is a solid, well-documented check and the empirical write-up is genuinely helpful. The two-column sharing logic is correct: per the Dataverse connectionreference entity reference, both `connectionparametersetconfig` ("connection parameters set that can be reused when creating connections in Copilot Studio") and `connectionparametersconfig` ("connection parameters that can be reused...") are the same OBO parameter-sharing surface, so ORing them is the right call and is more robust than a single column. A few things worth tightening before merge.

### Worth addressing

1. Scope is environment-wide, not agent-scoped.
`_AGENT_CONNECTION_REF_RE` matches any `{schema}.{guid}.{connector}` reference, so in a tenant that hosts more than one Copilot Studio agent the check evaluates other agents' connections too. That can produce a false FAIL on an unrelated agent, or a PASS that hides the ESS agent. The category and description say "the agent" but the logic says "every agent." Consider scoping to the connectors the ESS agent actually uses (you already cache the Workday refs in `runner._workday_connection_refs`, and `connectorid` is in the select), or rename to make the environment-wide scope explicit.

2. Naming-format drift degrades to a misleading NOT_CONFIGURED.
If Copilot Studio changes the dotted-GUID logical-name convention, `agent_refs` becomes empty and the check reports "No agent connection references found," hiding that the agent does have connections. Suggest cross-checking the cached `runner._workday_connection_refs` (WD-PKG-001 runs first): if it is non-empty while `agent_refs` is empty, emit WARNING instead of NOT_CONFIGURED.

3. Read-failure verdict is inconsistent with WD-PKG-001.
A connectionreferences read error is SKIPPED here but WARNING in WD-PKG-001. Same root cause, different label. Reserve SKIPPED for "no token" and use WARNING for an actual query/permission error.

### Test gaps

- Every test hand-builds the config column values, so they validate the code against its own assumption rather than against Dataverse. One test using a real captured row would lock down the column behavior.
- No test for the disclosed false-positive (a connector that supports neither column leaves both null and FAILs). A short documenting test would stop a future maintainer from "fixing" the accepted limitation.

### Optional hardening

- Add a one-line comment next to the OR citing both Learn column descriptions, so nobody later removes the second column thinking it is spurious.
- The check re-paginates the full connectionreferences table that WD-PKG-001 already pulled. Adding the two columns to that query's select and reusing the cache would save a second full pagination in large environments.
- `connectorid` is selected but unused; it is the natural key for the agent-scoping in point 1.

Nothing here is blocking. Mostly scoping clarity, clearer signals on degraded states, and locking the API assumption down with a real-row test.

apurvabanka (approved):

PR #174, opened 2026-06-26
feat(flightcheck): validate ServiceNow portal base URL env vars (US 7535608)

srideshpande:
## What & why
Adds ESS FlightCheck coverage for the update-safe ServiceNow portal base URL environment variables introduced for **US 7535608** (Option B). An empty base URL is the root cause of **ICM 820635151** — ServiceNow record hyperlinks only returned when explicitly requested.

## Changes
- **New checks** `SN-URL-001` (HRSD) / `SN-URL-002` (ITSM) in `servicenow.py`, wired into `run_servicenow_checks`. Each verifies its env var (`msdyn_ServiceNow{HRSD,ITSM}PortalBaseURI`) is present and a well-formed absolute URL:
  - **PASSED** – valid absolute URL
  - **FAILED** – definition exists but no value (the maker step was missed — the ICM bug)
  - **WARNING** – value set but not an absolute URL
  - **NotConfigured** – pack not installed
  - **SKIPPED** – no Dataverse token
- **New test** `test_servicenow_base_url.py` — 6 cases (both-set, missing-value, per-pack independence, malformed URL, definition-absent, no-token), mirroring the Workday env-var test and using the documented-tier Dataverse mock.

## Validation
`pytest tests/flightcheck/checks/test_servicenow_base_url.py` → 6 passed. Full ServiceNow + role-enforcement suite (37 tests) green.

## Related
Pairs with the ESSVivaCopilot solution PR that creates these env vars + wires the orchestrator flows (US 7535608 Tasks 2–3).


nehaoss:

daeunJe0ng:
Review summary — SN-URL-001 / SN-URL-002 (ServiceNow portal base URL env vars)

Verdict: Approve. I reviewed the design and logic, validated the checks against a live EmployeeHub environment, and applied one small efficiency fix with a test. Nothing else blocks merge.

Design.
This is the right home for the fix. The portal base URL moved to update-safe environment variables under US 7535608 precisely because the template-config value ships empty and resets on every managed-solution update, which is the root cause of ICM 820635151. Validating the env vars is validating the actual source of truth, so SN-URL-001/002 supersede the template-config check.

Live validation (EmployeeHub tenant).
I ran the check's real Dataverse queries against every accessible ServiceNow environment.
On ESS DA ServiceNow UpdateCase Validate the HRSD variable is provisioned as ServiceNowHRSDPortalBaseURI with no publisher prefix and holds a valid absolute URL, and the check correctly returned PASSED for SN-URL-001 and NotConfigured for SN-URL-002. This confirms the schema-name alignment from the 2026-07-24 commit matches what the shipped solution actually provisions.
On the other accessible ServiceNow environments the PortalBaseURI definitions are not present yet, and the check correctly returned NotConfigured for both packs.
The FAILED (defined-but-empty) and WARNING (non-absolute value) states are covered by unit tests but were not reproducible on a live tenant, because no accessible environment is currently in those states.

Fix applied.
The values lookup previously pulled every environmentvariablevalues record in the environment and filtered client-side. I narrowed it to query only the PortalBaseURI definition ids found by the definitions query, and to skip the values query entirely when no definitions exist. Behavior is unchanged; it just avoids a full environment-variable scan on every run. I verified the _environmentvariabledefinitionid_value eq <id> filter returns the correct single value against the live environment.

Tests.
Added an assertion that no environmentvariablevalues query is issued when no definitions are found. test_servicenow_base_url.py is 6 passed, the full tests/flightcheck suite is 1002 passed, and ruff is clean on both changed files.

The fix is pushed to this branch as commit 49e07b6.

daeunJe0ng (approved):

PR #175, opened 2026-06-29
feat(scaffold): [Master][MTK] | ESS Migration Toolkit dev-specs for SDD and initial scaffold

anil-adepu:
## Description
Initial scaffold for the **ESS NextGen Migration Toolkit** (CA → DA migration),
built under strict Specification-Driven Development (SDD). This lays the
foundation that subsequent, parallelizable task streams build on:

- **Repository scaffold (TASK-001)** — full `src/` package tree (core, models,
  constants, services, sdk, migration, pre/postprocessing, ui, utils),
  `tests/{unit,integration,golden,e2e}`, `logs/reports/scripts`, plus
  `pyproject.toml` (hatchling, src-layout), `.pre-commit-config.yaml`,
  `.gitignore`, and `README.md`. **NO BUSINESS LOGIC.**
- **Deterministic environment (uv)** — committed `uv.lock`, pinned
  `.python-version`, exported `requirements*.txt` for plain-pip consumers, and
  self-bootstrapping `scripts/setup.{sh,ps1}` (auto-install pip + uv if missing,
  then `uv sync`). Gives customers and internal devs a reproducible, dependency-
  error-free run.

Specs remain the source of truth (`dev-specs/`); `tools/` has the actual implementation.

## Master-Slave Architecture for SDD orchestration
This toolkit is built by a **master/worker (orchestrator/implementer) agent
setup** so that spec-driven tasks can be planned, delegated, and reviewed in a
controlled, parallelizable way. The two roles are defined as repo-local agents:

- **`.github/agents/mtk-master.md` — MTK Master (orchestrator).** Owns the
  backlog, selects disjoint task batches (max 2 in parallel), creates an
  isolated **git worktree + branch per task**, spawns workers as background
  subagents, then **reviews their diffs against Acceptance Criteria**, runs the
  gates, and **integrates serially**. It plans, assigns, reviews, and
  reports — it does **not** implement tasks itself.
- **`.github/agents/mtk-worker.md` — MTK Worker (implementer).** Assigned
  **exactly one** `TASK-XXX`, works **entirely within its dedicated worktree**
  (`../mtk-TASK-XXX`), resolves only the specs in its task's References, writes
  code in the correct architectural layer, and runs the quality gates. Worktree
  isolation is what guarantees parallel workers never collide on disk.

```text
You ──one conversation── MTK Master (.github/agents/mtk-master.md)
                              │ task(mode:background) ×2
                      ┌───────┴───────┐
                  Worker A         Worker B   ← .github/agents/mtk-worker.md
               worktree TASK-A   worktree TASK-B   (isolated dirs, own .venv)
```

The contract each role obeys lives alongside the specs:
- **Operating manual:** `dev-specs/ess-nextgen-migration-toolkit/AGENTS.md`
  (AI Execution Algorithm, Dependency-Based Loading Model, Specification
  Hierarchy).
- **Supreme invariants:** `dev-specs/ess-nextgen-migration-toolkit/00_META/INVARIANTS.md`.
- **Backlog index + per-task files:**
  `dev-specs/ess-nextgen-migration-toolkit/04_EXECUTION/TASKS.md` and
  `.../04_EXECUTION/tasks/TASK-XXX-*.md`.

This PR (TASK-001) is the scaffold that this master/worker loop consumes for
every downstream task (TASK-002+).

## Type of change
- [x] Documentation update
- [ ] Refactor / cleanup

## Testing
Quality gates run via `uv` in the toolkit folder
(`tools/ess-nextgen-migration-toolkit/`):

- `uv run ruff check .` — passes (35 source files)
- `uv run mypy src` — passes (strict; no issues)
- `uv run pytest -q` — no tests yet (scaffold only); contract imports/construct
  smoke-verified manually
- `uv lock --check` — lockfile up to date
- Editable install + `scripts/setup.sh` bootstrap verified in a throwaway venv;
  `setup.ps1` parse-checked with pwsh

## Checklist

N/A — this PR does not touch `samples/` (changes are scoped to
`tools/ess-nextgen-migration-toolkit/`, `dev-specs/`, and `.github/agents/`).

```text
Validation
- YAML parse: N-A
- AdaptiveDialog kind: N-A
- XML parse: N-A
```


nehaoss (commented):
Review comments posted inline on specific code lines.

tools/ess-nextgen-migration-toolkit/src/core/auth/token_provider.py
```diff
@@ -0,0 +1,200 @@
...
+        if account is not None:
+            self._logger.debug("Attempting silent token acquisition.")
+            result = self._app.acquire_token_silent(normalized_scopes, account=account)
+            if result is not None and self._token_needs_refresh(result):
+                self._logger.debug("Cached token is near expiry; forcing silent refresh.")
+                result = self._app.acquire_token_silent(
+                    normalized_scopes,
+                    account=account,
+                    force_refresh=True,
+                )
```
nehaoss, line 101: When \force_refresh=True\ returns \None\ (e.g. network glitch), the code falls through to \if result is None: if account is not None: raise\. This means a transient refresh failure with a valid (near-expiry) cached token raises an exception rather than returning the still-usable cached token. Consider falling back to the original near-expiry token when \force_refresh\ fails, especially since the original token hasn't expired yet (just within \
refresh_window_seconds\).

tools/ess-nextgen-migration-toolkit/scripts/mtk.sh
```diff
@@ -0,0 +1,198 @@
...
+
+# Echoes the uv path on stdout; all progress/log output goes to stderr so the
+# path can be safely captured with command substitution.
+ensure_uv() {
+  local uv
+  uv="$(find_uv || true)"
+  if [[ -z "$uv" ]]; then
+    echo "==> uv not found; installing the standalone uv (no Python required)..." >&2
+    if command -v curl >/dev/null 2>&1; then
+      curl -LsSf https://astral.sh/uv/install.sh | sh 1>&2
```
nehaoss, line 46: The \curl | sh\ pattern for installing uv has no checksum or signature verification. A compromised CDN or MITM (even on HTTPS with cert pinning issues) could inject arbitrary code. Consider either: (1) pinning to a known-good version with a checksum, e.g. \curl ... | sha256sum -c - && sh\, or (2) at minimum documenting the trust assumption explicitly so operators can make an informed decision. The same applies to the Windows \irm | iex\ pattern in \mtk.ps1\.

tools/ess-nextgen-migration-toolkit/pyproject.toml
```diff
@@ -0,0 +1,63 @@
...
+
+# The canonical layout
+# (dev-specs/ess-nextgen-migration-toolkit/03_ENGINEERING/REPOSITORY_STRUCTURE.md)
+# places top-level packages directly under src/. The build maps them explicitly.
+[tool.hatch.build.targets.wheel]
+packages = [
+    "src/constants",
+    "src/core",
+    "src/service",
+]
```
nehaoss, line 38: The \[tool.hatch.build.targets.wheel]\ only packages \src/constants\, \src/core\, and \src/service\ — but the \src/modules\ package (with \migration\, \preprocessing\, \postprocessing\ subpackages) is omitted. When later tasks implement these modules, the wheel won't include them unless this is updated. Consider adding \\src/modules\\ now to avoid a silent packaging miss when business logic is added.

tools/ess-nextgen-migration-toolkit/scripts/mtk.ps1
```diff
@@ -0,0 +1,181 @@
...
+
+function Ensure-Uv {
+    $uv = Find-Uv
+    if (-not $uv) {
+        Write-Host "==> uv not found; installing the standalone uv (no Python required)..."
+        powershell -ExecutionPolicy ByPass -Command "irm https://astral.sh/uv/install.ps1 | iex"
+        $uv = Find-Uv
+        if (-not $uv) {
+            Write-Error ("uv installation did not produce a usable binary. Open a new " +
+                "shell (so PATH refreshes) and re-run, or install uv manually: " +
```
nehaoss, line 60: PowerShell best practice: \Ensure-Uv\ uses an unapproved verb. PowerShell will emit a warning when loading modules with non-standard verbs, and \PSScriptAnalyzer\ will flag it. Rename to \Assert-Uv\ or \Initialize-Uv\ (approved verbs list: \Get-Verb\). Similarly \Find-Uv\ is fine (approved), but \Invoke-Provision\ / \Invoke-Run\ are good.

tools/ess-nextgen-migration-toolkit/.pre-commit-config.yaml
```diff
@@ -0,0 +1,45 @@
...
+        language: system
+        types_or: [python, pyi]
+        require_serial: true
+
+      - id: mypy
+        name: mypy (strict) — toolkit only
+        entry: uv run --project tools/ess-nextgen-migration-toolkit mypy tools/ess-nextgen-migration-toolkit/src
+        language: system
+        files: ^tools/ess-nextgen-migration-toolkit/.*\.py$
+        pass_filenames: false
```
nehaoss, line 45: The mypy hook uses an absolute-looking path (\mypy tools/ess-nextgen-migration-toolkit/src\) but pre-commit doesn't guarantee the cwd is the monorepo root — it's typically the git toplevel, which works in this case. However, the \iles:\ regex already restricts triggering to toolkit paths, yet \pass_filenames: false\ + the hardcoded path means mypy always type-checks the entire \src/\ tree even if one unrelated file matched the regex. This is fine for now but may cause slow hooks as the codebase grows. Consider changing to \ntry: uv run --project . mypy src\ with a \cwd:\ override if pre-commit ever adds that, or note this as a known trade-off.

tools/ess-nextgen-migration-toolkit/scripts/mtk.ps1
```diff
@@ -0,0 +1,181 @@
...
+param(
+    [Parameter(Position = 0)]
+    [ValidateSet("start", "refresh", "help")]
+    [string]$Command = "help",
+
+    [switch]$Dev
+)
+
+$ErrorActionPreference = "Stop"
+# Always operate from the toolkit root (this file lives in <root>\scripts\).
```
nehaoss, line 35: Line 35 does \Set-Location\ unconditionally which changes the caller's working directory when this script is dot-sourced or run without a child scope. While the root forwarder runs this in a child scope (via \&\), if a user ever sources this script directly (e.g. \. ./scripts/mtk.ps1 start\), their shell cwd will be permanently changed. Consider wrapping the entire script body in \Push-Location ... try { ... } finally { Pop-Location }\ to be safe regardless of invocation style.

tools/ess-nextgen-migration-toolkit/src/core/auth/token_provider.py
```diff
@@ -0,0 +1,200 @@
...
+            raise AuthenticationException("Authentication token acquisition failed.")
+
+        return access_token
+
+    def _token_needs_refresh(self, result: TokenResult) -> bool:
+        now = self._now()
+        expires_at = _token_expiry_epoch(result, now)
+        if expires_at is None:
+            return True
+        return expires_at <= now + self._config.refresh_window_seconds
```
nehaoss, line 126: When a token result has neither \xpires_on\ nor \xpires_in\ (some MSAL error-dict responses), \_token_expiry_epoch\ returns \None\ and \_token_needs_refresh\ returns \True\. However, this same path is hit for *error* results that lack \ccess_token\ — in that case the code correctly falls through to the \ccess_token\ check below. But what if MSAL returns a valid \ccess_token\ without expiry metadata (e.g. a buggy identity provider)? The provider would reject a usable token. Consider logging a warning and returning the token anyway when \ccess_token\ is present but expiry is unparseable, rather than unconditionally raising.

tools/ess-nextgen-migration-toolkit/scripts/mtk.sh
```diff
@@ -0,0 +1,198 @@
...
+  local dev="$1"
+  local UV
+  UV="$(find_uv || true)"
+  echo ""
+  echo "==> Starting the toolkit CLI..."
+  if [[ "$dev" == "1" ]]; then
+    exec "$UV" run python src/service/mtk_orchestrator.py
+  else
+    exec "$UV" run --no-dev python src/service/mtk_orchestrator.py
+  fi
```
nehaoss, line 130: \xec\ replaces the current process, which means any shell traps, cleanup, or subsequent lines after \cmd_run\ in \cmd_start\ will never execute. This is intentional for \start\ (provision-then-run ends the script), but if future commands are added that call \cmd_run\ as part of a larger sequence (e.g. \	est\ = provision + run-tests + report), the \xec\ will silently break the chain. Consider using a plain invocation (no \xec\) and explicitly \xit \True\ if you want to propagate the exit code without replacing the process.

tools/ess-nextgen-migration-toolkit/scripts/mtk.sh
```diff
@@ -0,0 +1,198 @@
...
+    "$UV" sync --no-dev
+  fi
+
+  if ! command -v uv >/dev/null 2>&1; then
+    echo ""
+    echo "Tip: add uv to your PATH so you can call it directly next time:"
+    echo "      echo 'export PATH=\"\$HOME/.local/bin:\$PATH\"' >> ~/.zshrc && source ~/.zshrc"
+  fi
+}
+
```
nehaoss, line 115: The PATH tip hardcodes \~/.zshrc\ — this will confuse bash users (still common on Linux). Consider detecting the active shell (\\\) or just printing a generic \xport PATH=\\C:\Users\nbethapudi/.local/bin:\\\ without the append-to-file part, letting the user choose where to put it.

amilandi (dismissed):
Reviewed the actual code (the bulk is specs/scaffold). Ran the token-provider unit tests locally — **6 passed** — and exercised the launcher scripts. Nice, well-structured scaffold. Two non-blocking notes below; otherwise LGTM.

## Strengths
- **`token_provider.py`** is security-conscious: MSAL cache is deliberately in-memory/process-local (no `SerializableTokenCache`, no `token_cache` path — DIAG-003), authority + scopes are HTTPS-validated, and `AuthenticationException` is sanitized so provider error details never leak into the message (explicitly asserted in `test_acquisition_failure_raises_sanitized_authentication_exception`). Protocol-based DI makes it cleanly testable.
- **Tests** cover cold-start-interactive, silent, near-expiry force-refresh, non-HTTPS rejection, secret-sanitization, and cache-isolation.
- **pyproject**: src-layout, `mypy` strict, `ruff` with `T20` (enforces the no-`print()` diagnostics rule), pip-free `uv` workflow.
- **Scripts**: `mtk.sh` uses `set -euo pipefail` with correct quoting and stdout/stderr separation for the `ensure_uv` command-substitution; `mtk.ps1` uses `ValidateSet` + `$ErrorActionPreference = "Stop"`.

## 1. Minor: `scripts/mtk.ps1` prints mojibake under Windows PowerShell 5.1 (the version it targets)
The file declares `#requires -Version 5.1`, but it contains em-dash characters (U+2014) on lines **4, 8, 92, 154**. Under Windows PowerShell 5.1 a no-BOM `.ps1` is decoded as ANSI/CP1252, so the em-dash is mangled. Line 154 is inside the `Show-Usage` here-string, so `mtk help` renders:

```
mtk â€" ESS NextGen Migration Toolkit
```

instead of `mtk — ...` (verified live under 5.1). It doesn't crash here — the em-dashes sit in comments/a here-string rather than a token-breaking position — but the user-facing help (and the `.SYNOPSIS` shown by `Get-Help`) is visibly garbled on the exact shell the script supports. This is the same root cause as #185; recommend replacing those four em-dashes with ASCII `-`. (The root `mtk.ps1` forwarder is already ASCII-clean.) `.py`/`.sh` files are unaffected — Python and bash default to UTF-8.

## 2. Observation: toolkit quality gates aren't wired into repo CI
`.github/workflows/ci.yml` lints only `solutions/ess-maker-skills/*`, so the new `tools/ess-nextgen-migration-toolkit/` `ruff`/`mypy`/`pytest` run only via local `pre-commit` and won't gate PRs. Reasonable to defer for a scaffold, but worth a follow-up so the gates actually enforce on incoming task-stream PRs.

tools/ess-nextgen-migration-toolkit/src/core/auth/token_provider.py
anil-adepu: Intentional fail-fast design. A near-expiry token that can't be refreshed indicates a broken auth. path. Returning a token that will expire in seconds just defers the failure to the Dataverse call, making the root cause harder to debug. The invariant is - "if we can't guarantee a usable token window, we surface the problem immediately".

tools/ess-nextgen-migration-toolkit/scripts/mtk.sh
anil-adepu: This is the standard uv install pattern from Astral's docs. Adding checksum verification would require pinning to a specific version and maintaining the hash - defeating the point of auto-bootstrapping the latest stable uv. The trust assumption (HTTPS to astral.sh) is the same as any `pip install` or `npm install`.

tools/ess-nextgen-migration-toolkit/pyproject.toml
anil-adepu: Updated

tools/ess-nextgen-migration-toolkit/scripts/mtk.ps1
anil-adepu: I'm making this change

tools/ess-nextgen-migration-toolkit/src/core/auth/token_provider.py
anil-adepu: MSAL's public API contract guarantees `expires_in` in successful token responses. A response with `access_token` but no expiry metadata would violate the MSAL/OAuth2 spec - defensive coding against spec-violating identity providers is out of scope for the initial scaffold. If we encounter this in practice, we'll add the fallback then.

tools/ess-nextgen-migration-toolkit/scripts/mtk.sh
anil-adepu: `exec` is intentional - `cmd_run` is always terminal (called as the last step of `cmd_start`). No future commands compose after it. If we add a `test` command, it would get its own function that invokes `uv run pytest` directly without going through `cmd_run`. Process replacement avoids a redundant parent shell hanging around

tools/ess-nextgen-migration-toolkit/scripts/mtk.sh
anil-adepu: Aligned. Updated as a TODO on user.

tools/ess-nextgen-migration-toolkit/.pre-commit-config.yaml
anil-adepu: pre-commit always runs hooks from the root of your repository. Closing as no-op.

tools/ess-nextgen-migration-toolkit/scripts/mtk.ps1
anil-adepu: The script is invoked via `& scripts/mtk.ps1` (child scope) from the root forwarder — dot-sourcing is not a supported invocation path and isn't documented anywhere. Adding Push/Pop-Location wrapping adds complexity to guard against an unsupported usage pattern. If someone dot-sources a script that wasn't designed for it, the cwd side-effect is the least of their problems.

amilandi (approved):

PR #178, opened 2026-07-01
spec(mtk): [Master] | super-pipeline + two-file session bundle redesign

anil-adepu:
Rework the ESS NextGen Migration Toolkit specs around a fluent super-pipeline and a two-file session-bundle output.

- Pipeline framework (PIPELINES.md v2.0): the toolkit is a fluent super-pipeline of three stage pipelines (Input -> Migration -> Output) over a shared, typed MigrationContext; generic Pipeline[TIn,TOut] / PipelineStep[TIn,TOut] with a type-threading Builder; the Migration Orchestrator is only the composition root. Add PIPE-008/009/010 and the keyed ComponentSet.
- Diagnostics (DIAGNOSTICS.md): every run produces exactly one session bundle output/session-<timestamp>/ with two files: migration_report.md (customer) and session.log (ESS engineer). Steps accumulate into MigrationContext collectors; a terminal GenerateMigrationReport() renders via the Reporter service. Dropped the optional support.json sidecar entirely.
- INVARIANTS: add PIPE-007 (typed super-pipeline, orchestrator-as-composition- root) and DIAG-004 (exactly one session bundle = two files).
- Rename debug/ -> output/ across specs, tasks, instructions mirror, and the physical scaffold (.gitignore, README, output/.gitkeep).
- Update DOMAIN_MODEL (MigrationContext collectors + ComponentSet), REPOSITORY_STRUCTURE, IMPLEMENTATION_GUIDE, TESTING, AGENTS, CHANGELOG, and TASK-001/002/003/005/006/007.

## Type of change
- [x] Documentation update
- [x] Refactor / cleanup

## Checklist

```text
Validation
- YAML parse: PASS|FAIL|N-A
- AdaptiveDialog kind: PASS|FAIL|N-A
- XML parse: PASS|FAIL|N-A
- Filename convention (new): PASS|FAIL|N-A
- Folder convention (new, incl. README.md): PASS|FAIL|N-A
- Diff scope (samples/ only): PASS|FAIL|N-A
- Secrets / internal URLs: PASS|FAIL|N-A
```

### DIFF
--- a/dev-specs/ess-nextgen-migration-toolkit/02_ARCHITECTURE/PIPELINES.md
+++ b/dev-specs/ess-nextgen-migration-toolkit/02_ARCHITECTURE/PIPELINES.md
@@ -1,7 +1,7 @@
 # PIPELINES.md
 
 # ESS NextGen Migration Toolkit — Pipeline Framework Specification
-**Status:** Draft v1.0
+**Status:** Draft v2.0
 **Owner:** Anil Kumar Adepu
 
 > **Purpose**
@@ -10,9 +10,57 @@
 >
 > The Pipeline Framework provides a deterministic, extensible execution engine for migration workflows.
 >
-> Business transformations are implemented as independent Pipeline Steps that are composed into executable Pipelines.
+> Business transformations are implemented as independent, strongly-typed Pipeline Steps that are composed into executable Pipelines.
 >
 > The framework itself is completely independent of ESS-specific migration logic.
+
+> **The toolkit is a super-pipeline (v2.0)**
+>
+> The product itself *is* a single deterministic, fluent super-pipeline composed
+> of three independently-extensible **stage pipelines**, each responsible for a
+> single concern:
+>
+> ```
+> Input Pipeline      (src/modules/preprocessing/)
+>       ↓
+> Migration Pipeline  (src/modules/migration/)
+>       ↓
+> Output Pipeline     (src/modules/postprocessing/)
+> ```
+>
+> The composition is expressed fluently:
+>
+> ```python
+> EssMigrationToolkit()
+>     .input(InputPipeline().use(...).use(...))
+>     .migrate(MigrationPipeline().use(...).use(...))
+>     .output(OutputPipeline().use(...).use(...))
+> ```
+>
+> **Each stage receives the output of the previous stage and operates over the
+> shared `MigrationContext`.** The Input Pipeline builds and enriches the
+> context (including the homogeneous keyed `ComponentSet` of customer-owned
+> components); the Migration Pipeline applies deterministic business
+> transformations to it; the Output Pipeline validates, persists, and reports on
+> it.
+>
+> **The Migration Orchestrator is only the composition root.** It builds the
+> super-pipeline, configures the execution mode (Discover / Preview / Migrate),
+> executes it, and returns the resulting reports and diagnostics. Orchestration
+> concerns are kept strictly separate from pipeline behaviour — the orchestrator
+> is *not* the primary abstraction; the super-pipeline is.
+>
+> **Typed framework foundation.** The reusable framework is generic —
+> `Pipeline[TInput, TOutput]` and `PipelineStep[TInput, TOutput]` — so the
+> Builder can type-thread steps and support type-changing steps where genuinely
+> needed. The three ESS stage pipelines instantiate this foundation over the
+> shared `MigrationContext` (`Pipeline[MigrationContext, MigrationContext]`),
+> which threads through the whole super-pipeline. (The generic `TInput, TOutput`
+> signature is the analogue of the C# `HeterogenousPipelineStepComputeUnitBase
+> <TInput, TOutput>`; a stage pipeline is the analogue of
+> `KeyedComputeUnitBase<...>`. C# runtime concerns — Autofac keyed registration,
+> `floorEpoch`/`currentEpoch` versioning, per-step `Equals`/`GetHashCode` — are
+> intentionally out of scope and not ported.)
 
 ---
 
@@ -81,6 +129,42 @@ Execution Mode selection belongs to Pipeline Steps, not the Pipeline Engine.
 Pipeline execution is fail-fast.
 
 A failed step terminates the pipeline unless explicitly configured otherwise.
+
+---
+
+## PIPE-008
+
+Pipelines and Pipeline Steps are strongly typed.
+
+The framework foundation is generic — `Pipeline[TInput, TOutput]` and
+`PipelineStep[TInput, TOutput]`. The Pipeline Builder type-threads adjacent
+steps so that the output type of one step is the input type of the next; an
+incompatible composition is a construction-time error, not a runtime error. The
+three ESS stage pipelines instantiate this foundation over the shared
+`MigrationContext`.
+
+---
+
+## PIPE-009
+
+A stage pipeline threads the shared `MigrationContext`.
+
+Within a stage, each step receives the `MigrationContext` produced by the prior
+step, enriches it, and passes it on. A step may Read, Enrich, or Validate the
+context; a step may never replace the context with an unrelated type nor smuggle
+state outside it. The context is the only object shared between steps.
+
+---
+
+## PIPE-010
+
+The toolkit is a super-pipeline of three stages.
+
+Migration executes as **Input → Migration → Output**, each a stage pipeline over
+the shared `MigrationContext`. The Migration Orchestrator is only the
+composition root: it builds the super-pipeline, configures the execution mode,
+executes it, and returns reports and diagnostics. A stage never reaches into
+another stage's steps.
 
 ---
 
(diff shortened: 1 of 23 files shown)


anil-adepu:

dev-specs/ess-nextgen-migration-toolkit/03_ENGINEERING/DIAGNOSTICS.md
```diff
@@ -132,59 +132,130 @@ Close Session
...
+`print()` from business logic is prohibited (DIAG-005). The Logger is
+initialized once at the **application entry point** (the composition root, before
+the pipeline runs) and is the single I/O boundary for all console and log output.
 
-The Logger writes simultaneously to:
+## 6.1 Transcript capture (session.log is a full CLI replay)
 
-* Console
-* Session Log
+When the Logger initializes, it **installs a stdout/stderr tee** for the process.
```
aditandadit, line 202: [Question] Since session.log captures every byte on stdout/stderr and it's the file customers forward, could it end up capturing secrets? TASK-008 is a token provider, and auth stacks often log tokens/PII at DEBUG. Do we want a redaction pass on the tee, or to classify session.log as internal-only? Is there already a scrubbing story I'm missing?

dev-specs/ess-nextgen-migration-toolkit/03_ENGINEERING/DIAGNOSTICS.md
```diff
@@ -132,59 +132,130 @@ Close Session
...
+  single blended stream.
 
-is prohibited.
+## 6.3 Report model (intermediate)
+
+Customer-channel calls accumulate into a structured, in-memory **report model**
+(the `MigrationContext` collectors — `Changes`, `Warnings`, `Errors`, plus
+summary counters; see DOMAIN_MODEL.md). The Reporter (section 9) renders this
+model into `migration_report.md` at the end of the run. Persisting the model as
+an intermediate `report.json` is permitted internally as a render input, but it
```
aditandadit, line 249: [Suggestion Non Blocking] Where's the report.json expected to live? If it can land inside output/session-<timestamp>/, that quietly breaks the DIAG-004 "exactly two files" rule. Could we add a line saying it must be written outside the session folder? Just saving the implementer from an easy trap.

dev-specs/ess-nextgen-migration-toolkit/02_ARCHITECTURE/PIPELINES.md
```diff
@@ -183,35 +267,85 @@ Construct executable pipelines.
...
 
 ## Fluent API
 
-```python
-pipeline = (
-    Pipeline.builder("Migration")
-
-        .use(DiscoverComponents())
+Each stage pipeline is built fluently over the shared `MigrationContext`. The
+generic Builder (`Pipeline.builder(name)` → `Pipeline[TInput, TOutput]`)
```
aditandadit: PIPELINES.md L278 shows Pipeline.builder(name), but TASK-003 L82 calls Pipeline.builder("input", input_type=MigrationContext). Is input_type a param we're keeping? If so, could we document it in PIPELINES.md so TASK-002 implements it?

dev-specs/ess-nextgen-migration-toolkit/03_ENGINEERING/DIAGNOSTICS.md
```diff
@@ -132,59 +132,130 @@ Close Session
...
 
-        migration_report.txt
-```
+Especially in **Discover** and **Preview** modes, `migration_report.md` *is* the
+deliverable. `session.log` is what the customer forwards (or a support engineer
+retrieves) when something needs investigation. Two files, one folder — nothing
+scattered across the filesystem.
 
-The timestamp format shall be:
+The session folder name is `session-<timestamp>`, where the timestamp format is:
```
aditandadit, line 160: Folder name is session-<timestamp> with YYYY-MM-DD_HH-MM-SS (L160–163), second resolution. 

Do we need uniqueness? might be overkill but just a thought.Any scenario where two runs start in the same second — parallel golden/e2e tests, or scripted back-to-back? If so we could collide/overwrite, which works against the determinism + golden invariants.

aditandadit (approved):

dev-specs/ess-nextgen-migration-toolkit/03_ENGINEERING/DIAGNOSTICS.md
anil-adepu: Good question. session.log is the raw stdout/stderr tee - it captures exactly what the user sees on screen. Token provider output goes through the framework Logger which is configured at INFO by default; MSAL's internal DEBUG logging (where tokens/PII could appear) is never surfaced at INFO. The Logger explicitly uses `AuthenticationException` with sanitized messages (no provider error details - tested by `test_acquisition_failure_raises_sanitized_authentication_exception`). A post-hoc redaction pass would add complexity for a risk that's already mitigated by the log-level boundary. If we add DEBUG-mode support later, we'd add redaction then - but session.log at INFO is safe to share.

dev-specs/ess-nextgen-migration-toolkit/02_ARCHITECTURE/PIPELINES.md
anil-adepu: Updated the documentation

dev-specs/ess-nextgen-migration-toolkit/03_ENGINEERING/DIAGNOSTICS.md
anil-adepu: `report.json` isn't a separate output file - section 9 describes the Reporter rendering model, not a new artifact. The Reporter generates migration_report.md (the customer facing artifact inside the session bundle). DIAG-004's "exactly two files" rule applies to the session bundle (migration_report.md + session.log), and the Reporter produces one of those two. No conflict.

dev-specs/ess-nextgen-migration-toolkit/03_ENGINEERING/DIAGNOSTICS.md
anil-adepu: IMO, Second-resolution is fine for now - the toolkit runs one session at a time (single CLI invocation, no daemon). Parallel golden/e2e tests each get their own tmp_path fixture (pytest provides unique temporary directories), so they never write to output/. If we ever add parallel session support, we'd add a short random suffix then - but solving for a non-existent concurrency model now adds complexity without value.

dev-specs/ess-nextgen-migration-toolkit/04_EXECUTION/TASKS.md
```diff
@@ -126,13 +124,13 @@ Instead, it establishes the complete framework, wiring, and developer experience
 | Task                                                     | Title                          | Status | Consumes |
 | -------------------------------------------------------- | ------------------------------ | ------ | -------- |
 | [TASK-001](tasks/TASK-001-repository-scaffold.md)        | Repository Scaffold            | DONE   | —        |
-| [TASK-002](tasks/TASK-002-pipeline-framework.md)         | Pipeline Framework             | TODO   | —        |
+| [TASK-002](tasks/TASK-002-pipeline-framework.md)         | Pipeline Framework             | ACTIVE | —        |
 | [TASK-003](tasks/TASK-003-migration-orchestrator.md)     | Migration Orchestrator         | TODO   | —        |
```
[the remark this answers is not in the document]
anil-adepu: updated documentation

dev-specs/ess-nextgen-migration-toolkit/04_EXECUTION/CHANGELOG.md
```diff
@@ -12,12 +12,44 @@ task (`TASK-XXX`) where applicable, per `IMPLEMENTATION_GUIDE.md`.
...
+  `migration_report.md` (customer-facing; summary + changes + warnings sections)
+  and `session.log` (ESS-engineer diagnostics). Steps accumulate into
+  `MigrationContext` collectors (`Logs`, `Warnings`, `Errors`, `Changes`); a
+  terminal `GenerateMigrationReport()` step renders the report via the Reporter
+  service (no direct file I/O in steps). Added `DIAG-004`.
+- **`debug/` → `output/` rename.** Replaced the `debug/logs` + `debug/reports`
+  split with the single `output/session-<timestamp>/` bundle across specs
+  (REPOSITORY_STRUCTURE, IMPLEMENTATION_GUIDE, TESTING, AGENTS, TASK-001/005/007)
+  and the physical scaffold (`tools/…/output/.gitkeep`, `.gitignore`, README,
+  toolkit `AGENTS.md`). Updated the instructions mirror.
```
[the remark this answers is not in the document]
anil-adepu: updated documentation

dev-specs/ess-nextgen-migration-toolkit/04_EXECUTION/tasks/TASK-005-diagnostics-framework.md
```diff
@@ -4,32 +4,62 @@
...
+2. **Two channels** — the Logger exposes an **engineer channel**
+   (`LogDebug`/`LogInfo`/`LogWarning`/`LogError`) that prints to the CLI as usual
+   (and is mirrored to `session.log` by the tee), and a **customer channel**
+   (`LogChange`/`LogAdvisory`) that does **not** touch the CLI or `session.log`
+   but instead appends structured entries to the report model (the
+   `MigrationContext` collectors) that the Reporter later renders into the fancy
+   `migration_report.md`. `LogChange` records a successful transformation
+   (`ChangeEntry` → `context.Changes` → `## Changes`); `LogAdvisory` records a
+   manual-review advisory (`DiagnosticEntry` → `context.Warnings`/`Errors`/`Logs`
+   by `severity` → `## Warnings — Manual Review Required`).
```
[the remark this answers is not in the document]
anil-adepu: moot now

PR #181, opened 2026-07-06
# Workday `/connect` setup orchestrator: 6 skills, shared components, and FlightCheck coverage

apurvabanka:
## Summary
Replaces the old 3-step Workday connect runbook with a **6-skill setup
orchestrator** driven by a single checklist and gated by FlightCheck
checkpoints. `/connect workday` now routes into `setup/`, which walks the maker
from an empty tenant to a working, custom Workday topic. Standardizes on the
**simplified Entra-SSO** extension pack (no ISU/Basic-auth service accounts).

## Setup skills (`src/skills/setup/`)
New `SKILL.md` router plus six Workday skills:
1. **Provision the Power Platform environment** (Dataverse)
2. **Install the Employee Self-Service base agent**
3. **Provision the Workday Entra SSO app** (SAML)
4. **Configure the Workday tenant** (API client, connection fields)
5. **Install & wire the Workday extension pack** (connections, REST URL, flows, user-context redirect, firewall)
6. **Author the first custom Workday topic** (Template Config + Shared Flow, SME-attested reference IDs, auto-generated eval set)

Single canonical `tasks.md` checklist with `prog` / `manual` / `attest` gates and
per-row checkpoint IDs.

## Command wiring
- `connect/SKILL.md` + `connect/step1.md`: Workday branch now delegates to the
  setup orchestrator.
- Removed the legacy `connect/workday/{step1,step2,step3,tasks}.md` runbook (~2k lines).
- Generalized `connect/azure/app-registration.md` (§B.4/§B.5) so Workday and
  ServiceNow share scope-expose / connector pre-auth logic.

## Shared components (`src/skills/setup/shared/`)
- `checklist-updater.md` — one place that enforces the manual/attest gate rule.
- `config-schema.md` — durable `.local/` config contract.
- `connection-fields.md` — capture/validate/trim Workday connection values.
- `permission-gate.md` — reusable role gate for each skill.

## FlightCheck (additional checks + infra)
- New checks: `entra_app.py`, `_workday_app_assignment.py`, `_saml_utils.py`,
  `workday_tenant.py`, `workday_extension.py`, `topics.py`, `solution.py`,
  `licensing.py`, plus `environment.py` additions.
- Refactored `authentication.py` (Entra logic split into `entra_app` / `_workday_app_assignment`).
- New `registry.py` checkpoint registry + `cli.py` selection/wiring and `runner.py` updates.

## Reference docs & CI
- `validation-matrix.md`, `role-gating.md`, `remediation-guide.md` updated for the new checkpoints.
- `ci.yml`: lint scoping for the new paths.

## Tests
New coverage for the setup router, checklist template, and every new checkpoint
family (`test_entra_app`, `test_workday_tenant`, `test_workday_extension`,
`test_topics`, `test_solution`, `test_env_capacity`, `test_registry`,
`test_registry_drift`, `test_workday_saml_certificate`) plus expanded Graph mocks.


apurvabanka:
@copilot resolve the merge conflicts in this pull request

amilandi:
### Telemetry: invocation source for the `/connect` setup incremental FlightChecks

Reviewing the FlightCheck telemetry angle here, since `/connect` now runs a per-checkpoint FlightCheck under the hood before advancing each step. I dug into how invocation source flows and want to flag a design decision before this merges. One clarification up front that changes the framing:

**Today, the incremental per-checkpoint runs emit *no* telemetry at all.** All 33 setup call sites invoke `python scripts/flightcheck/cli.py --checkpoint <ID>`, which routes to `_run_single_checkpoint()`. That path prints the summary, saves results, and `sys.exit()`s — it never reaches an `emit_*` call. Telemetry is only emitted on the `--scope` path (`cli.py` ~641-662), and `runner.py` doesn't emit either. So there isn't a flood of `cli`-attributed data flowing in from `/connect` right now — if anything, the setup capability's FlightCheck activity is currently **invisible** to telemetry.

Worth separating this from **scoped subset runs** (e.g. `--scope workday`, `--scope servicenow`), which are a different and already-instrumented path: those route through the normal `main` flow, so they *do* emit telemetry today with the `scope` value and an aggregate verdict. Users running a subset of checks that way are already captured on the dashboards. The gap is specifically the single-checkpoint (`--checkpoint <ID>`) mode that `/connect` drives.

That turns this into a "what do we want" decision rather than a bug:

**1. Do we want per-checkpoint setup runs to emit telemetry?** I'd argue yes — knowing *which* setup steps block makers, how often, and how long they take is exactly the kind of thing the FlightCheck dashboards should show. If we agree, we'd add an emit block to `_run_single_checkpoint` mirroring the scope-mode one (still gated by `--no-telemetry`, `scope=f"checkpoint:{target}"`).

**2. If we emit, they must NOT be attributed to `cli` (or `adk`).** The current enum is `["adk", "installer", "cli"]` (`cli.py:398`). `cli` should mean a human running the validator directly, and `adk` is the full `/flightcheck` slash-command run. The setup-driven single-checkpoint runs are a third thing: agent-orchestrated, single-row, and *high frequency* (33 sites, fired repeatedly across a setup session). If they land as `cli` they'll swamp the real direct-CLI bucket; if they land as `adk` they'll skew the full-run pass-rate and p50/p95 duration tiles with single-checkpoint rows. A distinct source lets the dashboards include/exclude them deliberately.

**3. On the name** — I'd steer away from a Workday-specific value like `connect-workday`. The same orchestrator (`src/skills/setup/`, `/connect`) already drives ServiceNow and will drive future connectors, so a connector-specific source fragments the taxonomy and forces a new enum value per connector. I'd recommend a connector-agnostic, short, lowercase token consistent with the existing enum: **`connect`**, keyed off the `/connect` entry point that drives these runs. (I'd avoid `setup`, since that reads as coming from the separate `/setup` capability, which isn't the trigger here.) If per-connector breakdown is ever needed, that's better captured by the existing `scope`/checkpoint fields than by exploding the invocation-source dimension.

**Suggested implementation (low-churn):**
- Add the new value to `choices` at `cli.py:399`.
- Don't append `--invocation-source connect` to all 33 markdown snippets — that's drift-prone (and `test_registry_drift.py` already guards this surface). Instead default the arg from an env var, e.g. `default=os.environ.get("ESS_FLIGHTCHECK_INVOCATION_SOURCE", "cli")`, and have the `/connect` orchestrator export it once per session. One change, no per-call-site edits.
- Add the emit to `_run_single_checkpoint` (if we go with #1).
- Update the `invocationSource` allowed-values wherever they're enumerated (Aria cube dimension + `flightcheck/validation-matrix.md` / telemetry docs) so the new source is recognized downstream.

Happy to pair on the emit + env-var default if useful. Nothing else in the telemetry path looks off — the `/flightcheck` and onboarding skills correctly pass `--invocation-source adk`.

amilandi:
Tracked the telemetry follow-up in ADO: **[7587431 — FlightCheck telemetry: add `connect` invocation source + emit telemetry for single-checkpoint (`/connect`) runs](https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7587431)**.

**No change needed on your end in this PR** — the `/connect` single-checkpoint runs are telemetry-silent today, so nothing here regresses telemetry, and wiring up the new `connect` source + emission is a separate change we'll pick up under that work item. Ship as-is.

amilandi:
## Holistic review

Following up my telemetry-specific comments above with a broader pass over the whole PR (the executable FlightCheck Python, the registry/orchestrator contract, and the tests). Verification I ran locally against `feature/worlday-ask-skills`: read the check modules + `registry.py`/`runner.py`/`cli.py`, statically resolved every `--checkpoint` ID referenced in the setup skill markdown against the registry, and ran the suite — **822 passed**.

**Overall: solid, ship-ready.** No blocking or should-fix issues in the executable code. The checks follow the established never-raise dispatcher pattern with consistent `_status`/`_error` sentinel handling and config/None guards, and the markdown↔registry checkpoint-ID contract is intact from both directions (`test_registry_drift.py` guards code→registry; I independently confirmed all 18 setup-markdown IDs — `WD-*`, `ENV-CAPACITY-001`, `ESS-SOLN-001`, `DV-CONN-001`, and the `TOPIC-*`/`WD-FLOW-*` families — resolve, so no runtime exit-2 risk).

### Architecture
The orchestrator/registry/checks structure hangs together well. The registry is a clean static source of truth: transitive-closure prerequisite resolution, canonical category ordering for cross-category state hydration (External Systems → Workday → Workday Extension), exact-before-longest-prefix-family matching, and an import-time `validate_registry()` that proves the prereq graph is resolvable and acyclic. Nice systemic choices: uniform never-raise emitters, **MANUAL degradation instead of confidently-wrong FAILs** where an API is unreadable (NameID `_status` probe, capacity `unreadable`, portal-only Workday-tenant checks), and genuine de-duplication of drift-prone logic into shared helpers (`_saml_utils`, `_workday_app_assignment`, `classify_copilot_studio_capacity`).

### Minor (non-blocking)
1. **Test-coverage gap on the core new execution path.** `registry.resolve/matches/transitive_requirements` are exhaustively tested, but the layers that *consume* them are not: `FlightCheckRunner`'s `target_matcher` result-filter incl. the `-ERR` sentinel-retention (`runner.py:156-166`) and the whole `cli._run_single_checkpoint` orchestration (client-init gating, hydrate-then-filter, exit codes) have **zero** direct tests. This is the most defect-prone new surface and the only new executable code without a guard. Suggest one test that registers a fake category emitting several IDs, runs with a `target_matcher`, and asserts only matching rows (+ any `-ERR` sentinel) survive, plus an exit-code assertion for pass/fail/unknown-ID.
2. **`"Solution"` is missing from `registry.CATEGORY_ORDER`** (`registry.py:76-93`). Every other setup category label is present; `_order_index()` falls back to `len(CATEGORY_ORDER)` for unknown labels, so `run_solution_checks` sorts *last* in registry-driven ordering — inconsistent with `cli.py FULL_SCOPE`, which places Solution right after Environment. Harmless today (ESS-SOLN-001's only prereqs are Environment checks and nothing consumes Solution's shared state), but a latent divergence between the two orderings. Add `"Solution"` after `"Environment"` to keep them in lockstep.
3. **Stale test citation** (doc-only). The NOTE in `_workday_app_assignment.py:466` names its regression guard as living in `tests/flightcheck/test_runner.py`, but it actually lives in `tests/flightcheck/test_runner_manual_status.py`. The guard exists and passes — only the cited path is wrong.

### Note
The telemetry angle (single-checkpoint `/connect` runs emit nothing; new `connect` invocation source) is intentionally **not** repeated here — it's covered in my comments above and tracked in [ADO 7587431](https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7587431), and requires **no change in this PR**.

amilandi (dismissed):

apurvabanka:
@copilot resolve the merge conflicts in this pull request

daeunJe0ng:
**[copilot-review]** Reviewed at HEAD `5d180ea`. Verdict: **Approve with nits** — the executable code is solid, the design is coherent, and `tests/flightcheck` is green (842 passed). No blocking issues.

**One new item (not raised yet):**

- **CI never runs the FlightCheck suite.** `ci.yml` runs only ruff + compileall + the installer smoke test, and the sole pytest job (`validate-samples.yml`) runs just `tests/test_validate_samples.py`. So the 842-test suite — including `test_registry_drift.py`, the guard that protects the markdown↔registry checkpoint-ID contract this PR relies on — isn't enforced anywhere. It's pre-existing, but since this PR both adds that contract-critical suite and already edits `ci.yml`, adding a job that runs `tests/flightcheck` here is low-cost and high-value. Otherwise a future checkpoint-ID drift can merge green.

**Confirming @amilandi's three minor items still apply at `5d180ea`:**

1. Single-checkpoint path is untested — `cli._run_single_checkpoint` and `runner`'s `target_matcher` (incl. `-ERR` sentinel retention) have no direct tests. Most defect-prone new surface.
2. `"Solution"` is missing from `registry.CATEGORY_ORDER` (registry.py:78-95) though it's registered as a `category_label` (registry.py:197), so it sorts last — diverging from `cli.py` `FULL_SCOPE` which puts it right after Environment. Nit.
3. Stale citation: `_workday_app_assignment.py:465` points to `tests/flightcheck/test_runner.py` (doesn't exist); the guard actually lives in `test_runner_manual_status.py:123`. Doc-only.

**Telemetry:** agree with shipping as-is — the single-checkpoint `/connect` runs emit nothing today, so there's no regression, and the new `connect` source is tracked in ADO 7587431.

**Verification:** cloned HEAD; `tests/flightcheck` = 842 passed. Full tree = 1201 passed / 10 failed; the 10 failures are pre-existing installer-script tests (scripts are byte-identical to `main` and the same 10 fail on a clean `main` checkout), not introduced by this PR.

daeunJe0ng (dismissed):

daeunJe0ng (approved):
**[copilot-review]** Re-reviewed at HEAD `ebbaf69`. Verdict: **Approve** — the one new commit since `5d180ea` resolves all three minor items I raised. `tests/flightcheck` is green (854 passed).

**Resolved in `ebbaf69`:**

1. **Single-checkpoint path now tested** (was the most defect-prone untested surface). `test_runner_single_checkpoint.py` covers `target_matcher` family-vs-exact filtering, `-ERR` sentinel retention when unmatched, sentinel shape/status/priority, and `NOT_READY` vs `failed` semantics; `test_cli_single_checkpoint.py` covers `_run_single_checkpoint` gates, unknown-ID handling, and hermetic run/exit-code paths.
2. **`"Solution"` added to `registry.CATEGORY_ORDER`** right after `Environment`, matching `cli.py` `FULL_SCOPE` ordering. The two orderings are now in lockstep.
3. **Stale citation fixed** in `_workday_app_assignment.py`: now points to `test_runner_manual_status.py` (confirmed the guard `test_html_report_preserves_multiple_results_with_same_checkpoint_id` lives there).

**Verification:** ran `tests/flightcheck` on `ebbaf69` in an isolated worktree = **854 passed** (up from 842 at `5d180ea`, +12 new tests). Diff-read all 4 changed files; fixes are surgical and correct.

**Still open (non-blocking, pre-existing):** CI does not run the FlightCheck suite, so `test_registry_drift.py` (the markdown-to-registry checkpoint-ID guard) isn't enforced anywhere. Not introduced here; reasonable as a fast-follow.

Approving.

PR #182, opened 2026-07-07
ADK telemetry: derived tenant_class (internal vs customer) + capability instrumentation

amilandi:
## Summary

Adds a derived **`tenant_class`** (`internal` | `customer` | `unknown`) dimension to ADK telemetry so dashboards can report Microsoft-internal dogfood/testing traffic separately from real external customer usage.

This branch also carries the earlier capability-instrumentation work (ADO 7570789): every ADK capability now emits `capability.use`, session-level capability tagging was dropped in favor of per-capability events, and the non-capability `onboarding` tag was removed (`discover` / `list-environments` are sub-steps of `setup`, not a distinct capability).

## What changed (tenant_class — ADO 7558661)

- **`classify_tenant()`** + an internal-tenant allow-list added to `flightcheck/telemetry.py` (the lowest-level module; `adk_telemetry` reuses it as `_fc.classify_tenant`).
  - Seeded with the **Microsoft corporate Entra tenant only** (`72f988bf-…`).
  - Extendable **without a code change** via `ESS_ADK_INTERNAL_TENANTS` (comma-separated GUIDs).
  - Classification happens at **emit time**, not in the Aria cube: 1DS RTA cubes map an event property straight to a dimension and cannot derive one dimension's value from another.
- Stamps `tenant_class` on every `adk.*` event (`common_dimensions`) and `tenantClass` on FlightCheck `Run`/`Check` events (`_run_data`/`_check_data`). Schema versions bumped.
- Synthetic seed pool now includes the corp tenant so the internal/customer split is visible on dashboards.
- Documents the derived `tenant_class` field in the **Telemetry & Privacy** README (non-identifying, lower sensitivity than the tenant ID it is derived from).

## Capability instrumentation (ADO 7570789)

- Every ADK capability emits an `adk.capability.use` event; session-level capability tagging was replaced with per-capability events.
- Wired the `connect` skill (ServiceNow / Workday setup) to emit `connect` via the `emit_capability.py` shim, like the other SKILL.md-driven skills, so it is a real dashboard slice rather than synthetic-only.
- Dropped the `onboarding` capability tag: it did not map to an actual capability and inflated the "Capability Usage by Type" breakdown. `discover` and `list-environments` are now attributed to `setup`, of which they are sub-steps.

## Tests

- `classify_tenant`: internal / customer / unknown, case- & whitespace-insensitive, env allow-list extension.
- `tenant_class` present and correct in both `adk.*` and FlightCheck event dimensions.
- Capability taxonomy updated (no `onboarding`; `discover`/`list-environments` → `setup`; `connect` wired via the shim).
- Targeted suite: `tests/test_adk_telemetry.py` + `tests/flightcheck/test_telemetry.py` — **54 passed**.

## Dashboards

The Aria `tenant_class` cube dimension + internal/customer split on the Usage / Performance / FlightCheck panels (dev + prod) are applied directly via the Aria document REST API (not in this repo).

ADO: https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7558661
ADO: https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7570789


daeunJe0ng (commented):
Nice work. This is a clean, well-scoped change and the test coverage is strong. The `tenant_class` derivation at emit time (with the env-extensible allow-list) and the `emit_capability.py` shim are both sensible, and the fail-open contract is well tested. `tenant_class` reads as genuinely lower sensitivity than the raw tenant ID already emitted, and it is documented. No blocking concerns from me.

Two things I would like to confirm before approving:

1. **Standalone `discover` / `list-environments` now emit nothing.** They previously emitted `onboarding`, and the PR removes that on the basis that they are sub-steps of `setup`. But both are independent CLI entry points with their own argparse, so a maker who runs `/discover` outside a full setup flow will no longer register any capability usage. Is that intended? If we still want those standalone invocations visible, emitting `setup` there would keep them counted.

2. **Possible `evaluations` double-count.** The evaluate path is already Python-wired to emit `evaluations`, and the evaluations create/update/delete `SKILL.md` files now also emit `evaluations` via the shim. In a single author-then-run flow that could fire twice. If the dashboard counts touches this is fine, but if it is meant to approximate distinct sessions it could inflate. Can you confirm which is intended?

Neither is a hard blocker. Happy to approve once these are clarified. Thanks!

amilandi:
Thanks for the careful review, @daeunJe0ng! Both are good questions — here's the reasoning:

**1. Standalone `discover` / `list-environments` emitting nothing — intended.**
There is no standalone `/discover` (or `/list-environments`) command. The menu (`.github/prompts/menu.prompt.md`) exposes: `/connect`, `/create`, `/update`, `/delete`, `/scan`, `/evaluate`, `/flightcheck`, `/push`, `/backup-template-configs`, `/restore-template-configs`, `/troubleshoot`, plus `/setup`. `discover.py` and `list_environments.py` are invoked **only** as sub-steps of the `/setup` → onboarding flow:
- `onboarding/step1.md` → `discover.py --list-environments`
- `onboarding/step1b.md` → `discover.py --url`
- `onboarding/step2.md` → `fetch_and_setup.py`, which emits exactly one `setup`

So there is no path where a maker runs discovery outside setup. If we emitted `setup` from `discover.py`/`list_environments.py`, every normal onboarding would fire `setup` 2–3× (list-env + discover + fetch_and_setup), inflating the setup count on the common path — which is the same over-counting problem the dropped `onboarding` slice had. The single `setup` emit in `fetch_and_setup` already represents one setup per run, so I've left discovery uninstrumented intentionally.

**2. `evaluations` firing on both author and run — intended.**
`Capability Usage by Type` counts capability *touches*, not deduplicated sessions. Authoring an eval set (create/update/delete `SKILL.md` → `emit_capability.py evaluations`) and running evals (`evaluate_evals.py` → `emit_capability_use("evaluations")`) are two distinct maker actions, so counting them separately is correct for a usage/touches metric. It's not meant to approximate distinct sessions, so there's no inflation concern here.

Happy to add a dedicated standalone `discover` capability later if we ever expose `/discover` as its own command. Thanks again!

solutions/ess-maker-skills/scripts/emit_capability.py
```diff
@@ -0,0 +1,92 @@
...
+        return 0
+
+    capability = args[0].strip()
+
+    try:
+        # block=True: short-lived CLI process — emit synchronously so the
+        # event isn't dropped when the interpreter exits and kills a daemon
+        # thread. emit_capability_use() no-ops when telemetry is disabled.
+        adk_telemetry.emit_capability_use(capability, block=True)
+        adk_telemetry.flush(timeout=3)
```
apurvabanka: emit_capability_use(..., block=True) already goes through _emit_sync, which posts synchronously and flushes the on-disk buffer (_buffer_flush). The extra adk_telemetry.flush(timeout=3) right after is redundant. The other inline hooks (backup_template_configs.py, restore_template_configs.py, fetch_and_setup.py) call emit_capability_use(block=True) with no flush — so the shim is inconsistent with them. Harmless, but I'd drop the flush line for consistency.

solutions/ess-maker-skills/scripts/flightcheck/telemetry.py
```diff
@@ -90,13 +90,57 @@
...
+
+    Empty / missing tenant -> ``unknown`` (we never guess). A tenant in the
+    internal allow-list -> ``internal``; anything else is an external
+    ``customer``. Case/whitespace-insensitive.
+    """
+    if not tenant_id:
+        return TENANT_CLASS_UNKNOWN
+    return (
+        TENANT_CLASS_INTERNAL
+        if str(tenant_id).strip().lower() in _internal_tenant_ids()
```
apurvabanka, line 152: classify_tenant reads the env var and rebuilds a frozenset on every call (every common_dimensions, every check event, and every synthetic event). Negligible in normal CLI use, but wasteful during synthetic seeding of large n. Consider caching at module load (accepting that env changes then require a restart — reasonable for a CLI).

apurvabanka (changes requested):

amilandi:
Thanks @apurvabanka - addressed both in 29d6af8:

**1. Redundant `flush(timeout=3)` in `emit_capability.py`** - removed. `emit_capability_use(block=True)` takes the synchronous path (`_emit` -> `_emit_sync`), which posts on the calling thread and flushes the on-disk buffer inline. `flush()` only joins the async emit threads, of which the `block=True` path creates none, so the call was a no-op. Dropping it also makes this hook consistent with the other inline emitters (backup/restore, fetch_and_setup).

**2. Rebuilding the internal-tenant frozenset every call in `telemetry.py`** - cached. Rather than a module-load constant (which would ignore `ESS_ADK_INTERNAL_TENANTS` set after import and break `test_classify_tenant_env_allowlist_extends`, which uses `monkeypatch.setenv`), I moved the build into an `@lru_cache`d `_parse_internal_tenant_ids(extra)` keyed on the raw env-var string. `_internal_tenant_ids()` still reads the env var each call (a cheap dict lookup) but only rebuilds the frozenset when the value actually changes - so hot paths avoid the rebuild and runtime env changes are still honored.

54 Python tests + 40 PowerShell installer tests green.

apurvabanka (approved):

PR #183, opened 2026-07-08
Add installer telemetry (start/step/completion) for setup reliability (ADO 7557940)

amilandi:
## Summary

Adds pseudonymous **installer telemetry** so we can measure setup reliability
for the ESS ADK one-shot installers (ADK, ADK-lite, standalone FlightCheck) and
split Microsoft-internal dogfood installs from external customer installs.

Because the installers run PowerShell/bash **before Python is guaranteed to
exist** (installing Python is part of their job), this adds two dependency-light
native emitters that reproduce the 1DS OneCollector envelope + POST used by the
Python SDK — rather than shelling out to Python.

## What it collects

`ESSMakerKit.Installer.Start` / `.Step` / `.Complete` events with:
- installer variant (`adk` | `lite` | `flightcheck`), platform + OS
- random per-install `instance_id` (dedup; **not** identity)
- raw tenant GUID (OII) + derived `tenant_class` (`internal` | `customer` | `unknown`), where available
- step reached / `stepIndex`, completion `outcome` (`success` | `failure` | `cancelled`), failing step, duration
- scrubbed error message / code / category (paths, URLs, emails, GUIDs stripped)

Same privacy model as the ADK/FlightCheck telemetry (no developer/user identity),
same unified opt-out (`ESS_ADK_TELEMETRY=off` or `python scripts/adk_telemetry.py off`),
and the one-time consent notice.

## Design

- **Strictly fail-open.** Every telemetry call swallows its own errors; telemetry
  never changes the installer's behaviour or exit code.
- Installers locate the emitter via `ESS_INSTALL_TELEMETRY_LIB` (set by the
  bootstrap) or the copy shipped in `setup/telemetry/`, and fall back to **no-op
  stubs** if it can't be found.
- Completion is guaranteed: PowerShell wraps the body in `try/catch/finally`
  (success in `finally`, failure in `catch`); bash uses an `EXIT` trap
  (success on exit 0, failure otherwise). Both are idempotent.
- All six bootstraps best-effort download the emitter into their temp dir and
  pass its path via `ESS_INSTALL_TELEMETRY_LIB` (download failure never blocks
  the install).
- **Lite mode is not instrumented** — the lite installer does not emit telemetry
  (guarded in both emitters).

## Files

- `setup/telemetry/install-telemetry.ps1` (Windows), `install-telemetry.sh` (macOS/Linux) — new emitters
- `setup/Install-EssAdk.ps1`, `setup/install-ess-adk.sh` — wiring (load, init, per-step hook, completion)
- `setup/bootstrap*.ps1`, `setup/bootstrap*-mac.sh` (×6) — download emitter + set env var
- `setup/Install-EssAdk.Tests.ps1` — wiring assertions + pure-function tests (38/38 pass)
- `setup/README.md`, `solutions/ess-maker-skills/README.md` — docs

## Testing

- `pwsh -File setup/Install-EssAdk.Tests.ps1` → **38/38 pass** (incl. an
  fi-FI regression test that the OneCollector envelope timestamp stays
  culture-invariant ISO-8601)
- Emitters validated end-to-end against the dev Aria project: PowerShell POST
  204; bash POST 200; `EXIT` trap confirmed to emit a `failure` completion under
  `set -euo pipefail`; step-key mapping verified for every installer step; scrub
  / tenant-class / opt-out / lite-guard verified.

## Notes / assumptions

- `tenant_id` is often empty at install time (before sign-in) → `tenant_class`
  falls back to `unknown`. Expected.
- Only the Microsoft corporate Entra tenant is treated as `internal` by default
  (overridable via `ESS_ADK_INTERNAL_TENANTS`), matching PR #182.
- Dashboards (Aria cubes + installer panels, dev + prod) are being built
  separately via the Aria document REST API.

Stacked on #182 (`amilandin/adk-capability-telemetry`).

ADO: https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7557940

### DIFF
--- a/setup/Install-EssAdk.Tests.ps1
+++ b/setup/Install-EssAdk.Tests.ps1
@@ -206,14 +206,13 @@ Test 'bash emitter exists with a shebang' {
 # Dot-source the PS emitter and exercise its pure (no-network) functions.
 . $psEmitter
 
-Test 'tenant class: the MS corp tenant is internal' {
-    if ((Get-EssTelTenantClass -TenantId '72f988bf-86f1-41af-91ab-2d7cd011db47') -ne 'internal') { throw 'corp tenant not internal' }
-}
-Test 'tenant class: any other tenant is customer' {
-    if ((Get-EssTelTenantClass -TenantId '11111111-1111-1111-1111-111111111111') -ne 'customer') { throw 'other tenant not customer' }
-}
-Test 'tenant class: an empty tenant is unknown' {
-    if ((Get-EssTelTenantClass -TenantId '') -ne 'unknown') { throw 'empty tenant not unknown' }
+Test 'installer telemetry emits no tenant dimension (unknowable pre-sign-in)' {
+    # The installer runs before sign-in, so the tenant is never known; emitting
+    # tenantId/tenantClass produced a meaningless 'unknown' on ~every event.
+    $data = Get-EssTelCommonData
+    if ($data.ContainsKey('tenantId'))    { throw 'tenantId should not be emitted by the installer' }
+    if ($data.ContainsKey('tenantClass')) { throw 'tenantClass should not be emitted by the installer' }
+    if (Get-Command -Name 'Get-EssTelTenantClass' -ErrorAction SilentlyContinue) { throw 'Get-EssTelTenantClass should be removed' }
 }
 Test 'scrub strips paths, urls, emails and guids' {
     $s = ConvertTo-EssTelScrubbed -Text 'boom C:\Users\x https://a.com/b user@contoso.com 72f988bf-86f1-41af-91ab-2d7cd011db47'
@@ -244,7 +243,7 @@ Test 'lite installer is not instrumented (no telemetry ready, no events)' {
     $old = $env:ESS_ADK_TELEMETRY
     try {
         $env:ESS_ADK_TELEMETRY = ''   # ensure telemetry is otherwise enabled
-        Initialize-EssInstallTelemetry -Installer 'lite' -TenantId ''
+        Initialize-EssInstallTelemetry -Installer 'lite'
         if ($script:EssTel.Ready) { throw 'lite installer should not be telemetry-ready' }
     } finally { $env:ESS_ADK_TELEMETRY = $old }
 }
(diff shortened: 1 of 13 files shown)


apurvabanka:
@amilandi Found a few gaps.

## Notable gaps


**1. `tenant_class` is effectively always `unknown` — undercuts a stated goal.**

`ESS_TENANT_ID` is read (`Install-EssAdk.ps1:343`, `install-ess-adk.sh:90`) but is **never set anywhere** in the repo. Since the installer runs before sign-in, `tenantId` is empty → `tenantClass = unknown` for essentially every installer event. The PR's headline goal of "split Microsoft-internal dogfood from external customer installs" won't work for installer telemetry as written. Worth calling out explicitly (or dropping the dimension until it can be populated).


**2. Cancellation is misreported on Windows (`cancelled` outcome is dead code).**

Neither emitter ever emits `Complete -Outcome 'cancelled'` — the enum value is unused. Worse, on Ctrl+C the PS installer's `finally` block runs `Complete-EssInstallTelemetry -Outcome 'success'` (Ctrl+C bypasses `catch`), so **a cancelled install is recorded as success**. bash is better (EXIT trap fires with code 130 → `failure`), but still labels it `failure`, not `cancelled`. The two platforms disagree and neither uses the `cancelled` path.


**3. Synchronous per-step POSTs add latency on flaky networks.**

Every `Write-Step`/`step()` does an inline HTTP POST with a 5s timeout (`-TimeoutSec 5` / `curl -m 5`). With ~10 steps, an unreachable/slow collector can add up to ~50s to the install. "Fail-open" still means "wait for the timeout." Consider fire-and-forget/backgrounding, or a much shorter timeout.

amilandi:
Thanks @apurvabanka — all three addressed in 78a5415.

**1. tenant_class always `unknown` → dropped the dimension.** You're right that the installer runs before sign-in, so the tenant is genuinely unknowable there and `ESS_TENANT_ID` is never set. Rather than emit a meaningless `unknown` on ~every event, I removed `tenantId`/`tenantClass` (plus the `Get-EssTelTenantClass`/`ess_tel_tenant_class` helpers and the corp-tenant constant) from both emitters and the `Initialize`/`ess_tel_init` signatures. The internal-vs-external split still works where it can be populated — in the post-auth ADK/FlightCheck telemetry.

**2. Cancellation misreported → fixed, both platforms now agree.** Success is now recorded at the *end of the normal try path* (not in `finally`), so a Ctrl+C — which bypasses `catch` and hits `finally` — no longer lands as `success`; `finally` now falls back to `cancelled` (idempotent). The bash EXIT trap maps `130`/`143` (SIGINT/SIGTERM) to `cancelled` instead of `failure`. The `cancelled` enum value is now actually used on both sides.

**3. Synchronous per-step POST latency → bounded.** Shortened the send timeout (5s → 3s) and added a circuit breaker: the first failed/timed-out send disables telemetry for the rest of the process. On an unreachable collector the worst-case added latency is now ~one timeout instead of steps × timeout.

Smoke tests updated to cover all three (42/42 passing).

apurvabanka (approved):

PR #185, opened 2026-07-08
fix(installer): make one-shot installer ASCII-safe for Windows PowerShell 5.1

amilandi:
## Description

The one-shot `winget` installer crashes under **Windows PowerShell 5.1** (the default shell on Windows) with a cascade of parse errors:

```
iex : Exception calling "Create" with "1" argument(s): "At line:740 char:39
+                 $raw = $raw -replace '^\s*\{', "{ $modeEntry,"
Unexpected token '^\s*\' in expression or statement.
```

Reported by a customer running the documented `iex (irm .../setup/bootstrap.ps1)` one-liner.

## Root cause

The setup scripts contain **em-dash characters** (`—`, U+2014) in comments and a few messages. Windows PowerShell 5.1 decodes a no-BOM file as **ANSI (CP1252)** rather than UTF-8, so each 3-byte UTF-8 em-dash is mangled into a byte sequence that derails `[ScriptBlock]::Create()` inside `bootstrap.ps1`. The visible error at line 740 is a cascade — the real trigger is the mis-decoded non-ASCII bytes earlier in the file. PowerShell 7 (what maintainers run) defaults to UTF-8, so the bug was invisible in dev.

Confirmed empirically: the script parses **clean** under 5.1 once non-ASCII characters are removed; real UTF-8 em-dashes parse fine when decoded correctly.

## Fix (defense in depth)

1. **Make all `setup/*.ps1` ASCII-only** — replace em-dashes with hyphens (and the `✓`/`✗` marks in the Pester test file with `[PASS]`/`[FAIL]`). With pure ASCII, every decode path (ANSI, UTF-8, Latin-1) yields identical bytes, so the parser can never break on encoding.
2. **Force UTF-8 in the bootstraps** — the three `bootstrap*.ps1` now read the fetched installer via `[System.IO.File]::ReadAllText($installer, [System.Text.Encoding]::UTF8)` instead of `Get-Content -Raw`, so the decode is explicit regardless of BOM. Prevents this class of bug from recurring if non-ASCII is reintroduced.

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)

## Testing

- Reproduced the exact customer error by parsing the current `main` `Install-EssAdk.ps1` under Windows PowerShell 5.1.
- After the fix, **all 5 `setup/*.ps1` parse clean** under Windows PowerShell 5.1 (`[Parser]::ParseFile` → no errors).
- Verified the UTF-8 read path executes a non-ASCII-containing script under 5.1 where the intent is to be encoding-agnostic.
- No functional/logic changes — only character substitutions in comments/strings and the file-read encoding.

Does not touch `samples/` (changes under `setup/`), so static sample validation is not applicable.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,3 +57,18 @@ jobs:
         PATHS: ${{ steps.discover.outputs.paths }}
       run: |
         python -m compileall -q $PATHS
+
+  installer-scripts:
+    name: Installer script smoke tests
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v6
+
+    # pwsh is preinstalled on GitHub-hosted ubuntu runners. The harness only
+    # parses and string-matches the setup scripts (no OS-specific execution),
+    # so it runs cross-platform. Guards against the Windows PowerShell 5.1
+    # parse crash by asserting the scripts stay ASCII-only. See PR #185.
+    - name: Run installer smoke tests
+      shell: pwsh
+      run: ./setup/Install-EssAdk.Tests.ps1
--- a/setup/Install-EssAdk.Tests.ps1
+++ b/setup/Install-EssAdk.Tests.ps1
@@ -28,7 +28,7 @@ Write-Host "Install-EssAdk.ps1 validation:" -ForegroundColor Cyan
 
 Test 'script parses without syntax errors' {
     $tokens = $null; $errors = $null
-    [System.Management.Automation.Language.Parser]::ParseFile($installerPath, [ref]$tokens, [ref]$errors)
+    $null = [System.Management.Automation.Language.Parser]::ParseFile($installerPath, [ref]$tokens, [ref]$errors)
     if ($errors.Count -gt 0) {
         throw "Parse errors: $($errors[0].Message)"
     }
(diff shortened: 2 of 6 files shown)


amilandi:
---

## Update: regression guard + CI (self-review follow-up)

Reviewed the PR and closed the gap that let this bug ship: the installer smoke-test harness (`setup/Install-EssAdk.Tests.ps1`) existed but **ran nowhere**, and its "parses without errors" tests use the running PowerShell (7 in CI), which decodes UTF-8 fine and never reproduces the 5.1 failure.

Added:
- **ASCII-only guard test** over every `setup/*.ps1` - parser- and version-independent, so it catches this class of bug regardless of which PowerShell runs. Verified it fails when a non-ASCII char is injected.
- **Test that the bootstraps decode the installer as UTF-8 explicitly** (guards the belt-and-suspenders layer against a revert to `Get-Content -Raw`).
- **Wired the harness into CI** as a new `Installer script smoke tests` ubuntu job so the guard is enforced on every PR. Confirmed green on this PR.
- Suppressed the AST dump from the existing `ParseFile` tests (harness output 163KB -> <1KB).

Harness now: **14/14 passing**.

apurvabanka (approved):
LGTM

PR #187, opened 2026-07-09
feat(review): advisory authoring-time conformance review skill (/review)

nkemms:
## Description

Adds `/review` — an advisory, non-blocking static-analysis skill that reviews maker-authored Copilot Studio topics (`.mcs.yml`) for conformance issues **before publish**. It reads and reports; it never modifies topics.

The skill runs six conformance checks over a topic (or a whole module):

- **Power Fx logic** (agentic) — open-ended reasoning over the topic's expressions, guided by heuristics rather than a fixed rule set; covers issues like hardcoded environment values, ungated `ParseValue`, and failure-branch gaps.
- **Dangling `Global.*`** (detector `scan_globals.py`) — reads with no matching definition.
- **Card/topic binding-mismatch** (detector `scan_bindings.py`) — UX contract.
- **ISV conformance** (agentic) — checks the topic's field/schema use against the backend's documented conventions and known pitfalls, guided by the ISV reference docs rather than a fixed checklist.
- **ISV integration-pattern** (agentic) — judgment on whether the topic follows the supported shared-orchestrator approach rather than a standalone-flow shortcut.
- **ServiceNow config/topic** (detector `scan_config.py`) — parsed-but-not-produced response fields (silent blanks).

Findings persist to a per-topic **catalog + resolution ledger** (`merge_findings.py`) with strict finding-contract validation, sha256 evidence-hash staleness, and a stdin handoff. Supports **single-topic and scoped (per-module)** review. Reference docs (ISV domain, runtime heuristics) are synced as needed.

When those reference docs are absent (e.g. an external maker without the checkout), the review **degrades gracefully**: it runs every check it can, scores reliability-finding severity from the topic's visible structure (delegation to the shared system topic) rather than the runtime docs, and reports **honest reduced coverage** — naming the one skipped check (backend-specific ISV field conformance) rather than implying a clean bill of health. The reliability invariants are also captured as forward authoring guidance in `authoring-invariants.md`, wired into `/create` and `/update`.

## Type of change

- [x] Documentation update

## Testing

- Detectors validated on the real HR agent: 0 false positives whole-agent; planted positives (typo Global, mismatched card binding, `ResolutionNotes` silent-blank) each flagged; reverted clean.
- Agentic lenses validated on real Workday + ServiceNow topics (both ISV directions).
- `merge_findings.py` exercised via CLI: schema-valid write, malformed-finding rejection (exit 2), stdin / file / not-found paths, per-run behavior.
- Live `/review` runs in VS Code (single-topic and scoped) across multiple models.
- Graceful reduced-coverage validated live: ISV docs present → full review, silent about coverage; a backend's doc absent → that topic reports reduced coverage while other backends stay full; runtime docs absent → still full coverage (severity scored structurally).
- Reference-doc presence is resolved by a **direct filesystem check** (the synced dirs are gitignored, so an index-respecting search would false-negative and silently skip the check).

## Checklist

N/A — this PR does not touch `samples/`.

### DIFF
--- a/solutions/ess-maker-skills/.github/prompts/menu.prompt.md
+++ b/solutions/ess-maker-skills/.github/prompts/menu.prompt.md
@@ -18,7 +18,7 @@ Here's what I can help you with:
 | `/update` | Type Enter to modify an existing topic or workflow |
 | `/delete` | Type Enter to delete a topic or workflow from your agent |
 | `/scan` | Type Enter to scan your agent for compile errors and fix them |
-| `/review` | Type Enter to review a topic for issues before publishing |
+| `/review` | Type Enter to review a topic (or a whole module's topics) for issues before publishing |
 | `/evaluate` | Type Enter to generate evaluation test sets for your agent |
 | `/flightcheck` | Type Enter to run a pre-deployment readiness check |
 | `/push` | Type Enter to push all local changes to Copilot Studio |
--- a/solutions/ess-maker-skills/.github/prompts/menu.prompt.md
+++ b/solutions/ess-maker-skills/.github/prompts/menu.prompt.md
@@ -18,6 +18,7 @@ Here's what I can help you with:
 | `/update` | Type Enter to modify an existing topic or workflow |
 | `/delete` | Type Enter to delete a topic or workflow from your agent |
 | `/scan` | Type Enter to scan your agent for compile errors and fix them |
+| `/review` | Type Enter to review a topic for issues before publishing |
 | `/evaluate` | Type Enter to generate evaluation test sets for your agent |
 | `/flightcheck` | Type Enter to run a pre-deployment readiness check |
 | `/push` | Type Enter to push all local changes to Copilot Studio |
--- a/solutions/ess-maker-skills/.github/prompts/review.prompt.md
+++ b/solutions/ess-maker-skills/.github/prompts/review.prompt.md
@@ -1,6 +1,6 @@
 ---
 mode: agent
-description: "Type Enter to review a topic for issues before publishing"
+description: "Type Enter to review a topic (or a whole module's topics) for issues before publishing"
 ---
 
 # Review
@@ -11,7 +11,8 @@ description: "Type Enter to review a topic for issues before publishing"
 
 and STOP. Otherwise proceed with the skill instructions below.
 
-You are helping a customer review an authored topic for issues **before they publish it**. This review is
+You are helping a customer review authored topics for issues **before they publish them** — either a single
+topic, or all the topics for a backend module (e.g. "review all the Workday topics"). This review is
 **advisory** — it surfaces findings and lets the customer decide; it never blocks.
 
 Read the skill instructions at `src/skills/topics/review/SKILL.md`, then follow the steps in order.
--- a/solutions/ess-maker-skills/.github/prompts/review.prompt.md
+++ b/solutions/ess-maker-skills/.github/prompts/review.prompt.md
@@ -1,0 +1,17 @@
+---
+mode: agent
+description: "Type Enter to review a topic for issues before publishing"
+---
+
+# Review
+
+**Setup-state check.** Read `.local/config.json`. If it does not exist, OR `setup` is not `"complete"`, show:
+
+> Welcome to the ESS Maker Kit. Before running `/review`, type `/setup` to set up your environment.
+
+and STOP. Otherwise proceed with the skill instructions below.
+
+You are helping a customer review an authored topic for issues **before they publish it**. This review is
+**advisory** — it surfaces findings and lets the customer decide; it never blocks.
+
+Read the skill instructions at `src/skills/topics/review/SKILL.md`, then follow the steps in order.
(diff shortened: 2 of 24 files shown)


apurvabanka:
## Review: `/review` conformance skill

Reviewed the skill, the 7 conformance docs, and the 6 scripts, and ran the
detectors against the OOB HR agent. The detector resolution chains work
correctly on real topics (ServiceNow scenario→config resolution and ISV
backend detection both fire). A few gaps worth addressing before merge.

### Blocking / logic

**1. Dismissed findings resurface every run (no suppression).** Reproduced:
when a maker dismisses a detector-sourced finding as `not-a-bug`/`wont-fix`,
`merge_findings.append_resolutions` writes a `resolved` ledger line, but on the
next run the deterministic detector re-emits the finding in `--current` and
`merge()` unconditionally sets it back to `active`. Net effect:
- catalog shows `active` while the ledger says `not-a-bug` — a live
  contradiction (not the "latent" one the docstring claims);
- every re-run appends another `:rN` ledger entry → append-only bloat;
- there's no mute/suppression list, so false positives can never be silenced.
Suggest consulting the ledger (scoped by `solution`) in `merge()` to keep a
dismissed `(solution, issue_id)` suppressed until its evidence hash changes.

**2. Silent skips reported as clean (ServiceNow config check).**
`scan_config.scenario_of()` only reads `ScenarioName` from the per-scenario
system topic resolved via `.component-map.json`. A user topic that
`BeginDialog`s straight into the shared `*SystemCommonExecution` and passes
`scenarioName` inline resolves to a topic with no literal `ScenarioName` →
returns `None` → topic is silently skipped. But `report-format.md` tells the
maker "everything else always runs" and only ISV conformance is disclosable as
reduced coverage. So a ServiceNow topic can get a clean response-field bill
when the check never ran, with no disclosure. Either resolve the inline-scenario
wiring or surface skipped topics in the coverage note.

**3. File-granular staleness over-triggers.** `evidence_hashes` hash the entire
topic file, and every finding's `files[]` is that whole file, so any unrelated
one-character edit flips *all* of a topic's findings to `evidence_stale`.

### Coverage / testing

**4. No tests.** Six non-trivial scripts (evidence hashing, cross-run
reconcile, YAML-with-reserved-key parsing, path guards) ship with zero tests
despite `tests/` existing and determinism being the stated rationale. Gaps
1–3 are exactly what unit tests would catch.

**5. Assembly step reintroduces the nondeterminism.** Only 3 of 6 checks are
script-backed; Power Fx, UX Part 2, and both ISV checks are fully agentic, and
cross-run identity depends on the LLM re-emitting the *same slug* per issue with
no script assist for matching (double-count risk is acknowledged but
unmitigated).

**6. YAML parse failure = false-clean.** `scan_bindings`/`scan_config` drop a
topic (stderr warning only) on any `YAMLError` and exclude it from two checks.
Parsed the OOB agent fine, but a maker topic with any other YAML-hostile
construct silently falls out of coverage.

### Doc inconsistencies

- `finding-contract.md` opens with "no lens writes anything to disk," then
  documents the persisted catalog + ledger (and Step 8 writes them).
- Same doc says the customer-facing translation is "Step 7"; it's Step 9 in the
  final SKILL (Step 7 is now "Consolidate").
- `powerfx-topic-local.md` uses `reachable: unreachable` phrasing that doesn't
  match the `REACHABILITY_KINDS` enum the script validates.
- Coverage note requires the agent to hand-translate probe ids
  (`servicenow-hrsd`, `workday`) to plain names with no lookup table.

### Solid
Path-traversal guards, cwd-immune `__file__`-anchored probe, contract
validation (exit 2 on malformed findings), "absence ≠ resolution" carry-forward,
and the reserved-`@`-key YAML shim all work correctly against the real agent.

CC: @nkemms

nkemms:
Thanks @apurvabanka — all blocking/logic items and doc nits are addressed on the branch (`ed586c2`, `df444ba`). Summary:

- **1 (suppression):** `merge()` now consults the ledger (scoped by `solution`) — a dismissed finding (`not-a-bug`/`wont-fix`/`false-positive`) is carried as a new `suppressed` status until its evidence changes, instead of resurfacing as `active`. `fixed` still reopens on re-detect. `append_resolutions` is idempotent (no `:rN` bloat). Validated end-to-end with a planted-then-dismissed finding staying suppressed on re-review.
- **2 (silent config skips):** the scanner now resolves the scenario via inline binding, `Switch` fan-out (unions produced fields across branches), and a transitive delegation hop — unresolved topics on the OOB agent went ~18 → 0. Genuinely unresolvable scenarios (and unloadable configs) are now disclosed as incomplete coverage, never reported clean.
- **4 (tests):** `tests/scripts/test_review_engine.py` now has 25 tests covering suppression, scenario resolution, and disclosure paths.
- **6 (YAML false-clean):** fixed earlier (`5e5bfdf`) — skipped topics warn + gate the clean message; exit stays 0.
- **Doc nits:** "no lens writes to disk" reworded + `status` documents `suppressed`; Step 7 → Step 9; `reachable: unreachable` → `NOT_REACHABLE_VIA_BOT_UI`; the probe now emits plain backend display names read verbatim (+ lookup table), so no hand-translation.

**Deferred (real, non-blocking):** finer per-finding staleness (#3), agentic-lens cross-run slug matching (#5), plus follow-ups for full multi-file evidence in the ledger and correlating the transitive hop with its output binding.

Thanks for confirming the path guards, cwd-immune probe, contract validation, carry-forward, and `@`-key shim.

apurvabanka (approved):

PR #193, opened 2026-07-14
feat(mtk): [Worker][TASK-002] | Generic typed pipeline framework

anil-adepu:
## Description

Implement the generic typed pipeline framework in `src/core/pipeline/` per
`02_ARCHITECTURE/PIPELINES.md` v2.0.

- `Pipeline[TInput, TOutput]` and `PipelineStep[TInput, TOutput]` — immutable,
  frozen generic abstractions with runtime type validation at every step boundary.
- Fluent, immutable `PipelineBuilder` with type-threading (incompatible adjacent
  step types fail at build time).
- `PipelineRegistry` — ordered step collection with duplicate-name rejection.
- `PipelineContext` protocol for shared-context threading (no globals).
- `StagedPipeline[TContext]` — generic composition of context-preserving stage
  pipelines, executed left-to-right.
- Typed runtime type-spec helpers and pipeline exceptions.

## Type of change

## Testing

Gates pass from `tools/ess-nextgen-migration-toolkit/`:
- `uv run pytest` — 12 pipeline tests passed
- `uv run ruff check .` — clean
- `uv run mypy src` — no issues


amilandi (dismissed):
## Approve — TASK-002 Pipeline framework

Read the whole framework (`pipeline.py`, `step.py`, `staged.py`, `registry.py`, `context.py`, `exceptions.py`, `typing.py`) and both test files. Ran the gates locally:

- `pytest tests/unit/core/pipeline/` — 12 passed
- `ruff check` — passed
- `mypy` (strict) — no issues

**Strengths**
- Solid immutable, generic design. `Pipeline` is a frozen `Generic[TInput, TOutput]` dataclass; `PipelineBuilder` threads types between adjacent steps at build time (`type_specs_compatible`) and rejects incompatible/duplicate steps, while `run()` re-validates runtime values at pipeline input, each step boundary, and pipeline output.
- Skippable steps via `can_execute`, failures propagate unsuppressed, and `StagedPipeline` cleanly generalizes context-preserving super-pipelines with an overridable `_ordered_stages` hook.
- The `typing.py` helpers (normalize/compatible/matches/name) are small, deterministic, and well-factored. Tests cover ordering, type threading, type-changing steps, duplicate rejection, empty identity, failure propagation, and bad runtime output.

No blocking issues found. LGTM.

tools/ess-nextgen-migration-toolkit/src/core/pipeline/pipeline.py
```diff
@@ -0,0 +1,172 @@
...
+
+        current: object = pipeline_input
+        for step in self._steps:
+            if not value_matches_type_spec(current, step.input_type):
+                raise PipelineExecutionError(
+                    f"Step '{step.name()}' expected input type {type_spec_name(step.input_type)}."
+                )
+
+            typed_current = cast(Any, current)
+            if not step.can_execute(typed_current):
```
aditandadit, line 105: [Question Not Blocking] Should a skipped step ever be one that changes the type, or would that trip things up?

Since a skipped step just passes the value through unchanged, if it was the one meant to change the type, the next step gets the wrong type and throws at runtime (line 99), even though the build looked fine.

tools/ess-nextgen-migration-toolkit/src/core/pipeline/staged.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+
+    _stages: tuple[Pipeline[TContext, TContext], ...] = field(default_factory=tuple)
+
+    @property
+    def stages(self) -> tuple[Pipeline[TContext, TContext], ...]:
+        """Return the ordered configured stages."""
+        return self._stages
+
+    def add(self, stage: Pipeline[TContext, TContext]) -> StagedPipeline[TContext]:
```
aditandadit, line 37: [Non Blocking] Should stages reject duplicate names the way steps already do?
The step registry blocks two steps sharing a name, but add here doesn't, so accidentally wiring the same stage twice would just run it twice with no warning.

tools/ess-nextgen-migration-toolkit/src/core/pipeline/step.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def name(self) -> str:
+        """Return the deterministic step name used for registration."""
+        return self._name
+
+    def description(self) -> str:
+        """Return a human-readable description of this step."""
+        return self._description
+
+    def supported_modes(self) -> frozenset[str]:
```
aditandadit, line 62: Is supported_modes meant to gate execution, or is that left entirely to each step?

It's stored and exposed, but nothing in run() or the default can_execute consults it, so a step declaring supported_modes=("MIGRATE",) still runs in every mode unless it also overrides can_execute.

aditandadit (dismissed):
nothing blocking, approving

tools/ess-nextgen-migration-toolkit/src/core/pipeline/pipeline.py
anil-adepu: Every step reads and returns the same type. A type-changing step that's also conditionally skippable would be a design error. The builder validates the full chain at build time assuming all steps participate. Closing as no-op.

tools/ess-nextgen-migration-toolkit/src/core/pipeline/staged.py
anil-adepu: Edit: We are handling this already in registry.py

tools/ess-nextgen-migration-toolkit/src/core/pipeline/step.py
anil-adepu: This is only a template code, actual step can either respect the can_execute state from base class or override it with its own execution logic.

savankayala05 (approved):

PR #195, opened 2026-07-15
Harden Workday setup: advisory topic review, scoped push, OOTB topics installer, Windows path & az rest fixes

apurvabanka:
## Summary

A batch of reliability, correctness, and UX improvements to the ESS Maker Kit's
Workday setup flow, FlightCheck, and push tooling. The changes are logically
independent and could be split, but are shipped together here.

### 1. Advisory topic review wired into setup and create flows
- New tracked checklist row **S6.3** (`advisory` gate) runs `/review` over the
  finished topic **after** `TOPIC-TRIGGER-*` and `TOPIC-INTEGRATION-*` pass and
  the topic is fully wired (setup step **P6.5**). It surfaces the report in chat
  verbatim and never blocks setup.
- Standalone `/create` gains the same review at **step 6.5**.
- Enforced by an always-loaded `runSubagent` directive in
  `.github/copilot-instructions.md`; the premature P6.1 authoring-time review is
  explicitly suppressed (IDs aren't wired yet, so it would false-flag).
- New `advisory` gate documented in `checklist-updater.md` / `config-schema.md`
  and guarded by `tests/setup/`.
- Files: `topics/create/SKILL.md`, `topics/review/SKILL.md`,
  `setup/workday/create-new-topic.md`, `setup/SKILL.md`, `setup/workday/tasks.md`,
  shared `checklist-updater.md` / `config-schema.md`, `validation-matrix.md`.

### 2. Surface MANUAL checkpoint results in chat (no browser popup)
- Single-checkpoint FlightCheck runs now print the full MANUAL verification steps
  inline instead of opening `report.html`; full runs auto-open the report only
  when a MANUAL row exists (`flightcheck/cli.py`).
- Setup playbooks now instruct "show the checkpoint result in chat first" per
  `checklist-updater.md` §U.0–U.0a before any attestation
  (`provision-power-platform-environment.md`, `provision-workday-entra-app.md`,
  `configure-workday-tenant.md`, `install-ess.md`).

### 3. `az rest` Windows encoding fix
- The Dataverse role-check URL used raw parentheses `systemusers({id})`, which
  breaks on Windows (the `az` launcher is a `cmd.exe` batch wrapper — a raw `)`
  fails with "was unexpected at this time"). Now percent-encoded
  `systemusers%28{id}%29` (with `%24select`) across all 4 playbooks that use it.

### 4. skill-2: remove redundant re-check
- When the ESS solution is already installed, the P2.1 pre-check pass now records
  the row directly instead of re-running the identical `ESS-SOLN-001` checkpoint
  in P2.2 (`install-ess.md`).

### 5. Optional OOTB Workday topics installer
- New deterministic `scripts/install_ootb_topics.py` copies vendored sample
  Workday topics (TOPICS ONLY — managed template-configs are excluded) with
  kit-convention filenames, removing the assistant from the file-naming loop
  (which produced mangled slugs). Offered between skill-5 and skill-6.
- Files: `install_ootb_topics.py`, `setup/workday/install-workday-ootb-topics.md`,
  `role-gating.md`, `setup/SKILL.md` router offer.

### 6. push.py: scoped push + stale-botId pre-flight
- `--only <glob>` (repeatable) / `--only-from <file>` scope a push and refresh
  only the matching baseline entries (`update_baseline_scoped`), so unscoped
  pending changes stay pending.
- New `_verify_bot_exists` pre-flight surfaces a stale/mismatched `botId` (404)
  with a clear message instead of the misleading "could not find agent
  components / solution not deployed".

### 7. Windows deep-path (MAX_PATH) robustness
- `windows_long_path()` helper + shorter extraction slugs (hash-suffixed) so
  checkpoint/baseline copies survive deep OneDrive-backed workspaces
  (`checkpoint.py`, `setup.py`; also fixes an `onexc`→`onerror` regression).

### 8. WD-CONN-102 SAML certificate: no-thumbprint comparison
- Workday's X509 Public Key object exposes no thumbprint, only Name / Valid From
  / Valid To / Base64 body. The MANUAL remediation now compares via validity
  dates or an externally-computed SHA-1 (PowerShell/openssl), and the skill-4
  wording/questions follow suit (`flightcheck/checks/workday.py`,
  `remediation-guide.md`, `configure-workday-tenant.md`,
  `test_workday_saml_certificate.py`).

### 9. Doc clarifications
- `connection-fields.md`: distinguish the Application ID URI (`api://{appId}`)
  from the connection's "Microsoft Entra resource URL"
  (`http://www.workday.com/{tenant}`).

## Testing
- `pytest tests/setup/` → green (structural guards updated for S6.3 + `advisory`
  gate + the new checklist row/count).
- Also run the touched suites:
  `pytest tests/setup tests/scripts/test_push_scope.py tests/scripts/test_install_ootb_topics.py tests/flightcheck/checks/test_workday_saml_certificate.py -q`

### DIFF
--- a/solutions/ess-maker-skills/.github/copilot-instructions.md
+++ b/solutions/ess-maker-skills/.github/copilot-instructions.md
@@ -332,6 +332,34 @@ before continuing. Do NOT proceed to review and push until the subagent
 has returned. Do NOT invoke the validate subagent after entire test set
 delete operations or after deleting the last remaining case in a category.
 
+**Topic review invocation:** When the maker runs `/create` **directly**, at
+**step 6.5 of the topic create flow** (`src/skills/topics/create/SKILL.md`) —
+after the scan and before the dry run/push — invoke `runSubagent` (the VS Code
+Copilot Chat tool) pointing the subagent to read
+`src/skills/topics/review/SKILL.md` as its first action, scoped to the **single
+topic just created** (pass the agent slug from `.local/config.json` and the
+topic stem — the filename without `.mcs.yml`) and asking it to present the
+**maker-facing report**. Running this review is **mandatory** in the direct
+`/create` flow: wait for the subagent to return, then paste its full report
+verbatim into the chat. Do NOT proceed to the dry run or push until the review
+has returned and its report is shown. The findings themselves are **advisory**
+— they never block the push; when findings exist, pause and let the maker
+choose to fix now (via `/update`) or push anyway. If the subagent or its
+detector scripts cannot run, say the review was skipped and continue — a review
+failure never blocks the push.
+
+**Do NOT invoke this step-6.5 review when the create-topic skill is running as
+the Workday setup flow's P6.1 authoring delegation**
+(`src/skills/setup/workday/create-new-topic.md`). At P6.1 the tenant reference
+IDs are not wired yet (P6.2 does that), so a review there would false-flag
+unresolved placeholders — there is **no** topic review at S6.1/P6.1. The Workday
+setup flow performs its **one and only** topic review later, at its **step P6.5
+(checklist row S6.3)** — *after* the `TOPIC-TRIGGER-*` and `TOPIC-INTEGRATION-*`
+checkpoints (S6.1, S6.2) pass and the topic is fully wired. That review uses the
+same invocation (`runSubagent` → `src/skills/topics/review/SKILL.md`, single
+topic, maker-facing report, displayed verbatim) and is `advisory` (row S6.3
+completes once the report is shown — it never blocks setup).
+
 **When the user asks to modify, delete, rename, or otherwise change an agent
 component, ALWAYS load and follow the corresponding skill file.** The skill
 contains the full checkpoint→edit→scan→push pipeline. Do NOT improvise a
--- a/solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py
+++ b/solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py
@@ -193,7 +193,7 @@ _REF_SUFFIX_ROLES = {
 #     SAML Identity Providers". This is NOT reachable via any public
 #     Workday API the kit talks to (the SOAP RaaS / Worker services
 #     don't expose tenant security configuration). Comparison of the
-#     two thumbprints is therefore an operator step.
+#     two certificates is therefore an operator step.
 #
 # WD-CONN-102 reads the Entra side automatically, surfaces the
 # current active-cert thumbprint and NotAfter date, and emits a
@@ -2723,7 +2723,8 @@ def _check_saml_certificate_health(runner) -> list[CheckResult]:
         ``CERT_EXPIRY_WARN_DAYS`` days, OR its NotBefore is in the
         future (not yet valid).
       * MANUAL — active cert is healthy. The operator must compare
-        its thumbprint against the row in Workday's "Edit Tenant
+        its certificate (validity dates, or an externally-computed
+        SHA-1 thumbprint) against the row in Workday's "Edit Tenant
         Setup - Security -> SAML Identity Providers" because that
         is not reachable from any Workday API the kit talks to.
       * NOT_CONFIGURED — no federated Workday SAML enterprise app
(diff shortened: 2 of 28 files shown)


amilandi (approved):
## Approve — Workday setup hardening (advisory review, scoped push, OOTB installer, path & az rest fixes)

Reviewed the full diff against the base branch (`feature/worlday-ask-skills`), focusing on the Python (`install_ootb_topics.py`, `push.py`, `flightcheck/cli.py`, `flightcheck/checks/workday.py`) and spot-checking the playbook/instruction changes. Ran the gates locally:

- `pytest tests/setup tests/scripts/test_push_scope.py tests/scripts/test_install_ootb_topics.py tests/flightcheck/checks/test_workday_saml_certificate.py` — **75 passed**
- Full suite — **1141 passed, 12 skipped** (see note on the 10 installer failures below)
- `ruff check` on all four changed scripts — clean

**Strengths**
- **`update_baseline_scoped`** is the crux of the scoped-push feature and it's correct: it refreshes only the scoped files' baseline entries (plus `.xml`→`.meta.json` companions) and applies scoped deletes, so unscoped pending changes stay pending. The comment explaining *why* full `update_baseline()` would be a correctness bug is excellent, and the tests pin exactly that (in-scope refresh, out-of-scope untouched, meta companion, scoped delete).
- **`_verify_bot_exists`** pre-flight cleanly converts Dataverse's misleading blanket-404 ("solution may not be deployed") into the real cause, exits `2` only on 404, and re-raises everything else so transient failures aren't masked. `env_url`/`bot_id` are resolved from config well before the call. Tested for all three branches.
- **OOTB installer** is pure, hermetic, idempotent (working + baseline existence check), dry-run aware, and — critically — TOPICS ONLY, with a test asserting no `template-configs/` dir is ever created. Deterministic PascalCase filename derivation kills the mangled-slug bug at the source.
- **`az rest` percent-encoding** (`systemusers%28{id}%29`, `%24select`) is applied consistently across all 4 playbooks — no raw `systemusers(` left — and matches the known Windows `cmd.exe`-wrapper portability rule.
- **SAML cert (WD-CONN-102)** remediation correctly drops the false premise that Workday displays a thumbprint; the validity-dates / externally-computed-SHA-1 comparison path is accurate, and the test now pins the `Valid From` / `SHA-1` / `does NOT` phrasing per the result+remediation assertion rule.
- Single-checkpoint FlightCheck surfacing MANUAL steps inline (and full runs only auto-opening the browser when a MANUAL row exists) is a sensible chat-first UX improvement.

**Note — not from this PR**
- The full suite shows 10 failures in `tests/scripts/test_installer_launch.py` / `test_installer_maker_profile.py`. I confirmed these **also fail on the base branch** (`feature/worlday-ask-skills`) — they are pre-existing and untouched by this PR. Flagging only so they aren't attributed here; worth a separate fix on the base.

**Minor / optional (non-blocking)**
- OOTB `rewrite_topic` repoints `msdyn_copilotforemployeeselfservicehr.topic.*` references to the target agent's schema. This is correct for the ESS pattern (system topics live under the agent's own schema) and a no-op when schemas match — just worth one sanity check against a real customer schema that differs from the sample default.
- `tests/setup/test_checklist_template.py` uses `re.match(r"[a-z]+", meta["gate"]).group(0) == "advisory"` where a plain `meta["gate"] == "advisory"` would read clearer. Harmless.

Solid, well-tested, well-commented work. LGTM.

daeunJe0ng (commented):
[copilot-review] Thanks for this - solid, well-reasoned hardening, and the rationale in each section made it easy to follow. 

Nice fixes on the mangled-filename bug, the Windows cmd.exe parenthesis break, and the misleading 404. 

The thumbprint -> certificate correction is factually right and applied consistently.

One thing I'd like to sort before merge, plus a couple of small notes.

Should fix before merge
- Tests: install_ootb_topics.py's own docstring calls it "fully unit-testable," and push.py picked up non-trivial glob-scoping + scoped-baseline logic, but I don't see unit tests for either in the diff. Given this is net-new deterministic file-manipulation code, could we add coverage (basename normalization, rewrite_topic substitutions, idempotent skip-exists, matches_only, update_baseline_scoped)?

Worth confirming
- cli.py now opens the HTML report on a full run only when result.manual > 0. That means a full run with FAILED but no MANUAL rows no longer opens the report. Intended? The compact chat table covers it, just want to confirm it's deliberate.

Minor / nits (non-blocking)
- rewrite_topic only substitutes <TENANT_NAME> when tenant is truthy, so an unresolved tenant silently writes the placeholder through (caught later by TOPIC-INTEGRATION-*, but a write-time warning might be friendlier).
- select() silently drops unmatched --scenarios names; a heads-up when some don't match would help.
- config-schema.md shows ootbTopics.selected as msdyn_ schema names, but the installer matches folder / PascalCase base names - slight mismatch.

Process-wise: totally your call, but since the five changes are independent, splitting into smaller PRs would make each easier to review and revert. Not blocking.

PR #196, opened 2026-07-15
FlightCheck telemetry: tenant name, runOutcome, Kusto query pack, connect invocation source

amilandi:
## Summary

FlightCheck / ADK telemetry improvements from the latest round of PM + team feedback (parent User Story: [ADO 7590579](https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7590579)). Four self-contained changes:

### 1. Emit tenant display name — [ADO 7590589](https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7590589)
Adds a `tenantName` dimension (org display name, OII — privacy-approved, best-effort) to both the FlightCheck and ADK telemetry event families, resolved once from the already-authenticated Graph `/organization` record. Lets the dashboards show a human-readable tenant instead of only the raw GUID.

### 2. Add `runOutcome` verdict bucket — [ADO 7590584](https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7590584)
Derives a `runOutcome` (`errored | failed | warnings | ready`) from the run result so the "Runs by Verdict" donut can distinguish a FlightCheck that **broke at runtime** from one that legitimately returned `NOT_READY`. This replaces the retired, misfiring "errors by category" chart (which reported more errors than runs).

### 3. Self-serve Kusto query pack — [ADO 7590585](https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7590585)
Adds `scripts/telemetry_queries.kql`: a documented pack of KQL queries (usage, verdicts/success rate, checks-by-category, runs-to-fix, time-to-green excl. weekends, per-tenant, run duration) so PMs and the broader team can query the Aria/Asimov cluster directly. Includes an Event-reader access-prerequisite header. Failure-oriented queries count both `Failed` **and** `Error` check statuses (errored checks are readiness blockers too, matching #2's `runOutcome` intent), and the check success-rate denominator is restricted to gating statuses (`Passed`/`Failed`/`Error`) so non-gating outcomes (`Skipped`/`NotConfigured`/`Manual`/`Warning`) don't deflate it.

### 4. Wire real `deploy_target` (sandbox vs production) — [ADO 7609605](https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7609605)
Previously `agent.deploy` always emitted `deploy_target=production` (the env var that fed it was never set), so the "Deploy by Target" tile couldn't distinguish real sandbox vs production deploys. Now `push.py` resolves the target from the deployment environment's **Power Platform SKU** and `adk_telemetry.classify_deploy_target()` buckets it into `sandbox` vs `production` (non-prod SKUs: sandbox/trial/developer/teams/subscriptionbasedtrial/support/playground). Resolution is priority-ordered and fully best-effort (never breaks `/push`):

1. explicit `ESS_ADK_DEPLOY_TARGET` override — honored **only** when it names a current bucket (`sandbox`/`production`); a stale `test`/`staging` value is ignored so retired wedges can't resurface,
2. `environmentSku` cached in config (`setup.write_config` now preserves it),
3. best-effort **silent-only** BAP lookup via `PPAdminClient.get_environment_sku_by_dataverse_url()` (`authenticate_silent()` never prompts), cached back to config,
4. `production` default.

> `test`/`staging` were dropped — they are not real Power Platform environment types (they only ever came from the synthetic emitter), and there is no reliable signal to distinguish them. The synth emitter and the dashboard "Deploy by Target" tile descriptions (dev + prod) were updated to `sandbox`/`production` accordingly.

## Testing
- `pytest tests/flightcheck/test_cli.py tests/flightcheck/test_telemetry.py` — all pass.
- `pytest tests/test_adk_telemetry.py` — all pass (incl. `tenantName`, `runOutcome` precedence, and `classify_deploy_target` mapping: non-prod→`sandbox`, prod/unknown/empty→`production`, case/whitespace-insensitive).
- `pytest tests/test_push.py` — `_resolve_deploy_target` precedence (valid override wins & skips lookup; retired/blank override ignored → falls through to SKU classification; cached SKU → silent lookup+cache → `production` default) and `_cache_environment_sku` write/no-op.
- `pytest tests/flightcheck/test_pp_admin_client.py` — `get_environment_sku_by_dataverse_url` host matching + `environmentType` fallback, and `authenticate_silent` returning `False` (never prompting) when no token cache exists.
- Telemetry emit paths are best-effort / fail-open and never affect FlightCheck or `/push` exit codes.

## Dashboard rollout
Dev Aria dashboards already updated and reviewed; prod replication of the non-regressive changes (descriptions + retired-chart deletion + Deploy-by-Target and installer error-category tile descriptions) is done. Registering `tenantName`/repointing the verdict donut to `runOutcome` in **prod** is intentionally deferred until this branch merges and deploys, since prod doesn't emit those fields yet (no synthetic seeding of prod).

ADO: https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7590579

### DIFF
--- a/solutions/ess-maker-skills/scripts/adk_telemetry.py
+++ b/solutions/ess-maker-skills/scripts/adk_telemetry.py
@@ -184,7 +184,7 @@ _SYNC = os.environ.get("ESS_ADK_TELEMETRY_SYNC", "").strip().lower() in (
 # subsequent events inherit it so api.call / build / deploy don't each need
 # to re-derive it. We deliberately do NOT keep any developer/user identifier
 # here — only a random install ``instance_id`` and the raw tenant GUID.
-_IDENTITY: dict[str, str] = {"instance_id": "", "tenant_id": ""}
+_IDENTITY: dict[str, str] = {"instance_id": "", "tenant_id": "", "tenant_name": ""}
 
 # Track dispatched daemon threads so flush() can join them (tests / session end).
 _THREADS: list[threading.Thread] = []
@@ -198,7 +198,9 @@ _SESSION_LOCK = threading.Lock()
 
 
 # --- Identity model (spec) ------------------------------------------------
-def set_identity(tenant_id: str = "", instance_id: str | None = None) -> dict[str, str]:
+def set_identity(
+    tenant_id: str = "", instance_id: str | None = None, tenant_name: str = ""
+) -> dict[str, str]:
     """Record the install + tenant identity for all subsequent events.
 
     ``instance_id`` is a random GUID identifying the ADK installation (not a
@@ -206,12 +208,16 @@ def set_identity(tenant_id: str = "", instance_id: str | None = None) -> dict[st
     developer identifier. When omitted it defaults to the persisted per-install
     GUID (``flightcheck.telemetry.get_instance_id``). ``tenant_id`` is stored
     RAW (approved Data Profile: OII, no transformation) so dashboards can
-    filter by the literal Entra tenant GUID. No developer/user id is collected.
+    filter by the literal Entra tenant GUID. ``tenant_name`` is the tenant's
+    organization display name (OII identifying the enterprise tenant, not a
+    person; privacy-approved without a Data Profile update) — best-effort,
+    stored as ``""`` when unavailable. No developer/user id is collected.
     """
     _IDENTITY["instance_id"] = (
         _fc.get_instance_id() if instance_id is None else (instance_id or "")
     )
     _IDENTITY["tenant_id"] = tenant_id or ""
+    _IDENTITY["tenant_name"] = tenant_name or ""
     return dict(_IDENTITY)
 
 
@@ -357,9 +363,11 @@ def common_dimensions(
     session_id: str = "",
     instance_id: str | None = None,
     tenant_id: str | None = None,
+    tenant_name: str | None = None,
 ) -> dict[str, Any]:
     """Build the dimensions present on every event (spec Common Dimensions)."""
     tid = _IDENTITY["tenant_id"] if tenant_id is None else tenant_id
+    tname = _IDENTITY["tenant_name"] if tenant_name is None else tenant_name
     return {
         "schema_version": SCHEMA_VERSION,
         "instance_id": (
@@ -367,6 +375,7 @@ def common_dimensions(
         ) if instance_id is None else instance_id,
         "tenant_id": tid,
         "tenant_class": _fc.classify_tenant(tid),
+        "tenant_name": tname,
         "session_id": session_id,
         "surface": surface,
         "adk_version": _fc.get_adk_version(),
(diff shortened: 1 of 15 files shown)


apurvabanka:
I've reviewed the full diff and traced the key code paths. The changes are well-structured and defensively coded (best-effort, fail-open). Here are the gaps I found, by significance.

## Notable

**1. The `connect` invocationSource likely never actually fires (change #4's core goal).**
The mechanism is: `/connect`'s `SKILL.md` exports `ESS_FLIGHTCHECK_INVOCATION_SOURCE=connect`, and `cli.py` uses that as the *default* for `--invocation-source`. But every real FlightCheck run is launched with the flag set **explicitly**:
- `src/skills/flightcheck/SKILL.md` → `cli.py --scope {SCOPE} --invocation-source adk`
- `src/skills/onboarding/step3-flightcheck.md` → same, `--invocation-source adk`

An explicit `--invocation-source adk` overrides the env-var default, so a run triggered off `/connect` (which just points the user at the `/flightcheck` command in `workday/step3.md`) is still attributed `adk`. Additionally, `/connect` never runs `cli.py` itself, and env vars exported in one agent/shell turn don't reliably survive to a later `/flightcheck` turn. Net: nothing consumes the `connect` default. To make it work, `/connect` needs to invoke FlightCheck directly with `--invocation-source connect`, or the flightcheck skill must stop hardcoding `adk` and defer to the env var.

## Medium (KQL query-pack correctness in `telemetry_queries.kql`)

**2. "Failures by category" excludes errored checks.** The `Status` enum has both `"Failed"` and `"Error"`, and change #2 deliberately elevates the *errored* bucket. But section 3 filters `status == "Failed"` only, so errored checks are dropped from the per-category failure counts — inconsistent with the PR's own runOutcome intent.

**3. Check success-rate denominator is inflated.** Section 3 uses `Total = count()` (all checks) with `Passed = countif(status == "Passed")`. That denominator includes `Skipped`/`NotConfigured`/`Manual`/`Warning`, deflating the "success rate."

**4. runs-to-fix / time-to-green treat only `"Failed"` as failing** (sections 4/5 filter `status in ("Passed","Failed")`), so a check that `Error`s then passes isn't counted as a fix. May be intentional, but worth a comment.

## Low / nits

**5. `ESS_ADK_DEPLOY_TARGET` override is returned verbatim, unvalidated** (`_resolve_deploy_target`). A stale `test`/`staging` value would resurface exactly the retired buckets change #5 set out to eliminate.

**6. adk `tenant_name` is gated on `tenant_id`.** In `cli.py`, `_adk.set_identity(..., tenant_name=...)` only runs `if tenant_id:`. If the tenant GUID is empty but the org name resolved, `tenant_name` is dropped from the adk dimensions. (The legacy event family doesn't have this gap — it always passes `tenant_name`.)

**7. Extra Graph `/organization` round-trip.** `authentication.py` already calls `graph.get_organization()`; the telemetry block fetches it again. Best-effort and harmless, but redundant — could be cached on the client.

CC: @amilandi

amilandi:
Thanks @apurvabanka — really thorough pass. All addressed in f7edb8f:

**1. `connect` invocationSource never fires — reverted change #4.** You're exactly right: the flightcheck skill hardcodes `--invocation-source adk` (which overrides the env-var default), and env vars set in the `/connect` turn don't survive to a separate `/flightcheck` turn/process anyway, so nothing ever consumed the `connect` default. Rather than push a bigger UX change (having `/connect` run FlightCheck itself), I've reverted the change entirely — cli.py choices/default, `connect/SKILL.md` export, `validation-matrix.md` docs, and the `test_cli.py` env-var tests. ADO 7587431 will be re-evaluated separately.

**2. Failures-by-category dropped errored checks — fixed.** Failures-by-category (and runs-to-fix) now count `status in ("Failed", "Error")`, consistent with the `runOutcome` intent of surfacing errored checks.

**3. Success-rate denominator inflated — fixed.** The check success-rate query now restricts the denominator to gating statuses (`Passed`/`Failed`/`Error`); `Skipped`/`NotConfigured`/`Manual`/`Warning` no longer deflate it.

**4. runs-to-fix treated only `Failed` as failing — fixed** (see #2; `firstFail` now includes `Error`). Section 5 (time-to-green) keys off run-level `overall` (`READY`/`READY_WITH_WARNINGS` vs not), so `not(ready)` already captures errored runs — left as is.

**5. Unvalidated `ESS_ADK_DEPLOY_TARGET` override — fixed.** The override is now honored only when it names a current bucket (`sandbox`/`production`); a stale `test`/`staging` (or any other) value is ignored and we fall through to SKU detection, so it can't resurface a retired wedge.

**6. adk `tenant_name` gated on `tenant_id` — fixed.** `cli.py` now calls `set_identity` when `tenant_id` **or** `tenant_name` is present, so a resolved org name isn't dropped when the tenant GUID is empty.

**7. Redundant Graph `/organization` round-trip — fixed.** `GraphClient.get_organization()` is now memoized per client instance, so the telemetry lookup reuses the auth-flow result.

apurvabanka (approved):

PR #197, opened 2026-07-16
flightcheck: add INFRA-003 external endpoint reachability check

daeunJe0ng:
## Summary

Adds FlightCheck check **INFRA-003: External endpoint reachability**. It enumerates the external endpoints the agent's extensions use (Workday, ServiceNow, SAP SuccessFactors, custom HTTP) and verifies each is reachable **from the Power Platform environment's own egress** (the path the agent runtime uses), so blocked endpoints are caught at preflight instead of by employees at runtime.

Work item: [7485387](https://dev.azure.com/O365Exchange/O365%20Core/_workitems/edit/7485387/)
Design record: `docs/design-infra-003-endpoint-reachability.md`.

## How it works

- **Egress-only probe.** The opt-in `--runtime-reachability` path stands up a transient Dataverse cloud flow (category 5), triggers one HTTP request from the environment''s egress, reads the status, then deletes the flow. Lifecycle: create -> activate -> `listCallbackUrl` -> invoke -> delete. Any integer status = reachable (PASS); `null` = blocked by DNS/TLS/connection/DLP (FAIL). A local laptop probe was removed because it runs from the wrong network and never sends HTTP.
- **Consent-gated.** This is the kit''s only tenant-mutating path, so it never runs without an explicit decision. `--runtime-reachability` forces on, `--no-runtime-reachability` forces off, interactive runs prompt Y/N, non-TTY/CI stays read-only, and the ADK/chat skill asks conversationally and passes the flag on YES.
- **MANUAL when not probed.** No consent, missing prerequisites, an indeterminate probe, or an endpoint with no recorded URL all return `Status.MANUAL` with re-run guidance and the official outbound-IP + Azure service-tags links (no silent skips, no misleading verdict).
- **Read-only and idempotent by default.** A default run neither mutates nor probes. The mutating path uses a deterministic flow name, sweeps orphans before and after, and guarantees cleanup in `finally`.

## Findings

Unreachable FAIL uses the Shared Steps 5-field role-aware schema (`Probable cause / Configuration Area or Scope / What it implies / Next steps / Responsible role`), names the endpoint URL plus a combined probable cause (the egress probe distinguishes only reached vs blocked, so it never pinpoints one layer), and cross-references the DLP checkpoint.

## Key files

- `scripts/flightcheck/live_egress_probe.py` - the only mutating path: lifecycle, reachability interpretation, orphan sweep, guaranteed cleanup.
- `scripts/flightcheck/consent.py` - consent UX single source of truth (tri-state, offer/skip/manual copy, links).
- `scripts/flightcheck/cli.py` - `--runtime-reachability` flag + `_apply_runtime_reachability_consent`; infra scope acquires just the probe''s tokens when the flag is passed.
- `scripts/flightcheck/checks/infrastructure.py` - INFRA-003 finding assembly (reachable/unreachable/manual rows, known-limitation notes).

## Testing

- Unit tests, deterministic (cassette/mock): enumeration, NOT_CONFIGURED, MANUAL paths, egress reachable/blocked (asserts the 5-field FAIL schema), indeterminate/missing-prereqs, cleanup and delete-failure resilience, `interpret_probe_response`, and the consent gate (forced on/off, interactive Y/N, non-TTY auto-skip, ADK never prompts, infra-scope offer suppression). Full suite: **932 passed**.
- **E2E manual validation against a real demo environment and agent:** confirmed all scenarios behave as expected - **PASS** (reachable endpoint), **FAIL** (invalid endpoint, no HTTP response), **NOT_CONFIGURED** (no endpoints), **NO / declined** (MANUAL guidance), and **MANUAL** (endpoint with no testable URL). The transient flow was created and deleted cleanly with no residue.

## Risks

- The egress path is mutating by design. Risk is contained: consent-gated (explicit YES only), isolated module, deterministic naming, orphan sweep before/after, guaranteed `finally` cleanup.
- Reachability is only verified when `--runtime-reachability` runs; otherwise INFRA-003 returns MANUAL guidance rather than a pass/fail verdict.
- No secrets are committed; the cassette is redacted (bearer tokens, SAS `sig=`, org/env ids, env hostnames all scrubbed).


docs/design-infra-003-endpoint-reachability.md
```diff
@@ -0,0 +1,136 @@
+# Design Decision: INFRA-003 External Endpoint Reachability
+
+Status: Accepted
+Decision: Option B, HTTP probe via transient test flow
+Date: 2026-06-29
+Owner: ESS FlightCheck
```
apurvabanka, line 6: Although I agree the doc is very insightful, I suggest we move it to the ADO ticket.

solutions/ess-maker-skills/scripts/flightcheck/cli.py
```diff
@@ -127,6 +128,84 @@
...
+
+    # The egress probe only lives inside the Infrastructure category (INFRA-003).
+    infra_in_scope = any(fn is run_infrastructure_checks for _, fn in checks)
+    if not infra_in_scope:
+        runner.runtime_reachability = flag is True
+        return
+
+    systems = _endpoint_systems_for_offer(runner)
+    label = consent.system_label(systems[0] if systems else None)
+
```
apurvabanka: The consent offer names only the first system and build_offer_prompt is singular ("a network request … to your {label} endpoint"). But check_external_endpoint_reachability iterates every endpoint (for ep in endpoints: run_live_probe(...)). So for an agent wired to Workday and ServiceNow, the user approves seeing only "Workday", yet a transient flow is also stood up to probe ServiceNow. Since informed, per-tenant-mutation consent is the central premise of this PR (Approach C), naming only one of N probed systems is a meaningful hole. Consider listing all systems in the offer, or probing only the consented one.

solutions/ess-maker-skills/scripts/flightcheck/cli.py
```diff
@@ -326,6 +405,11 @@ def _run_single_checkpoint(args):
     runner.powerplatform = powerplatform
     runner.azure_arm = None
 
+    # Resolve runtime-reachability consent for --checkpoint INFRA-003 runs too,
+    # so the mutating egress probe is reachable (and consent-gated) from the
+    # single-checkpoint path, not only the broad --scope path.
+    _apply_runtime_reachability_consent(args, runner, plan.ordered_fns)
+
```
apurvabanka: **`--checkpoint INFRA-003` single-checkpoint wiring is effectively dead**

`_run_single_checkpoint` now calls `_apply_runtime_reachability_consent(...)` with a comment "for --checkpoint INFRA-003 runs too." But `registry.py` has **no `CheckpointSpec` for any INFRA-xxx** and never imports `run_infrastructure_checks`, so INFRA-003 isn't individually targetable. Also, in that path `infra_in_scope` compares against `run_infrastructure_checks`, which never appears in `plan.ordered_fns` (those are leaf check fns), so it always takes the `flag is True` branch and never offers. The new single-checkpoint code and comment describe a path that can't currently be reached. Either register INFRA checkpoints or drop the misleading wiring/comment.

solutions/ess-maker-skills/scripts/flightcheck/cli.py
```diff
@@ -470,6 +554,21 @@ def main():
...
+        help=(
+            "Runtime-reachability egress probe for INFRA-003. Stands up a "
+            "transient test flow, triggers it once to probe each external "
+            "endpoint from the Power Platform environment's OWN egress, then "
+            "deletes it — the only FlightCheck path that mutates the tenant. "
+            "--runtime-reachability forces it on (consent given); "
+            "--no-runtime-reachability forces it off. Omit both to be asked "
+            "interactively during a normal run; non-interactive runs stay "
+            "read-only (local probe only)."
+        ),
```
apurvabanka, line 597: Stale "local probe" references contradict the PR's core change
The PR removed the local TCP/TLS probe, but four spots still tell users/readers it exists — two are user-visible:

cli.py --runtime-reachability help: "non-interactive runs stay read-only (local probe only)."
cli.py console on PP auth failure: "(Runtime-reachability probe will fall back to the local probe)" — it actually degrades to MANUAL, not a local probe.
live_egress_probe.py: LiveProbeResult and run_live_probe docstrings say "fall back to the local probe."
infrastructure.py _live_probe_context docstring: "degrades to the local probe."
These directly contradict the design ("A local TCP/TLS probe … was removed") and could mislead an operator into thinking reachability was locally verified.

solutions/ess-maker-skills/scripts/flightcheck/checks/infrastructure.py
```diff
@@ -299,6 +306,22 @@ def _host_from_url(url: str | None) -> str | None:
     return parsed.hostname or None
 
 
+def _port_from_url(url: str | None) -> int:
+    """Extract the TCP port from a URL, defaulting by scheme (https→443, http→80).
+
+    An explicit port in the URL (e.g. ``https://host:8443``) wins so endpoints
```
apurvabanka: Minor: leftover port plumbing
_port_from_url / _ExternalEndpoint.port remain from the deleted local TCP probe. port is now only used for dedup; the egress probe uses the full ep.url. Harmless but dead-ish.

apurvabanka (changes requested):

solutions/ess-maker-skills/scripts/flightcheck/checks/infrastructure.py
daeunJe0ng: I looked at removing it but kept `port`: it still keys endpoint de-duplication, and there's deliberate coverage (`test_same_host_different_port_not_deduplicated`) that two systems on the same host but different ports stay distinct and both get probed — removing it would collapse them into one. I added a comment on `_port_from_url`/the field documenting that it's retained for dedup (not the deleted local probe). Happy to switch dedup to keying on the full URL instead if you'd prefer to drop `port` entirely.

solutions/ess-maker-skills/scripts/flightcheck/cli.py
daeunJe0ng: Fixed in bb3e7c4. Corrected all of them: the `--runtime-reachability` help text, the PP-auth-failure console line, the `LiveProbeResult`/`run_live_probe` docstrings, and `_live_probe_context`. Also updated `validation-matrix.md`, which still described the old "default local TCP/TLS probe" path. They now correctly state the probe degrades to **MANUAL guidance** with no local fallback.

docs/design-infra-003-endpoint-reachability.md
daeunJe0ng: Thanks — glad it's useful. I've kept the design record in-repo for now since it's versioned alongside the code and surfaces in review/history, and it's linked from the ADO work item (7485387). Happy to also mirror it into the ticket if you'd prefer the canonical copy there.

solutions/ess-maker-skills/scripts/flightcheck/cli.py
daeunJe0ng: Good catch — fixed in bb3e7c4. The offer now names **every** discovered system via a new `consent.systems_label()` (e.g. "Workday and ServiceNow"), and the copy was reworded so it reads naturally for one or many. The skill's conversational gate was updated the same way ("name all of them, since the probe tests every configured endpoint"), so consent now covers all endpoints the probe touches.

solutions/ess-maker-skills/scripts/flightcheck/cli.py
daeunJe0ng: Fixed in bb3e7c4. Dropped the `_apply_runtime_reachability_consent(...)` call and the misleading comment from `_run_single_checkpoint`, and replaced it with an accurate note: INFRA-003 isn't individually targetable today (no INFRA `CheckpointSpec`; `plan.ordered_fns` holds only leaf check fns), so the egress probe never runs on that path — consent is resolved on the `--scope` path only.

apurvabanka (approved):

PR #198, opened 2026-07-16
fix(flightcheck): scope ENV-004 connection-reference check to the ESS…

apurvabanka:
## Problem

FlightCheck `ENV-004` judged **every** connection reference in the Power Platform
environment. On a shared tenant, a customer can have many connections/references
created for unrelated apps and use cases. Any placeholder, unbound, or orphaned
reference — even one the ESS agent never touches — produced a false **FAILED**.

This was especially noticeable on **Workday simplified deployments**, where a
reference is shipped but not used, and on environments hosting other solutions.
Customers reported ENV-004 failing on connections that have nothing to do with ESS.

## Fix

ENV-004 now scopes its verdict to only the connection references the **ESS agent
actually uses**. Resolution chain:

1. Read the agent's **enabled** topics (`botcomponents`, `componenttype eq 9`,
   filtered by `_parentbotid_value` + `statecode eq 0`).
2. Regex-extract each topic's `InvokeFlowAction` `flowId:` GUIDs.
3. For each flowId, call the **per-flow detail** endpoint (`pp.get_flow` →
   `.../flows/{id}`) and read `properties.connectionReferences`.
4. Union the resulting logical names + connectors into an `AgentRefScope`.

The env-wide connection-reference list is still fetched, but is now **filtered to
that scope** before ENV-004 forms a verdict. Out-of-scope references are dropped
(not counted, not displayed). The "N connection(s) in environment" line stays
env-wide (informational); "N reference(s) used by this agent" and the `UC` unbound-
connection check are now agent-scoped.

### Behavior changes

- `ENV-004-OR/UR/MR-nnn` (orphan / unbound / missing **in-scope** reference) → **FAILED** (unchanged severity, now scoped).
- New `ENV-004-MR-nnn` row: agent flow references a connection reference that doesn't exist in the environment.
- New **SKIPPED** outcome when the agent scope can't be resolved (no botId, no creds, no discoverable flows/refs) — avoids a silent pass or a false fail.
- New **WARNING** outcome when the flow-detail API errors (401/403 or exception) so scope can't be trusted.
- `ENV-004-UC-nnn` (unbound connection) is now limited to the agent's connectors.

### Important implementation note

The BAP flow **listing** (`.../v2/flows`) does **not** include
`properties.connectionReferences` — that block exists only in the **per-flow
detail** (`.../flows/{id}`). The scope builder calls `pp.get_flow` per flowId,
matching the existing `LIC-FLOW-001` pattern in `licensing.py`.


amilandi:
## Review — ENV-004 agent-scoping

Strong PR: the agent-scoping design is sound and well-documented, follows the tier/cassette conventions, and the full flightcheck suite (664 tests) passes locally. One real bug to fix before merge, plus minor notes.

### 🐛 Bug — a test lost its `def` and silently merged into the previous test

In `tests/flightcheck/checks/test_env_004_remediation_links.py`, the `def test_env_004_summary_orphan_remediation_links_to_solutions(monkeypatch):` signature was removed but its **body was kept**. That body now dangles at the tail of `test_env_004_unbound_connection_scoped_to_agent_connectors` (right after its last assert):

```python
    assert not any("Other App" in r.description for r in uc_rows)



    """The top-level ENV-004 row, when orphan refs are present, must
    point the operator at the Power Apps Solutions page ..."""
    from flightcheck.checks import environment as env_mod
    runner = _make_runner(connections=[_conn("real-conn-id")])
    ...
    assert summary.status == "Failed", ...
```

Consequences:
1. The orphan-ref-summary test **no longer exists as a discrete test** — its assertions run as a continuation of the connector-scoping test.
2. It runs under the **wrong scope**: that test overrode `build_agent_ref_scope` to `{msdyn_used}`, so the appended block's `_ref("missing-conn-id")` falls out of scope and the row FAILs as a **missing ref**, not an **orphan ref**. `status == "Failed"` passes for the wrong reason and validates nothing about orphan-ref remediation links.
3. The green suite masks this — the orphaned body happens to satisfy its asserts.

**Fix:** restore the function signature before that docstring:
```python
def test_env_004_summary_orphan_remediation_links_to_solutions(monkeypatch):
    """The top-level ENV-004 row, when orphan refs are present, ..."""
```
Once standalone, the `autouse` `_scope_to_all_present_refs` fixture puts the ref in scope → `missing-conn-id` isn't in `conn_map` → a genuine **orphan ref** → FAILED, which is what the test intends.

### ✅ What's good
- Correct root-cause fix: env-wide ENV-004 → agent-scoped via topics → `InvokeFlowAction flowId` → per-flow `connectionReferences`, mirroring LIC-FLOW-001. The "listing omits `connectionReferences`, detail has them" note is accurate and documented in `INDEX.md`.
- Honest failure modes: `None` → SKIP, API error → WARNING (raised, not swallowed). New `ENV-004-MR-*` missing-ref row is a real improvement.
- UC-connector scoping is consistent: both the connection `properties.apiId` and the flow `apiDefinition` pass through `normalize_connector_id` (lowercased last path segment), so `shared_x` matches across surfaces. `test_env_004_unbound_connection_scoped_to_agent_connectors` correctly asserts the out-of-connector connection is dropped.
- Good null-safety, multi-agent botId handling, and `statecode eq 0` (enabled topics only).

### Minor / optional
- **Perf**: one `pp.get_flow` per discovered flowId (serial) — fine for typical agents and consistent with LIC-FLOW-001, but worth noting if agents grow large.
- The OData filter interpolates `bot_id` directly; it's a config-sourced GUID so no practical injection concern — no change needed.

daeunJe0ng:
## [copilot-review] ENV-004 agent-scoping

Reviewed the full diff. Root-cause fix is correct: `ENV-004` moves from env-wide to agent-scoped (botId -> enabled topics -> `InvokeFlowAction` flowIds -> per-flow `connectionReferences`), mirroring `LIC-FLOW-001`. SKIP / WARNING / new `MR` paths are honest. Summary counts and scope partitioning check out.

> Building on @amilandi's review, not repeating it. The lost-`def` blocker in `test_env_004_remediation_links.py`, the serial `pp.get_flow` perf note, and the OData interpolation note are already covered there. That test bug is the one **must-fix before merge**.

### Additional notes (both optional)

| # | Sev | Location | Note |
|---|-----|----------|------|
| 1 | Low | `environment.py` (scope call) | Bare `except Exception` around `build_agent_ref_scope` masks programming errors as a benign `WARNING`. The builder already raises `RuntimeError` for real API failures and returns `None` for unresolvable scope, so consider narrowing to `except RuntimeError` (or logging the exception type) to avoid disguising future defects. |
| 2 | Nit | `environment.py` | `has_orphan_refs` now also gates on `unbound_refs` + `missing_ref_names`; `has_failing_refs` would read truer. Cosmetic. |

### Verified
- Logic, requirements, and the intentional split (env-wide "N connection(s) in environment" line vs agent-scoped verdict) are correct.
- New `_agent_connection_refs.py` has thorough, well-targeted tests.
- No security concerns.

Did **not** run the suite (branch not checked out locally); findings are from static diff review. Nice work overall.

apurvabanka:
Thanks for the thorough reviews! Addressed all three points:

🐛 Must-fix — restored the lost test def ( test_env_004_remediation_links.py )
Restored the  def test_env_004_summary_orphan_remediation_links_to_solutions(monkeypatch):  signature before its docstring. It's now a discrete test again (file collects 37, was 36). Standalone, the autouse  _scope_to_all_present_refs  fixture puts the ref in scope →  missing-conn-id  isn't in  conn_map  → a genuine orphan ref →  FAILED , so the assertion now passes for the right reason and actually validates orphan-ref remediation links.

Note 1 (Low) — narrowed the except around  build_agent_ref_scope 
Changed  except Exception  →  except RuntimeError  in  environment.py . Confirmed the builder's contract raises only  RuntimeError  for genuine API failures and returns  None  for unresolvable scope. Genuine API errors still surface as WARNING; a future programming defect now propagates and lands as a loud ERROR (via the runner's try/except) instead of a disguised benign WARNING. Added a comment explaining the rationale. The  warns_when_scope_build_raises  test raises  RuntimeError , so that path is preserved.

Note 2 (Nit) — renamed  has_orphan_refs  →  has_failing_refs 
Reads truer now that it also gates on  unbound_refs  +  missing_ref_names .

ENV-004, agent-connection-refs, and environment suites all green (54 tests).

CC: @daeunJe0ng @amilandi

daeunJe0ng (approved):
## [copilot-review] ENV-004 agent-scoping

**Verdict: Approve.** Correct root-cause fix, all earlier review points addressed, and the FlightCheck suites pass locally on the PR branch. No blockers.

**Why this PR exists (and is it the right change):** ENV-004 used to judge every connection reference in the environment, so unrelated or orphaned references (common on shared tenants and Workday simplified installs) produced false FAILEDs. This scopes the verdict to only the references the ESS agent actually uses (botId to enabled topics to `InvokeFlowAction` flowIds to per-flow `connectionReferences`), mirroring `LIC-FLOW-001`. Right approach at the right layer.

### Findings

**Blocking:** none.

**Should-fix:** none.

**Nit / optional:** none outstanding. The two optional notes from the prior round (narrow the `except`, rename `has_orphan_refs`) were both taken.

### Follow-ups verified in the current diff

| # | Point | Status |
|---|-------|--------|
| 1 (was must-fix) | Lost `def` in `test_env_004_remediation_links.py` | Fixed. File collects 37 (was 36); `test_env_004_summary_orphan_remediation_links_to_solutions` is a discrete test again and passes for the right reason. |
| 2 (Low) | Narrow `except Exception` around `build_agent_ref_scope` | Fixed. Now `except RuntimeError` with a rationale comment. Builder contract verified: flow-detail failures and 401/403 are wrapped as `RuntimeError`, so genuine API errors surface as WARNING while a programming defect propagates loudly. |
| 3 (Nit) | `has_orphan_refs` -> `has_failing_refs` | Renamed. |

### What is good

Honest failure modes (SKIP when scope is unresolvable, WARNING on API error rather than a silent pass), the new `ENV-004-MR-*` missing-ref row, consistent connector normalization, and thorough, well-targeted tests in `test_agent_connection_refs.py`.

### Verification

Checked out the PR branch and ran the tests locally.
- ENV-004 / agent-connection-refs / environment suites: green (including `test_scope_raises_on_flow_detail_auth_error` and `test_scope_raises_on_flow_detail_exception`, which back point 2).
- Full suite: 990 passed, 12 skipped, 10 failed. The 10 failures are installer-script tests (out of FlightCheck scope) and reproduce identically on `main`, so they are pre-existing and unrelated to this PR.

Nice work.

amilandi (approved):

PR #199, opened 2026-07-16
feat(ess-maker): startup auto-update nudge with one-click git pull

amilandi:
## Summary

Adds a startup **auto-update nudge** to the bundled ESS Maker VS Code extension. On activation it detects when the local ESS ADK git clone is behind `origin/main` and offers a one-click `git pull --ff-only`, so makers who cloned the kit (and have nothing prompting them to pull) stop silently running stale code.

## Screenshot
Screenshot 2026-07-16 203233

## Behavior

- **Startup check** (fire-and-forget, never blocks activation, all errors logged & swallowed): compare local `HEAD` to `git ls-remote origin main`.
- **Periodic re-check**: in addition to the startup check, the extension re-runs the check every 4 hours so long-lived sessions (VS Code left open for days) still get nudged without a restart.
- **Guarded to `main`**: developers on feature branches are never nagged and never get an unwanted pull.
- **Notification**: `Update now` / `Later`.
  - *Update now* → `git pull --ff-only` with progress; failures classified (offline / diverged / dirty / unknown) into actionable, non-technical guidance; offers **Reload Window** on success.
  - *Later* (or dismissing with the X) → re-asked on the next startup and at the next 4-hour re-check. There is intentionally **no permanent per-user "don't ask again"**, so makers stay on the latest ADK. A concurrency guard prevents the timer from stacking a second notification while one is still open.
- **Policy opt-out**: `essMaker.autoUpdateCheck` setting (default true) is the only supported opt-out, intended for IT/Intune-managed boxes.
- **Extension self-staleness**: a code pull can't refresh the installed VSIX, so after a successful pull we compare the running extension version to the repo's `package.json` and guide reinstall when needed.

## Changes

- `extension.js`: update-nudge logic (pure helpers + git/VS Code orchestration) wired into `activate` as fire-and-forget, plus a 4-hour `setInterval` re-check (disposed with the extension).
- `package.json`: `0.4.23 → 0.4.24`, new `essMaker.autoUpdateCheck` config.
- `CHANGELOG.md`: 0.4.24 entry.
- Repackaged `ess-maker-profile-0.4.24.vsix` (replaces 0.4.23) — the installer installs the newest committed VSIX, so repackaging is required to actually ship the change.
- `extension.test.js`: 15 new unit tests for the pure helpers (parseLsRemoteSha, localIsBehind, compareVersions, extensionIsStale, classifyPullError) + config test; fixed 3 pre-existing stale tests. **35/35 pass.**

## Testing

- `node extension.test.js` → 35 passed, 0 failed.
- `node --check` on extension.js and extension.test.js.
- Verified repackaged VSIX structure and that its `extension.js` is byte-identical to source.

ADO: https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7569528


daeunJe0ng (commented):
[copilot-review] Review — startup auto-update nudge

Summary: This is a well-constructed change that closes a real staleness gap for cloned installs. Approving in spirit; no must-fix issues found. All notes below are optional.

What I verified locally:
- Tests: node extension.test.js passes 35/35. node --check is clean on extension.js and extension.test.js.
- Shipped VSIX ess-maker-profile-0.4.24.vsix contains a byte-identical extension.js (SHA256 match) and reports version 0.4.24, so the repackage is correct.
- The checkForUpdate call and the 4-hour timer sit inside the workspaceFolders guard in activate, not nested in a lite/standard mode branch, so they run on every activation. The timer is disposed via context.subscriptions.

Logic: Sound. Fire-and-forget with all errors logged and swallowed. Guards are correct: only nudges on main, returns early when offline, and the _updateNudgeActive guard is set and checked synchronously with no await in between, so there is no race, and it is held across the whole prompt/pull chain so the periodic timer cannot stack a second nudge.

Security: Clean. Git runs via execFile with array args (no shell), windowsHide, a 20s timeout, and a bounded maxBuffer. No injection surface and no secrets.

Tests: Good coverage on the pure helpers, including edge cases for empty or missing shas, case and whitespace insensitivity, and all four pull-error classes.

Optional notes (none block merge):
1. Redundant -C: performUpdate calls runGit(['-C', repoRoot, 'pull', '--ff-only'], repoRoot), which passes -C repoRoot and also cwd=repoRoot. Every other call site relies on cwd alone. Consider dropping -C repoRoot for consistency.
2. Multiple windows on the same clone: If a user has two VS Code windows open on the same clone, both can fire git pull near-simultaneously and the loser hits an index.lock, which classifies as unknown. Low severity and UX-only, but worth a brief acknowledging comment.
3. Naming: localIsBehind returns true whenever the shas differ, including a diverged or ahead clone. The existing code comment already calls this out and a diverged pull fails gracefully, so this is fine as documented.
4. Mixed scope: The three pre-existing test fixes (flightcheck to setup requires) are unrelated to the feature. They fix genuinely broken tests so including them is reasonable, just flagging the mixed scope.

daeunJe0ng (dismissed):

daeunJe0ng (approved):
**[copilot-review] Approve — startup auto-update nudge**

Approving. This closes a real staleness gap for cloned installs and is well-guarded.

Verified:
- Logic is sound: fire-and-forget, all git errors logged and swallowed. Only nudges on `main`, returns early when offline, and `_updateNudgeActive` is set/checked synchronously so the 4-hour timer cannot stack a second nudge.
- New commits-behind path (`fetch` then `rev-list --count HEAD..<remoteSha>`) is non-fatal and falls back to the generic message on any failure. Singular/plural handled.
- Security: git runs via `execFile` with array args (no shell), `windowsHide`, 20s timeout, bounded `maxBuffer`. No injection surface, no secrets.
- Tests: good coverage on the pure helpers, including empty/missing SHAs and all four pull-error classes.

Optional (non-blocking):
1. `performUpdate` still passes both `-C repoRoot` and `cwd=repoRoot` to `runGit(['-C', repoRoot, 'pull', '--ff-only'], repoRoot)`; every other call site relies on `cwd` alone. Minor consistency nit.
2. Two VS Code windows on the same clone can both fire `git pull` and the loser hits `index.lock` (classified `unknown`). Low severity, UX-only.

PR #200, opened 2026-07-20
Emit tenant_name on all ADK events + connect telemetry for single-checkpoint runs

amilandi:
ADO: [7590589 — tenant_name on all charts](https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7590589) · [7587431 — connect telemetry for single-checkpoint runs](https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7587431)

## Summary

Two related telemetry improvements so **every** ADK/FlightCheck Aria tile can be filtered by tenant, and so **incremental single-checkpoint runs are no longer invisible**.

### 1. Emit `tenant_name` on all ADK events (persist-and-reuse cache)

- `flightcheck/telemetry.py`: `cache_tenant_name` / `get_cached_tenant_name` — best-effort, keyed by `tenant_id` (no cross-tenant leakage), stored in gitignored `.local/.tenant_name`.
- `adk_telemetry.py`: `set_identity` seeds the cache when a name is present; `common_dimensions` (the single choke point every `emit_*` routes through) falls back to the cache, so all ADK events carry `tenantName`.

### 2. Populate `tenant_name` at `/setup` (silent-only Graph lookup)

So the name is cached even when the maker never runs FlightCheck.

- `flightcheck/graph_client.py`: `resolve_tenant_display_name_silent(tenant_id)` — silent-only (**never** prompts), minimal `Organization.Read.All` scope, returns `""` on any failure. Factored the secure MSAL cache write into `_persist_token_cache`.
- `auth.py`: telemetry bootstrap (runs on every Dataverse sign-in incl. `/setup`) now best-effort resolves + caches the name after `start_session`. Double-wrapped so it never breaks auth.

### 3. Emit `connect` telemetry for single-checkpoint FlightCheck runs (ADO 7587431)

`cli.py --checkpoint <ID>` runs gate incremental setup/connect steps (e.g. `WD-ENTRA-CONSENT-001`) but previously emitted **no** telemetry, so they never appeared on the dashboards.

- Added a best-effort emit to `_run_single_checkpoint` mirroring the `--scope` emit (legacy `ESSMakerKit.FlightCheck.*` + `adk.*` families), gated by `--no-telemetry`, `scope=checkpoint:<ID>`. Never affects the exit code.
- Added `connect` to `--invocation-source` choices. Checkpoint-mode runs **default** to `connect` (scope runs stay `cli`), so none of the 30+ markdown call sites need editing.

No Aria cube change needed — the `invocationSource` / `tenantName` dimensions auto-enumerate new values.

### 4. Fix missing completion telemetry for FlightCheck-only Windows installs

`setup/Install-EssAdk.ps1`'s FlightCheck-only branch `return`s from inside the top-level `try` **before** the normal-path `success` completion, so a successful FlightCheck-only install never emitted an `Installer.Complete` `success` event — the `finally` net mislabeled it `cancelled` (or emitted nothing). Result: the **Installer Outcomes** and **by-Platform** tiles stayed empty while **Attempts** (Start-fed) kept incrementing.

- Emit `Complete-EssInstallTelemetry -Outcome 'success'` before the early `return` (idempotent; the `finally` `cancelled` becomes a no-op).
- Brings the Windows installer to parity with `install-ess-adk.sh`, which already records `success` explicitly before the FlightCheck-only exit.

## Testing

- New `tests/flightcheck/test_graph_client.py` (silent-only contract).
- New checkpoint-telemetry tests in `test_cli_single_checkpoint.py` (connect source + checkpoint scope, explicit source wins, `--no-telemetry` suppresses).
- New `Install-EssAdk.Tests.ps1` regression test (FlightCheck-only branch emits `success` before its early return); **43/43** installer source-assertion tests pass.
- **912 flightcheck + adk telemetry tests pass.**

## Notes

- Silent success for #2 depends on the tenant having consented to `Organization.Read.All` for the first-party Graph app; otherwise it stays empty until any interactive Graph flow runs (intended silent-only tradeoff).
- `tenant_name` is OII (privacy-approved); `tenant_id` is never persisted.

### DIFF
--- a/setup/Install-EssAdk.Tests.ps1
+++ b/setup/Install-EssAdk.Tests.ps1
@@ -304,6 +304,15 @@ Test 'Install-EssAdk.ps1 records success on the normal path, not in finally' {
         throw "finally must not force a 'success' outcome (mislabels Ctrl+C)"
     }
 }
+Test 'Install-EssAdk.ps1 records success for a FlightCheck-only install before its early return' {
+    # Regression: the FlightCheck-only branch returns from inside the top-level
+    # try before the normal-path success emit, so it must emit its own success
+    # completion. Otherwise the finally net mislabels it 'cancelled' and the
+    # Installer Outcomes / by-Platform tiles never receive a real completion.
+    if ($src -notmatch "(?s)Python not found\..*?Complete-EssInstallTelemetry -Outcome 'success'\s*\r?\n[^}]*?return") {
+        throw "FlightCheck-only branch must emit a 'success' completion before its early return"
+    }
+}
 Test 'PS emitter uses a short send timeout and trips a circuit breaker on failure' {
     $emitterSrc = Get-Content $psEmitter -Raw
     if ($emitterSrc -notmatch 'TimeoutSec 3') { throw 'PS emitter should use a short (3s) send timeout' }
--- a/setup/Install-EssAdk.ps1
+++ b/setup/Install-EssAdk.ps1
@@ -1101,6 +1101,16 @@ if ($FlightCheckOnly) {
         Write-Warn2 "  cd $workspace"
         Write-Warn2 '  python scripts/flightcheck/cli.py --scope full'
     }
+    # Record the FlightCheck-only install as a success HERE, before the early
+    # return below. This branch returns from inside the top-level try well
+    # before the normal-path success emit at the end, so without this the run
+    # has no completion event and the finally net mislabels a successful
+    # FlightCheck-only install as 'cancelled' -- leaving the Installer Outcomes
+    # / by-Platform tiles (both Complete-fed) empty while Attempts (Start-fed)
+    # keeps climbing. Idempotent: the finally's 'cancelled' is a no-op once any
+    # outcome has been emitted. (The FlightCheck run's own pass/fail verdict is
+    # separate telemetry emitted by cli.py; this outcome is the install's.)
+    Complete-EssInstallTelemetry -Outcome 'success'
     # ``return`` (not ``exit 0``) so the script ends without terminating
     # the PowerShell host. When this installer is invoked via
     # ``iex (irm .../bootstrap-flightcheck.ps1)`` from an interactive
--- a/solutions/ess-maker-skills/scripts/adk_telemetry.py
+++ b/solutions/ess-maker-skills/scripts/adk_telemetry.py
@@ -218,6 +218,12 @@ def set_identity(
     )
     _IDENTITY["tenant_id"] = tenant_id or ""
     _IDENTITY["tenant_name"] = tenant_name or ""
+    # When a Graph-capable flow (FlightCheck) resolves the org display name,
+    # persist it so ADK events emitted later in a *different* process (which
+    # only has a Dataverse/BAP token and can't resolve it) can reuse it. The
+    # cache is keyed by tenant_id in flightcheck.telemetry.cache_tenant_name.
+    if tenant_name:
+        _fc.cache_tenant_name(_IDENTITY["tenant_id"], tenant_name)
     return dict(_IDENTITY)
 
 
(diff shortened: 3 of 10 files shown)


daeunJe0ng (dismissed):
## Approve

Clean, low-risk telemetry change. `tenant_name` now flows to every ADK/FlightCheck event through the single `common_dimensions` choke point, and single-checkpoint runs are no longer invisible on the dashboards. No blocking issues.

### Why the design is sound
- Caching the name at `common_dimensions` covers every event family with one change.
- The cache is keyed by `tenant_id` with an explicit cross-tenant guard, so one org's name cannot leak onto another org's events.
- The silent-only Graph lookup avoids surprise sign-in prompts, and everything is best-effort and wrapped so telemetry can never break auth or a run.
- `--invocation-source` handling is correct: `--checkpoint` dispatches and `sys.exit()`s before the `-> "cli"` normalization, so real checkpoint runs default to `connect` (the unit tests call `_run_single_checkpoint` directly, so this depends on `main()` order, which I checked).

### Verification
- Fetched PR head `644b227` into an isolated worktree.
- Ran the 3 affected test files (`test_graph_client.py`, `test_cli_single_checkpoint.py`, `test_adk_telemetry.py`): **63 passed**.
- Confirmed `tenant_id`/`graph` are initialized to `None` at the top of `_run_single_checkpoint` (no NameError), `.local/**` is gitignored, and the `main()` dispatch order. Did not run the full suite.

### Non-blocking follow-ups (optional, not gating)
- `auth.py` resolves the name on every Dataverse sign-in. Consider short-circuiting when `telemetry.get_cached_tenant_name(tenant_id)` is already set, to skip a repeat silent MSAL call plus Graph `/organization` GET on the auth hot path.
- `resolve_tenant_display_name_silent` uses `accounts[0]`. For a maker signed into multiple tenants, consider selecting the account whose realm matches `tenant_id`.
- Please confirm the privacy approval for `tenant_name` explicitly covers the on-disk persistence (`.local/.tenant_name`) and emission across all ADK event families, not just FlightCheck outcome events.

daeunJe0ng (approved):
## Approve

Re-review after my earlier approval was auto-dismissed by two new commits (FlightCheck-only installer fix `9ad20de` + `main` merge `d6206b7`). Both verified, including the merge itself. No blocking issues.

**Why it's sound**
- `tenant_name` reaches every ADK/FlightCheck event through the single `common_dimensions` choke point; cache keyed by `tenant_id` with a cross-tenant guard (no leakage).
- Silent-only Graph lookup never prompts; all of it is best-effort and wrapped so telemetry can't break auth or a run.
- `--checkpoint` dispatches and exits before the `-> "cli"` normalization, so real checkpoint runs correctly default to `connect`.
- Installer fix is correct and safe: the FlightCheck-only branch emits `success` before its early `return`, and `Complete-EssInstallTelemetry` guards on `$script:EssTel.Completed`, so the `finally`'s `cancelled` is a genuine no-op. Now at parity with `install-ess-adk.sh`.

**Verification (merge-aware)**
- Approval is pinned to the current head `d6206b7`, which is a merge commit. The merge integrated `main` into the branch and pulled about +144 lines into `auth.py` from the `main` side, so I re-inspected the merged file: the PR's silent-Graph `tenant_name` resolution is correctly nested inside the outer telemetry try (double best-effort wrap), and `main`'s auth changes coexist cleanly. The merge resolved without clobbering either side.
- `Install-EssAdk.Tests.ps1` (the CI gate): 43/43 pass, including the new FlightCheck-only regression test (whose regex I confirmed matches the real source).
- On head `d6206b7`: `test_graph_client.py`, `test_cli_single_checkpoint.py`, `test_adk_telemetry.py`, `test_telemetry.py`, and `tests/scripts/test_auth.py`. Each affected file passes in isolation. Isolated worktree; did not run the full suite.
- One pre-existing observation (NOT introduced by this PR): `test_adk_telemetry.py::test_emit_happy_path_posts_envelope` fails only when `tests/scripts/test_auth.py` runs before it in the same session (that file leaves `adk_telemetry` module globals set). It reproduces identically on base `main` (`3c1bb61`) without this PR, so it is a pre-existing test-isolation issue in `main`. Flagging so it is not mis-attributed to #200.

**Non-blocking follow-ups (optional, not gating)**

New in this re-review:
1. `success` is emitted even when Python isn't found and FlightCheck never runs. Defensible since it's the installer's outcome, not FlightCheck's verdict (the comment says so), just confirming it's intended.
2. The pre-existing `test_auth.py` to `test_adk_telemetry.py` order-dependency noted under Verification is worth a separate cleanup (reset `adk_telemetry` module state in a fixture). Out of scope for this PR.

The three other non-blocking follow-ups (auth.py hot-path short-circuit when the name is already cached, `accounts[0]` multi-tenant account selection, and the `tenant_name` privacy-approval scope confirmation) carry over unchanged from my earlier review and are not repeated here.

PR #206, opened 2026-07-21
fix(push): Flow registration, readiness and Publishing tooling

nkemms:
## What this does

Makes `push` deploy a **fully agent-invocable** Copilot Studio flow end-to-end. Before this, `push` created flows as orphan `workflows` records — no modern-flow type, no connection, no topic link, not activated — so an agent couldn't resolve them at runtime. It also adds standalone **publish** and **readiness-validation** capabilities, a **`--repair`** path to finish/re-drive registration, and a backend-agnostic **Power Fx + Power Automate authoring reference**.

## Why

A maker (or an agent acting for them) authoring a new flow via the ADK had no path to a runtime-resolvable flow: the flow existed in Dataverse but the agent returned `flowNotFound` / CloudFlow-not-found. Completing the registration inside `push` — and giving an honest post-push signal — is what closes that gap.

## Testing

- **Unit (TDD):** pure planners/helpers for payloads, connref mirror/plan, topic-flow links, registration evaluation, `--repair` selection, and validate scoping. Full suite **1080 passed** (10 pre-existing installer-text failures are unrelated to this diff).
- **Live-verified** on a managed preprod environment: id resolution, client-GUID honored, `modernflowtype=1`, connref mirrored byte-identical to a known-good one, `--repair` connref pre-check skip + idempotent activation (exit 0), post-repair `validate.py` READY on all 5 checks, and validate's informational-vs-gating exit semantics.
- CredScan-equivalent pass; real environment identifiers scrubbed from test fixtures.

## Notes for reviewers

- Registration steps are intentionally **best-effort** (they never fail the atomic map save); `validate.py` / `--repair` are the completion path. This is the one place "correct" and "transactional" diverge — a deliberate trade-off for an external-maker CLI.
- Solution-installed orchestrators show `NOT READY` under `validate.py` by design (they register via a shared design connref, not a flow-scoped one); the no-arg run is informational for this reason.
- The new authoring doc lives under `ess-docs/customization/` alongside the other kit-authored guidance (`authoring-invariants.md`, the `conformance/*` lenses); despite the `ess-docs` README's "vendored snapshot" framing, kit-authored reference already coexists there and is wired via explicit SKILL references.


amilandi:
## Review — flow registration / readiness / publishing tooling

Strong PR overall: unusually well-documented, best-effort registration is correctly isolated from the atomic map save, and the `auth.py` additions are sound. One legitimate correctness bug worth fixing before merge; everything else I scrutinized checks out.

### 🟡 Medium — topic↔flow link is silently skipped on a GUID case mismatch (and `--repair` can't recover it)

`_plan_topic_workflow_links` decides which links to create with a **case-sensitive** membership test:

```python
flow_ids = [fid for fid in _extract_flow_ids(content) if fid in created_workflow_ids]
```

- `created_workflow_ids` is built from `create_record`'s return value, which now comes from the **`OData-EntityId` response header** (`_entity_id_from_header`) — Dataverse emits that GUID **lowercase/canonical**.
- `_extract_flow_ids` returns the topic's authored `InvokeFlowAction.flowId` **verbatim** — no case normalization.

So on a fresh flow create, if the authored `flowId` isn't lowercase, `fid in created_workflow_ids` is `False`, the `botcomponent_workflow` link is **never planned**, and there is **no error and no `registration_incomplete` entry** — the flow is left non-invocable but reported as a clean success.

`validate.py` then correctly shows `botcomponent_workflow_link` ❌ and tells the maker to run `python push.py --repair`, but repair re-derives the link from the same case-sensitive `target_ids` set (`_run_repair`), so it **silently skips too** — the advertised recovery path doesn't work for this case.

**Easy to miss because:** the *adopt-on-existing* branch sets the id to the authored-case GUID (which matches the topic), so only the **first successful create** (lowercase header id) is affected. The new tests (`TestExtractFlowIds`, `TestPlanTopicWorkflowLinks`) use all-lowercase GUIDs, so the mismatch is never exercised.

**Consistency note:** GUIDs are lowercased before comparison everywhere else in the codebase — `fetch_and_setup.py:176` (`{r.get("workflowid", "").lower() ...}`), `flightcheck/checks/_agent_connection_refs.py:130` — and even this PR's own `_connref_needs_create` uses `.casefold()`. This one path is the exception.

**Suggested fix:** normalize both sides of the test, e.g. compare `fid.casefold()` against `{wid.casefold() for wid in created_workflow_ids}` in `_plan_topic_workflow_links`, and likewise for `target_ids` in `_run_repair`. Add a mixed-case-GUID case to `TestPlanTopicWorkflowLinks`.

### ✅ Areas reviewed and found clean

- **`auth.py`:** `OData-EntityId` header parse + primary-key/`id` body fallback (fixes the old `result.get("botcomponentid", ...)` that returned `None` for `workflows`/`connectionreferences`); `associate_ref` (`/$ref` POST), `publish_bot` (bound `PvaPublish`), and `record_exists` (404 probe) all follow the existing `401 → AuthExpiredError` / `raise_api_error` / `_emit_api_call` pattern. No token logging; `verify=True` intact.
- **Dataverse payloads:** `modernflowtype=1` / category=5 / type=1 correct; connref mirror uses plain string attributes (not `@odata.bind`) correctly; OData `$filter` values consistently single-quote-escaped (`.replace("'", "''")`).
- **Activation:** `{"statecode": 1, "statuscode": 2}` is the correct Activated pair and is consistent with `_evaluate_flow_registration`.
- **`--repair` idempotency:** connref create guarded by `_connref_needs_create` (case-insensitive) + benign-duplicate handling; re-activation PATCH is harmless; correctly scoped to on-disk maker-authored flows. No double-registration/double-activation (aside from the link-case issue above).
- **`validate.py` semantics:** READY/NOT-READY conditions match `_evaluate_flow_registration`; gating (named flow or `--strict` → exit 1; bare/`--all` → informational exit 0) is self-consistent.
- **Best-effort vs. atomic map save:** registration steps append to `registration_incomplete` rather than incrementing `errors`, so they can't break the `errors == 0 and success > 0` map-save gate; the honest exit-code 2 / `REGISTRATION_INCOMPLETE` signal is wired correctly.

nkemms:
Confirmed and fixed in `42f7add`. The link test compared the authored `flowId` against `created_workflow_ids` case-sensitively with neither side normalized, so any case delta silently skipped the link (no error, no `registration_incomplete`), and `--repair` skipped identically since it shares that function.

Fix: `casefold()` both sides, and emit the id from `created_workflow_ids` so the `/$ref` targets the real record. One change covers `--repair`. Added a mixed-case-GUID test — the existing ones were all lowercase, which is why the path was unexercised.

Thanks for the precise catch.

amilandi (approved):

PR #207, opened 2026-07-21
refactor(mtk): [Human] | Dev mode, verbose session logs, framework Logger in token provider

anil-adepu:
## Description

Three improvements to the diagnostics and developer experience, applied on top
of the TASK-015 input pipeline work:

### 1. VS Code debugger support + auth scope fix (`9fa2b2c`)

- Added `.vscode/launch.json` with "MTK: Start (debug)" config — launches the
  orchestrator with `--dev` flag and breakpoints enabled.
- Fixed MSAL auth constants: client ID `51f81489...` (Power Platform CLI),
  scope `{env_url}/user_impersonation` (Dataverse delegated), tenant discovered
  from WWW-Authenticate challenge.
- `_normalize_scopes` now returns `list[str]` (what MSAL expects).

### 2. Dev mode with hardcoded config (`3d8d2a8`)

- `--dev` flag forwarded from `mtk.sh`/`mtk.ps1` to the orchestrator.
- In dev mode, reads `.local/dev-config.json` — if `token` is non-empty, skips
  MSAL interactive auth and uses the hardcoded token (paste from browser dev
  tools). If empty, falls through to interactive auth.
- `.local/` gitignored. Example config at `.local/dev-config.example.json`.

### 3. Verbose session logs (`8570ecb`)

- `session.log` now captures ALL log levels (DEBUG through FATAL) — verbose
  for ESS engineer debugging of customer issues.
- Console output still respects the level filter (INFO+ for user-friendly CLI).
- Below-threshold lines write directly to `session.log` only, above-threshold
  lines go through the TeeStream (no duplication).
- Log format updated: `[timestamp] [LEVEL] [Stage/Step] message`
- Token provider migrated from stdlib `logging` to the framework Logger
  (mandatory positional param — no fallback, no hidden state).

## Type of change
- [ ] Bug fix
- [x] Refactor / cleanup

## Testing

- `uv run pytest` — 50 passed
- `uv run ruff check .` — clean
- `uv run mypy src` — no issues

## Checklist

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,6 +6,7 @@
       "type": "debugpy",
       "request": "launch",
       "module": "service.mtk_orchestrator",
+      "args": ["--dev"],
       "cwd": "${workspaceFolder}/tools/ess-nextgen-migration-toolkit",
       "env": {
         "PYTHONPATH": "${workspaceFolder}/tools/ess-nextgen-migration-toolkit/src"
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,32 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "MTK: Start (debug)",
+      "type": "debugpy",
+      "request": "launch",
+      "module": "service.mtk_orchestrator",
+      "cwd": "${workspaceFolder}/tools/ess-nextgen-migration-toolkit",
+      "env": {
+        "PYTHONPATH": "${workspaceFolder}/tools/ess-nextgen-migration-toolkit/src"
+      },
+      "python": "${workspaceFolder}/tools/ess-nextgen-migration-toolkit/.venv/bin/python",
+      "console": "integratedTerminal",
+      "justMyCode": false
+    },
+    {
+      "name": "MTK: Run Tests (debug)",
+      "type": "debugpy",
+      "request": "launch",
+      "module": "pytest",
+      "args": ["-q"],
+      "cwd": "${workspaceFolder}/tools/ess-nextgen-migration-toolkit",
+      "env": {
+        "PYTHONPATH": "${workspaceFolder}/tools/ess-nextgen-migration-toolkit/src"
+      },
+      "python": "${workspaceFolder}/tools/ess-nextgen-migration-toolkit/.venv/bin/python",
+      "console": "integratedTerminal",
+      "justMyCode": false
+    }
+  ]
+}
--- a/tools/ess-nextgen-migration-toolkit/src/core/auth/token_provider.py
+++ b/tools/ess-nextgen-migration-toolkit/src/core/auth/token_provider.py
@@ -146,19 +146,19 @@ def _default_now() -> float:
 def _normalize_scopes(
     requested_scopes: str | Sequence[str] | None,
     default_scopes: tuple[str, ...],
-) -> tuple[str, ...]:
+) -> list[str]:
     scopes = default_scopes if requested_scopes is None else _coerce_scopes(requested_scopes)
     if not scopes:
         raise ValueError("at least one scope must be provided.")
     for scope in scopes:
         _validate_https_url(scope, "scope")
-    return tuple(scopes)
+    return list(scopes)
 
 
-def _coerce_scopes(scopes: str | Sequence[str]) -> tuple[str, ...]:
+def _coerce_scopes(scopes: str | Sequence[str]) -> list[str]:
     if isinstance(scopes, str):
-        return (_resource_to_default_scope(scopes),)
-    return tuple(_resource_to_default_scope(scope) for scope in scopes)
+        return [_resource_to_default_scope(scopes)]
+    return [_resource_to_default_scope(scope) for scope in scopes]
 
 
 def _resource_to_default_scope(scope_or_resource: str) -> str:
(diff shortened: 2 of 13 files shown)


.vscode/launch.json
```diff
@@ -0,0 +1,33 @@
...
+      "name": "MTK: Start (debug)",
+      "type": "debugpy",
+      "request": "launch",
+      "module": "service.mtk_orchestrator",
+      "args": ["--dev"],
+      "cwd": "${workspaceFolder}/tools/ess-nextgen-migration-toolkit",
+      "env": {
+        "PYTHONPATH": "${workspaceFolder}/tools/ess-nextgen-migration-toolkit/src"
+      },
+      "python": "${workspaceFolder}/tools/ess-nextgen-migration-toolkit/.venv/bin/python",
```
aditandadit, line 14: **[Suggestion · Non-blocking]**

The new "MTK: Start (debug)" config can't find the interpreter on Windows, so the debugger support this PR adds only works on macOS/Linux.

Both configs hardcode `.venv/bin/python` (L14, L28), so could we drop the `python` key and let VS Code use the selected interpreter, or handle `Scripts/python.exe`?

tools/ess-nextgen-migration-toolkit/src/core/logging/logger.py
```diff
@@ -243,11 +243,18 @@ def _write_engineer(
...
+            # Above console threshold — write to console (TeeStream mirrors to session.log)
+            stream = sys.stderr if level >= LogLevel.ERROR else sys.stdout
+            stream.write(line)
+            stream.flush()
+        elif self._log_file is not None:
+            # Below console threshold — write directly to session.log only
+            try:
+                self._log_file.write(line)
+                self._log_file.flush()
+            except Exception:  # noqa: BLE001 — DIAG-001
```
aditandadit, line 259: **[Question · Non-blocking]**

A failed write to `session.log` is swallowed, so the verbose log this PR adds can silently drop lines and hide the write bug.

The below-threshold branch catches `Exception` and passes (L259), so should it narrow to `OSError` or at least surface the failure once?

tools/ess-nextgen-migration-toolkit/src/modules/preprocessing/steps/gather_input_with_auth_step.py
```diff
@@ -29,40 +29,76 @@
...
+                pipeline_step=self.name(),
+            )
+            claims = _decode_claims(token)
+            context.tid = _as_string(claims.get("tid"))
+            context.oid = _as_string(claims.get("oid"))
+            context.upn = _as_string(claims.get("upn")) or _as_string(
+                claims.get("preferred_username"),
+            )
+            # Create a token provider that always returns the hardcoded token
+            config = _build_provider_config(environment_url)
```
aditandadit, line 80: **[Question · Non-blocking]**

The offline "paste a token to skip auth" path still fails, because it hits the network first.

Line 80 runs `_build_provider_config` -> `_discover_tenant` before bypassing MSAL, so could we skip discovery when a static token is set?

tools/ess-nextgen-migration-toolkit/src/modules/preprocessing/steps/gather_input_with_auth_step.py
```diff
@@ -29,40 +29,76 @@
...
+            token_provider = MsalTokenProvider(config, self._logger)
+            # Monkey-patch get_token to return the static token for dev
+            token_provider.get_token = lambda *_a, **_kw: token  # type: ignore[method-assign]
```
aditandadit, line 83: **[Suggestion · Non-blocking]**

Patching `get_token` on a live provider is a footgun the next dev will trip on.

The dev branch overwrites it with a lambda and a type-ignore (L83), so would a small `StaticTokenProvider` be cleaner to hand to `DataverseClient`?

tools/ess-nextgen-migration-toolkit/src/core/auth/token_provider.py
```diff
@@ -72,15 +71,15 @@ class MsalTokenProvider:
     def __init__(
         self,
         config: MsalTokenProviderConfig,
+        logger: Logger,
```
aditandadit, line 74: **[Question · Non-blocking]**

A missed caller would bind `logger` to the wrong argument with no type error.

`logger` is now a required 2nd positional (L74), so is anything outside this diff still constructing it positionally?

aditandadit (approved):
nothing blocking, feel free to take the questions and suggestions per your judgement - approving

tools/ess-nextgen-migration-toolkit/src/core/auth/token_provider.py
anil-adepu: Nope

tools/ess-nextgen-migration-toolkit/src/modules/preprocessing/steps/gather_input_with_auth_step.py
anil-adepu: Will clean it in follow up. I don't foresee multiple devs working on the MTK module in parallel though

tools/ess-nextgen-migration-toolkit/src/modules/preprocessing/steps/gather_input_with_auth_step.py
anil-adepu: Yeah monkey patching was a hacky logic, will be addressed in follow up as it's in dev mode

.vscode/launch.json
anil-adepu: The debugger support is only internal and not targeted to customers.

PR #208, opened 2026-07-21
flightcheck: autolink bare URLs in report remediation

daeunJe0ng:
## Summary

FlightCheck report `remediation` text supports two link forms: markdown `[label](url)` and bare `https://…` URLs. Only the markdown form was linkified. Bare URLs were HTML-escaped and rendered as plain text, so they showed up but were not clickable. Because different checks author URLs differently, operators saw links that worked on some rows and not others.

The concrete live case is the **WD-CONN** connection-health check (`checks/workday.py`), whose WARNING and FAILED remediations embed a bare `make.powerautomate.com/environments/<env>/connections` URL built via f-string. That URL is the operator's path to fix a broken Workday connection, and today it renders as non-clickable plain text.

This PR makes both link forms render as clickable anchors, and additionally stops the report from masking GUIDs **inside** those URLs (which was rewriting some deep-link targets into broken links on MANUAL checks).

### Autolink bare URLs
- `_md_links_to_html` (in `runner.py`) now runs a second pass, `_autolink_bare_urls`, after the markdown pass.
- The bare-URL pass steps over anchors created by the markdown pass, so `[label](url)` is never double-wrapped.
- Trailing sentence punctuation (`.,;:!?`) and an unbalanced closing paren are trimmed off the link target, so `see https://aka.ms/x.` keeps the period outside the `href`. Balanced parens are preserved, so `.../Foo_(bar)` stays intact; only an unbalanced trailing `)` (as in `(see https://x)`) is pushed back into the text. A `;` that terminates an HTML entity (`&amp;`) is not stripped, so query strings stay whole.
- A trailing escaped **delimiter** entity (`&lt;`, `&gt;`, `&quot;`) is peeled back into the text. An angle-bracket-wrapped URL `<https://x>` escapes to `&lt;https://x&gt;`; without this, the trailing `&gt;` was pulled into the `href`. An unescaped `<`, `>` or `"` is never a valid URL character, so the wrapping bracket/quote stays as text and only the URL is anchored. A real query-string `&amp;` is still kept whole.
- `_multiline_html` and the manual checklist renderer call `_md_links_to_html`, so they inherit the fix with no separate change.

### Keep GUIDs inside deep-link URLs unmasked (addresses review note 1)
- `_mask_sensitive` redacts operator-identifying values (email/UPN local parts, resource GUIDs) from remediation text on the MANUAL render path. It was also masking GUIDs that are **path segments of a deep link** (e.g. Copilot Studio `.../environments/<env_id>/bots/<bot_id>/overview`), rewriting the `href` so the link stopped resolving. The publishing QA-* checks emit their Copilot Studio deep link as MANUAL, so those links were masked into broken hrefs.
- `_mask_sensitive` now splits the text into `http(s)://` URL spans and prose, and masks **only the prose**. URL spans are kept verbatim, so deep-link GUIDs survive and the link resolves. Standalone GUIDs and emails/UPNs in prose are still masked exactly as before. Those `env_id`/`bot_id` ids already appear in every deep link the report emits (including unmasked FAILED/WARNING cards), so keeping them in the URL leaks nothing new.

## Design rationale

The fix lives in the renderer, not in each check. That fixes every existing report immediately and means check authors cannot reintroduce a non-clickable link by pasting a raw URL, or a broken deep link by embedding a GUID. The alternative (a lint rule forcing markdown links) would leave current reports broken and add an ongoing authoring burden. The renderer is the single choke point every remediation string already passes through, so it is the correct layer for both the autolink pass and the masking fix.

## Testing

- `tests/flightcheck/test_runner_link_rendering.py` (15 tests): bare URL becomes an anchor, trailing punctuation and unbalanced parens excluded, balanced parens preserved, query-string `&` stays HTML-safe and a trailing `&amp;` is not split, angle-bracket-wrapped and quote-wrapped URLs exclude the escaped delimiter, markdown links still work and are not double-wrapped, mixed markdown + bare, plain text untouched except escaping, `_multiline_html` linkifies and keeps line breaks, plus end-to-end check cards render clickable links.
- `tests/flightcheck/test_runner_report_layout.py` (3 new tests): GUID inside a bare deep-link URL is preserved while a standalone prose GUID in the same string is still masked; GUIDs inside a markdown link target are preserved; end-to-end MANUAL check renders a deep-link `href` with real GUIDs while a UPN in the same remediation is still masked.
- Full FlightCheck suite (`tests/flightcheck`): 1014 passed.

## Scope note

This PR makes bare URLs **clickable** and keeps deep-link URL GUIDs intact. It does not check that a link's host is real/intended (typo'd or hallucinated domains) — that host-level hygiene gate is #209.

## Risks and assumptions

- Change touches shared report rendering, used by every check. Mitigated by the full-suite run and by keeping the markdown behavior and prose-masking behavior identical (verified by the no-double-wrap test and the existing masking tests).
- Assumes remediation URLs are `http`/`https` only. Other schemes are not autolinked, which is intentional for a security-reviewed report surface.
- The masking fix targets URLs only. A GUID inside a non-URL context (e.g. a PowerShell `-EnvironmentName <guid>` snippet in a remediation) is still masked; that is a separate concern from clickable links and is left untouched here.


apurvabanka:
**No code blockers.** The only thing gating merge is the required approving review (`REVIEW_REQUIRED` / `BLOCKED`) — everything else is ready.

A few non-blocking notes / gaps worth acknowledging:



1. **GUID masking inside deep-link URLs (please confirm):** `_mask_sensitive` can rewrite a GUID embedded in a link target, which alters the `href`. The PR calls this out as out of scope, but the WD-CONN remediation URL this change is meant to fix contains an environment GUID — so a masked link could point somewhere invalid. Can we confirm that's acceptable for the live case, or is it worth a follow-up?

2. **Host validation** — bare URLs are now clickable but not validated (typo'd/hallucinated domains). Already tracked as #209, just noting it here.

3. **Minor untested edge case:** an angle-bracket-wrapped URL like `<https://x>` escapes to `&lt;https://x&gt;`, and the regex would pull `&gt;` into the `href`. Unlikely in remediation text and low risk — flagging for awareness only.


@daeunJe0ng

daeunJe0ng:
Thanks @apurvabanka. Went through all three:

**1. GUID masking inside deep-link URLs.** Verified the live case does not hit this. The WD-CONN remediation you flagged is WD-CONN-101 ("Workday connection token health"), which emits `Status.FAILED` (connection referenced by a flow) or `Status.WARNING` (orphan). `_mask_sensitive` runs only on `Status.MANUAL` remediations, through `_render_manual_checklist`. FAILED/WARNING remediations render via the unmasked `else` branch in `_render_check_card`, so the `make.powerautomate.com/environments/{env_id}/connections` link keeps its real env GUID and resolves. The masking-vs-deeplink interaction is only reachable if a MANUAL-status remediation embeds a GUID deep-link, which is not the live case. Keeping it out of scope for this PR as documented, no follow-up.

**2. Host validation.** Agreed, tracked as #209. Out of scope here.

**3. Angle-bracket-wrapped URL (`<https://x>` escaping to `&lt;https://x&gt;`).** Now handled and tested. `_autolink_bare_urls` peels a trailing escaped delimiter entity (`&lt;`/`&gt;`/`&quot;`) back into the text before anchoring (commit 70d469d, "exclude escaped angle-bracket/quote from autolinked URL"). Covered by `test_angle_bracket_wrapped_url_excludes_brackets` and `test_quote_wrapped_url_excludes_quotes` in `test_runner_link_rendering.py`, 15/15 passing. The regex no longer pulls `&gt;` into the href.

No code changes needed from these notes. Thanks for the review.

daeunJe0ng:
Update on note 1 (GUID masking inside deep-link URLs): rather than leave this out of scope, I fixed it in this PR. _mask_sensitive now keeps http(s):// URL spans verbatim and masks only prose, so MANUAL-check deep links (e.g. the publishing QA-* Copilot Studio links, whose env_id/bot_id GUIDs were being masked into broken hrefs) resolve again. Standalone GUIDs/emails in prose are still masked. Added 3 tests in 	est_runner_report_layout.py; full 	ests/flightcheck suite is 1014 passed. PR description updated.

apurvabanka (approved):

PR #209, opened 2026-07-21
flightcheck: offline https link gate + run FlightCheck suite in CI

daeunJe0ng:
## Summary

Adds a deterministic, offline CI gate on FlightCheck report links, and wires the previously-unrun FlightCheck test suite into CI.

FlightCheck emits URLs into the HTML report (`doc_link` buttons and remediation deep-links). This PR adds one durable, offline-checkable guarantee about them: a URL an operator is meant to fetch uses `https`. It also makes the whole FlightCheck pytest suite run on every PR, which it did not before.

> Reworked from the original host-allowlist approach. See "What changed vs. the first version" below.

## What this does

- `tests/flightcheck/url_hygiene_rules.py`: the rule set. One fail-closed rule: every fetchable URL must be `https`, unless it is a known namespace / identifier URI (SOAP / SAML namespaces, the Workday issuer identifier) that is legitimately `http`. Exemptions are matched at **URI granularity, not bare host**, so a dual-use host like `www.workday.com` (both a SAML issuer identifier and a real fetchable domain) is exempt only in its issuer-identifier shape. A structural host-shape helper enforces DNS label rules on captured hosts.
- `tests/flightcheck/test_url_hygiene.py`: scans FlightCheck source (`.py` plus `.yaml`/`.yml`/`.json` data files, stdlib only, no imports of check modules, no network) for every `http(s)` URL and asserts each fetchable URL uses `https` and each static host looks like a real domain. A test also asserts every `http` exemption corresponds to a real namespace/identifier URI in source, so a dead or lazily-added exemption fails the suite.
- `.github/workflows/ci.yml`: replaces the old single-file job with `flightcheck-tests`, which installs runtime + test deps and runs the **entire** offline FlightCheck suite. None of these tests ran in CI before, so this is where the real enforcement now lives.

## What changed vs. the first version

The first version required every host appearing in FlightCheck source to be registered in a curated allowlist (`url_registry.py`). That was dropped because:

- **Self-referential.** The same author who writes a URL also edits the allowlist, so a typo copied into the allowlist passes. It only caught "typo the URL but forget the registry".
- **High-maintenance and false-positive-prone.** Every new legitimate host, including hosts that only appear in docstrings and never reach a report, was a red build until hand-registered. It flagged legitimate hosts like `schema.management.azure.com` (an ARM namespace URI) and `www.microsoft.com` (the Azure service-tags download page).
- **The bug it targeted is rare** and already covered by human review plus the repo's "No fabricated URLs" rule.

The fail-closed https rule keeps the one property worth enforcing offline, covers new hosts by default with no upkeep, and stops flagging legitimate hosts.

## Scope and honest limits

- This gate checks link **scheme** (https) and basic host shape. It is offline and deterministic, so it is safe to gate every PR on.
- It does **not** make links clickable — the report renderer does that (#208).
- It does **not** verify a path is live. A well-formed `https://learn.microsoft.com/<moved-article>` that now 404s or redirects to the docs home still passes. That "stale path" problem needs a live, networked checker, intentionally kept out of this deterministic suite.
- It scans the **text** of source (string literals, docstrings, and comments alike), not only URLs proven to reach the report. It is fail-closed: any insecure `http` URL in scanned source fails, including one written in a docstring. Distinguishing an emitted URL from an illustrative one would need full AST dataflow analysis, deliberately out of scope for this guard.
- Known blind spots, stated plainly: a scheme split across a concatenation (`"http://" + host`) cannot be seen (a templated host over http, `http://{...}`, **is** caught); Markdown docs are not scanned; and offline we cannot fully prove an exempted host is a non-fetchable namespace (mitigated by URI-granular matching plus the exemption-justification test).

## Testing

- `python -m pytest tests/flightcheck/test_url_hygiene.py -q` -> 10 passed.
- `python -m pytest tests/flightcheck -q` (the full suite the CI job runs) -> 1044 passed.
- `ruff check` on both changed files -> clean.
- `ci.yml` parses as valid YAML.

## Review response (apurvabanka)

Second-round review addressed in the follow-up commit:

- **Coarse host exemption (#2) / honor-system exemption (#5):** exemptions moved from bare-host to URI granularity. A dual-use host (`www.workday.com`) is exempt only in its issuer-identifier shape; a fetchable `http://www.workday.com/<page>` is now caught. Removed the dead `sts.windows.net` entry (only ever appears as https). Added a test asserting every exemption maps to a real namespace/identifier `http` URI in source.
- **Only `.py` scanned (#3):** scan extended to `.yaml`/`.yml`/`.json` data files under the source dir. Markdown docs stay excluded (documentation, not report-rendered).
- **Dynamic scheme blind spot (#4):** a templated host over http (`http://{org}.crm.dynamics.com`) is now flagged. A later commit (`67820b3`) fixed the scan regex, which had split the host at the first `}` and skipped it, so this is now caught at the scan level (with a scan-level test), not just asserted on the rule helper. The remaining `"http://" + var` concatenation blind spot is documented explicitly.
- **Lax host shape (#6):** `looks_like_host` now rejects underscores and leading/trailing hyphens per DNS label rules.
- **Scans all text, not only emitted URLs (#1):** kept the text scan by design and reworded the docstrings + this description to say so honestly — the gate is fail-closed on any insecure `http` URL in scanned source, including docstrings/comments. Isolating only report-emitted URLs would need AST dataflow analysis, out of scope.

## Relationship to #208

Complementary, not overlapping. #208 makes report URLs **clickable**; this PR makes sure a fetchable link uses **https** and runs the FlightCheck suite in CI.


apurvabanka:
### Design/correctness gaps


**1. The scan doesn't isolate "URLs emitted into the report" — it scans *all* text in every `.py`, including comments, docstrings, and namespace constants.**

The PR sells the https-rule as fixing the allowlist's false-positive problem, but this class of false positive remains. I confirmed the regex captures illustrative examples: `See http://www.evil.com in a docstring` → `('http','www.evil.com')`. An author who writes an *example* of a bad `http://…` config in a docstring will red-build the suite even though nothing emits it. The exemption list (`schemas.xmlsoap.org`, `docs.oasis-open.org`, `www.workday.com`, `sts.windows.net`) *only exists* because the scan indiscriminately picks up SOAP/SAML namespace constants — i.e., the design is already compensating for over-broad scanning rather than targeting emitted URLs.


**2. Per-host exemption is too coarse → blind spot.** `www.workday.com` is exempted as a *whole host*, but it's both a namespace/entity-ID identifier **and** a real fetchable domain. A future genuinely-fetchable `http://www.workday.com/<page>` an operator is meant to click would pass the gate as fine. The "namespace URIs may be http" property is per-URI/path, but it's enforced at host granularity. Same concern for `sts.windows.net`.


**3. Only `.py` files are scanned.** FlightCheck's source dir also has `data/*.yaml` and `.md` files (e.g. `data/flow_licensing_skus.yaml` contains a `learn.microsoft.com` URL). If a URL is ever rendered from a data file, the gate can't see it. The gate covers less than the "every fetchable URL FlightCheck emits" claim.


**4. Dynamic/concatenated schemes are silently skipped.** `f"https://{base}.crm.dynamics.com"` captures only `{base}` (skipped), so the real host after the template is never checked — `f"http://{sub}.evil.com"` would be entirely invisible. `"http://" + host` matches nothing. Acknowledged partly, but worth stating the blind spot explicitly.


### Weaker guards


**5. The exemption is honor-system.** Nothing stops someone adding a normal fetchable host to `HTTP_ALLOWED_IDENTIFIER_HOSTS` to silence a failure — no test asserts exempted hosts are actually namespace/identifier URIs. Given the PR's own critique of the old allowlist being "self-referential," this exemption has the same weakness at smaller scale.


**6. `looks_like_host` is lax.** It allows `_` in labels (not DNS-valid) and doesn't reject leading/trailing hyphens (`-a.b`). Minor.


### Non-issues I checked (so you don't have to)



- `actions/checkout@v6` / `setup-python@v6` — **consistent** with the repo's existing `ci.yml` (lines 21/24), not broken.

- `solutions/ess-maker-skills/scripts/requirements.txt` and the flightcheck source path both exist.

- `[tool.setuptools] py-modules = []` correctly avoids flat-layout auto-discovery.

- `live` tests are skipped by default via the existing marker config — CI won't hang on network. (Corollary worth noting: any future `live`-marked test never runs in CI at all.)

@daeunJe0ng

daeunJe0ng:
Thanks @apurvabanka, sharp review. Pushed a follow-up (`3fa14ad`) addressing it:

- **#2 / #5 (coarse host exemption + honor-system):** exemptions now match at URI granularity, not bare host. `www.workday.com` is exempt only in its issuer-identifier shape (`http://www.workday.com/<tenant>`); a fetchable `http://www.workday.com/<page>` is now caught. Removed the dead `sts.windows.net` entry (only ever appears as `https` in source). Added a test asserting every exemption maps to a real namespace/identifier `http` URI in source, so a dead or lazily-added exemption fails the suite.
- **#3 (only `.py` scanned):** scan now covers `.yaml`/`.yml`/`.json` data files under the source dir. `.md` docs stay out (documentation, not report-rendered).
- **#4 (dynamic scheme blind spot):** a templated host over http (`http://{...}`) is now flagged. The `"http://" + var` concatenation split is a genuine blind spot I can't see statically, so I documented it explicitly rather than paper over it.
- **#6 (lax `looks_like_host`):** now rejects `_` and leading/trailing hyphens per DNS label rules.
- **#1 (scans all text, not only emitted URLs):** correct, it scans everything including docstrings/comments. I kept that on purpose (an insecure `http` example shouldn't live in source either), and isolating only report-emitted URLs needs AST dataflow, heavier than this guard warrants. Reworded the docstrings + PR description to state this honestly.

Full offline suite 1043 passed; ruff clean. Thanks for pre-checking the non-issues too.

daeunJe0ng:
Follow-up on #4: correction. When I said the templated host over http was flagged, that held only for the rule helper (`is_allowed_http_uri`), not the actual file scan. The scan regex split the host at the first `}`, so `http://{org}.crm.dynamics.com` captured just `{org}` (dotless) and was skipped. Fixed in `67820b3`: the host group now matches a run of literal + `{template}` segments, so a templated host with a dotted suffix is captured whole and flagged at scan level, with a scan-level test proving it (and placeholders like `https://host:8443` still drop out). Offline suite 1044 passed, ruff clean. Thanks for the push on this one.

apurvabanka (approved):

PR #213, opened 2026-07-24
Fix FlightCheck Entra SSO app selection

apurvabanka:
## Summary
Lets standalone FlightCheck target a specific Entra SSO app / connection, plus
a related WD-PKG-001 prompt fix.

## Changes
- **Entra SSO app / connection selection:** standalone-scope FlightCheck now lets
  the operator choose which Entra SSO app / connector connection to validate via
  `--select-targets` (`auto` | `always` | `never`), instead of validating all.
- **Stored SSO app confirmation:** when a Workday SSO app is already persisted,
  the run confirms/reminds the operator before reusing it.
- **Connection pin:** `check_connector_connections` can narrow to the selected
  connection (e.g. dev/test/prod) so tenants with multiple connections for the
  same connector report on just the one being verified.
- **Installer:** FlightCheck invocation now passes `--select-targets always`
  (`Install-EssAdk.ps1`, `install-ess-adk.sh`).
- **WD-PKG-001 prompt fix:** `--checkpoint WD-PKG-001` no longer blocks on the
  "Test Employee ID" / ISU credential prompts. Interactive Workday runtime prompts
  are now gated to runs that actually consume them (`WD-WF*` and full/scope runs).
- **Tests:** added coverage for target selection, Entra app resolution, stored-app
  confirmation, and WD-PKG-001 prompt gating.


apurvabanka:
@copilot resolve the merge conflicts in this pull request

nkemms:
Reviewed the full diff and ran the tests — **LGTM**.

Clean, defensive change: every narrowing path (connection pin, SAML app scope, target selection) fails open, so a stale/typo'd pin never masks a real problem. Traced the WD-PKG-001 prompt-gating fix — `scope="checkpoint:<target>"` parses correctly, so `WD-PKG-001` no longer prompts while `WD-WF*` still does. Selection runs after auth, before checks.

PR's 4 new test files: 100 pass. (The 8 `test_workday_personal_data.py` failures reproduce identically on `main` — pre-existing Python 3.14-vs-3.12 env drift, not this PR.)

Two optional NITs, non-blocking: `_prompt_choice` returns "All" if a chosen candidate has an empty `appId`/`name` (unreachable, fails open); the `always` docstring vs the single-candidate picker-skip is a minor wording nuance.

nkemms (approved):

solutions/ess-maker-skills/scripts/flightcheck/checks/workday.py
```diff
@@ -2838,6 +2838,47 @@ def _check_saml_certificate_health(runner) -> list[CheckResult]:
...
+    # the SAME ``_workday_hints`` path AUTH-005 / WD-ASSIGN-001 already use so
+    # every Workday-SSO-app check agrees on the target. When the pin matches a
+    # discovered SAML SP we narrow to it; when it matches none (e.g. the
+    # operator pinned the OAuth Workday app, which is not a SAML app) we keep
+    # the full set and say so rather than silently validating an unrelated
+    # sibling. No pin ⇒ unchanged all-apps behavior (single-checkpoint mode
+    # never sets one).
+    from ._workday_app_assignment import _workday_hints
+
+    app_id_hint, _ = _workday_hints(getattr(runner, "config", None))
```
nehaoss, line 2858: This new `_workday_hints`-based narrowing isn't actually gated to scope-mode runs the way the comment above claims ("single-checkpoint mode never sets one" / "No pin ⇒ unchanged all-apps behavior"). `_workday_hints` falls back to `.local/connect/workday/config.json` on disk, and `--checkpoint WD-CONN-102` sets `runner.config` too (`_run_single_checkpoint`), so a targeted `--checkpoint WD-CONN-102` run will silently narrow to the persisted `entraAppId` whenever one exists — which is the normal post-setup case. That's a behavior change for checkpoint mode: if the persisted app is stale/wrong, WD-CONN-102 now scopes away the actually-broken app and can report green. Please either gate the narrowing to the scope-mode selection (so `--checkpoint` keeps the old all-apps behavior), or drop the "single-checkpoint mode never sets one" claim and document that checkpoint runs narrow via the persisted hint.

PR #215, opened 2026-07-24
ci: pin ruff (version + rule set) and add pre-commit lint so an upstream ruff release can't fail every PR

anil-adepu:
## Problem

The repo-wide **CI → "Lint Python"** job (`.github/workflows/ci.yml`) runs on every
PR to `main` and installed ruff via an **unpinned** `pip install ruff`. Ruff 0.16.0
broadened its default rule set, so the job silently picked it up and flagged ~265
pre-existing findings in code no PR had touched — turning **every open PR red**,
including docs-only ones. It was non-deterministic by construction: identical commits
passed before the ruff release and failed after.

## Fix

Make lint deterministic + reproducible, and shift it left.

1. **Pinned rule set** — `pyproject.toml` gets an explicit `[tool.ruff]` selecting
   ruff's classic default `["E4", "E7", "E9", "F"]` (`target-version = "py311"`) —
   exactly what the code is clean under. Rules no longer drift with the tool version.
2. **Pinned version** — `ci.yml` installs `ruff==0.16.0`.
3. **Pre-commit hook** — new repo-root `.pre-commit-config.yaml` runs ruff (pinned to
   `v0.16.0`) over `solutions/`, `tests/`, and `tools/`, mirroring the CI lint so
   contributors catch it before pushing. Enable with `pre-commit install`.
   (`.gitignore` allow-lists the config; the repo root uses an ignore-all-then-allow
   pattern.)
4. **Lint debt** — broadening the hook to `tests/` surfaced 26 pre-existing findings
   (unused imports, redundant `f`-string prefixes, an ambiguous variable name `l`, one
   dead local, one import-not-at-top); all fixed. No behavior change.

## Scope / safety

- The **ESS NextGen Migration Toolkit** (`tools/ess-nextgen-migration-toolkit/`) is
  **excluded** from the root hook and unaffected by the root ruff config: it has its
  own `[tool.ruff]`, its own uv-locked ruff (0.15.20), and its own pre-commit config.
  Ruff resolves the nearest config, so its stricter rules still apply to its own tree.
- The rule set is kept intentionally minimal to restore green **without** rewriting
  findings in another solution. Grow `select` deliberately, alongside the fixes —
  never via a silent ruff upgrade.
- **No dependency changes** — `requirements-dev.txt` / `uv.lock` / build config are out
  of scope for this PR.

## Verification

- CI-equivalent ruff run over `solutions/`, `tests/`, `tools/` (excl. toolkit) →
  **clean (exit 0)**.
- `pre-commit run ruff-check --all-files` installs the pinned ruff and **passes**.
- MTK toolkit gates still green; config isolation confirmed (toolkit files get their
  stricter `B`/`UP`/`I`/`T20` rules; the covered trees get `E4/E7/E9/F`).

## Files
- `.github/workflows/ci.yml` — pin `ruff==0.16.0`
- `pyproject.toml` — explicit `[tool.ruff]` rule set
- `.pre-commit-config.yaml` — new pinned, scoped ruff hook
- `.gitignore` — allow-list `.pre-commit-config.yaml`
- `tests/**` — 26 lint fixes to pass under the pinned rule set

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,10 @@ jobs:
         python-version: '3.11'
 
     - name: Install ruff
-      run: pip install ruff
+      # Pinned for reproducibility: an unpinned install silently upgrades and a
+      # Ruff release that broadens the default rule set would fail every PR. The
+      # rule set itself is pinned in pyproject.toml ([tool.ruff.lint] select).
+      run: pip install ruff==0.16.0
 
     - name: Discover Python source paths
       id: discover
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@
 !setup/
 !no-commit/
 !pyproject.toml
+!.pre-commit-config.yaml
 !requirements-dev.txt
 !dev-specs/
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,36 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+#
+# Repo-root pre-commit hooks. Lints the repo's Python sources under solutions/,
+# tests/, and tools/ with Ruff, mirroring the "Lint Python" job in
+# .github/workflows/ci.yml so contributors catch lint failures locally before
+# pushing — the same gate, one commit earlier.
+#
+# Ruff is pinned to the SAME version as CI (see .github/workflows/ci.yml) and
+# reads its rule set from pyproject.toml ([tool.ruff.lint] select), which is
+# pinned to Ruff's classic default (E4/E7/E9 + F). Pinning both the version and
+# the rule set keeps lint deterministic: a future Ruff release that broadens the
+# default rule set can't silently fail every PR (which is exactly what happened
+# and prompted this config).
+#
+# EXCLUDED: tools/ess-nextgen-migration-toolkit/ is intentionally NOT covered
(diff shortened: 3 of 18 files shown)


daeunJe0ng (approved):
Approve. Well-scoped, deterministic CI/lint fix that solves the right problem at the right layer.

What's good:
- Root-causes the flakiness: unpinned pip install ruff picked up 0.16.0's broadened default rule set and turned every PR red. Pinning both the tool version (ci.yml + pre-commit) and the rule set (pyproject [tool.ruff.lint] select = E4/E7/E9/F) makes lint reproducible.
- Shifts the same gate left via pre-commit, with clear rationale in comments.
- Correctly isolates the ess-nextgen-migration-toolkit (its own config/version wins via nearest-config).
- Test-file cleanups are all ruff-verified safe: removed imports/locals are genuinely unused (e.g. the three _acquire_token_* funcs delegate HTTP to _post_token_attempts, which keeps its own import requests), and l -> line removes an ambiguous name. No behavior change.

Verification: all 7 checks green (Lint Python, CodeQL, Validator unit tests, Static validation, Installer smoke tests, Analyze python, CLA). I did not run the suite locally.

Nits (non-blocking):
- pre-commit iles: scopes solutions|tests|tools; confirm that matches CI's discovered path set so the two gates can't diverge.
- Consider a short note in CONTRIBUTING pointing devs at pre-commit install.

PR #218, opened 2026-07-29
flightcheck: add TOPIC-020 — verify handoff agent id is not the shipped placeholder

### LINKED ISSUE #152: FlightCheck: Add check for unreplaced placeholder agent identifiers in enabled auto-handoff topic

**Frequency in analyzed window:** 1 of 121 analyzed customer incident(s)
**Estimated applicability:** Conditional on auto agent handoff being configured/enabled
**Affected feature:** Auto agent handoff (HR/IT agent-to-agent handoff template topic)

## What should be checked

ESS ships a **template auto-handoff topic** that orchestrates handoff to another
agent. The topic contains placeholder target-agent identifiers (agent name / schema
name / app id) that the maker must replace with real values for their environment.

When a maker *enables* the handoff template topic **without replacing the
placeholder agent identifiers**, the handoff orchestration signal fails **silently**
— the agent does not hand off and users get a broken experience, with no error
surfaced at publish time. Because the topic is syntactically valid, publishing
succeeds and the misconfiguration is only discovered at runtime in production.

FlightCheck should validate that, **if** an auto-handoff topic is present and
enabled, its target-agent identifier fields have been changed from the shipped
placeholder values to concrete, non-placeholder values.

Note: the existing `CONFIG-014` check inspects topic **Description** fields for
placeholder text (`TODO`, `TBD`, `[describe ...]`, `<placeholder>`). It does **not**
inspect the **handoff target-agent identifier** fields inside the handoff topic's
orchestration logic, so this gap is not covered by `CONFIG-014`.

## Suggested implementation

- **Checkpoint ID:** TOPIC-020
- **Category:** Topics / Configuration
- **Priority:** HIGH
- **Conditional?** Yes, only runs when an auto-handoff template topic is present and enabled in the agent.
- **Validation logic:** Locate the auto-handoff topic in the local agent topic
  files (and/or Dataverse). If the topic state is enabled, parse the action(s)
  that set the handoff target-agent identifier(s) and confirm the value is not a
  shipped placeholder (e.g. `<agent-id>`, `PLACEHOLDER`, `[target agent]`, the
  default template GUID, or an empty/unset value). Fail with guidance to copy the
  template topic, replace the placeholder agent identifiers with the customer's
  real target agent, and re-enable.
- **Gating logic (if Conditional):** Skip entirely if no auto-handoff topic is
  detected, or if the handoff topic is present but disabled.

## Verification

- **OOTB requirement?** No, conditional on auto agent handoff being configured.
- **ESS docs reference:** Auto agent handoff customization (Copilot Studio handoff topic customization).
- **ESSVivaCopilot code path:** Handoff orchestration topic shipped in the ESS solution template.
- **ESS Toolkit sample reference:** n/a — applies to any agent that enables the handoff template topic.
- **Existing FlightCheck pattern to mirror:** `CONFIG-014` (placeholder detection in topic descriptions) for the placeholder-matching mechanics; `SN-FLOW-*` / `WD-FLOW-*` for the "only validate when the feature/topic is enabled" conditional-gating pattern.
- **Why a FlightCheck check is appropriate (vs. docs-only fix or no action):** The failure is silent at publish time and only manifests at runtime in production, which is exactly the pre-deployment class of misconfiguration FlightCheck exists to catch. It is generalizable to any deployment that enables the handoff template topic.

## Context

daeunJe0ng:
Implements the FlightCheck check requested in #152.

**Checkpoint:** `TOPIC-020`

**Category:** Agent Handoff

**Priority:** High

**Conditional?** Yes â€” fires only for ENABLED handoff topics (`status == "Active"`). Not-configured PVA, missing bot id, fetch errors, and no-enabled-handoff-topic all return no results.

**What identifies a handoff topic:** a topic contains a `SetVariable` node assigning `Topic.HandoffAgentId`. Identity is that variable, NOT the schemaName. Makers clone and rename the sample template (agent-handoff.md Step 1) or enable an out-of-the-box Handoff Accelerator, so real enabled topics do not carry the `Agenthandoff-scenarioname` name. Keying on the name would skip exactly the topics the maker enabled and misconfigured â€” the case #152 exists to catch.

**What it validates:** For each enabled handoff topic, resolve `Topic.HandoffAgentId` and confirm it is a concrete GPT id, not a shipped placeholder and not blank. Placeholder detection covers the whole family: the sample template ships `"AgentIdentifier"`; the OOTB accelerators (Workday / ServiceNow) ship system-prefixed variants such as `"ServiceNowAgentIdentifier"`. Match is a case-insensitive `...AgentIdentifier` suffix â€” real target ids are prefixed GUIDs (`T_<guid>`, `P_<guid>`) and never end that way. A placeholder or blank value makes the handoff fail at runtime.

**Remediation (two paths):** On failure the check offers both correct end states â€” set a valid target agent id (agent-handoff.md Step 4), OR disable the topic if the handoff was unintended (returns it to the OOTB default). Either resolves the failure.

**API tier:** validated

**Mock source:**

  - `tests/fixtures/cassettes/island_gateway_botcomponents.yaml` (interaction 2, the POST botcomponents response). The captured handoff `DialogComponent` ships `status="Inactive"` and `expressionText: "\"AgentIdentifier\""`. Read via `pva_client.get_dialog_components(bot_id)`, mirroring `get_knowledge_sources`.

  - Inference note: `status == "Active"` meaning "enabled" is inferred; the captured cassette only holds the disabled template. Enabled shape is consistent with other live components but unproven pending a captured enabled cassette. Validated live instead (below).

**Tests:** `tests/flightcheck/checks/test_agent_handoff.py` â€” gating (no client / not configured / no bot id / no handoff topic), identity by SetVariable (a handoff-named topic without the SetVariable is ignored; a renamed clone with the placeholder fails), GOOD (concrete id â†’ Passed), BAD (sample placeholder â†’ Failed; accelerator-prefixed `ServiceNowAgentIdentifier` â†’ Failed; blank â†’ Failed), and two-path remediation text. DialogComponent dicts are built inline from the cassette shape.

**Live E2E (EmployeeHub tenant):** ran the real check across the accessible ServiceNow environments. Of the handoff topics found, 4 were enabled: a blank-target accelerator â†’ Failed; an `AgentIdentifier` template â†’ Failed; an `AgentIdentifier` template on a second env â†’ Failed; a renamed accelerator with a concrete `T_<guid>` id â†’ Passed. 2 of the 4 enabled topics had schemaNames without `agenthandoff`, which the previous schemaName gate would have skipped. No regression.

Local runs: 15 targeted (`test_agent_handoff.py`), 1011 full flightcheck suite â€” all passing; ruff clean on both changed files.

Fixes #152

---

## Follow-up commits (review fixes + live validation)

**`cbe5a36` - MANUAL for reference/expression handoff ids (addresses review point #2):** `Topic.HandoffAgentId` can be assigned from a variable reference or a non-literal Power Fx expression, not only a typed string literal. The kit cannot statically resolve those, so reporting FAILED (placeholder/blank) for them is a false positive. `_resolved_agent_id` now returns `None` for non-literal values, and the check emits **MANUAL** (does not fail readiness) with observed-state + remediation text directing the operator to verify the resolved id in Copilot Studio. Added `TestReferenceOrExpressionValue` (3 tests: variableReference -> Manual, non-literal formula -> Manual, MANUAL sibling not swept into placeholder FAILED). Also added the missing `handoff` scope to the cli.py module docstring.

**`b87d28e` - authenticate PVA for `--scope handoff` (found during live validation):** `SCOPE_MAP['handoff']` wired the check, but `main()`'s Copilot Studio (PVA) auth gate only covered `full`/`local`/`graphconnector`. So `--scope handoff` left `runner.pva=None` and `run_handoff_topic_checks` returned `[]` without ever querying the tenant - 0 checks, a false 'ready'. The check only actually ran under `--scope full`. Extracted the gate into a `PVA_SCOPES` constant, added `handoff`, and added regression tests (end-to-end: the handoff scope instantiates `PVAClient`; plus a constant guard).

**Live validation (test-dawn sandbox, `org1bdc69c8`, HR agent):** Ran the real check end-to-end against a live tenant after the gate fix, manually enabling / editing / publishing the handoff topic in Copilot Studio between runs to exercise every TOPIC-020 branch.

| Handoff topic state | `Topic.HandoffAgentId` value | Result |
| --- | --- | --- |
| Disabled (`status=Inactive`) | `"AgentIdentifier"` | no row |
| Non-handoff topics (17) | n/a | no rows |
| Enabled, shipped placeholder | `="AgentIdentifier"` | **FAILED** (High) - "still holds shipped placeholder" |
| Enabled, blank | `=""` | **FAILED** (High) - distinct "is blank" message |
| Enabled, concrete id | `="<guid>"` | **PASSED** - overall READY, exit 0 |
| Enabled, Power Fx formula | `=Concatenate("agent","handoff")` | **MANUAL** - overall still READY (does not fail) |

All four TOPIC-020 branches (placeholder / blank / concrete / formula) plus both no-row cases are now confirmed against a live published tenant, not just unit tests. This closes the earlier "MANUAL covered by unit tests only" and "enabled shape unproven pending a captured enabled cassette" gaps.

Note on the formula case: a value typed as `=System.Activity.Text` is persisted by Copilot Studio as a **quoted** string literal (`"System.Activity.Text"`), so the resolver reads it as a concrete id -> PASSED (correct: a quoted literal is the literal runtime value). Only values that persist as **unquoted** formulas/references (`Concatenate(...)`, variable references, `expressionText` absent) resolve to MANUAL. Confirmed the resolver returns `None` for each unquoted shape using the exact live sibling `SetVariable` nodes on the topic.

**Test counts (updated):** `test_agent_handoff.py` 18 targeted; `test_cli.py` PVA-gate regression added; full flightcheck suite **1016 passing**; ruff clean on all changed files.


solutions/ess-maker-skills/scripts/flightcheck/checks/agent_handoff.py
```diff
@@ -0,0 +1,201 @@
...
+                    "contains the SetVariable node that sets "
+                    f"{_HANDOFF_VAR}. See agent-handoff.md Step 4."
+                ),
+                doc_link=DOC_LINK,
+            ))
+            continue
+
+        agent_id = _resolved_agent_id(setvariable)
+
+        if not agent_id or agent_id == _PLACEHOLDER:
```
daeunJe0ng: Nit (non-blocking, message accuracy): this branch merges two distinct states — a genuinely **blank** value (`not agent_id`) and the **placeholder** value (`agent_id == _PLACEHOLDER`) — but the `result` text always asserts the variable *"still holds the shipped placeholder 'AgentIdentifier'"*. For the blank case that statement is factually wrong (the value is empty, not the placeholder), which could confuse a maker who deliberately cleared it. The FAILED verdict itself is correct in both cases; only the explanatory string is off. Consider branching the `result` wording on `not agent_id` vs. `== _PLACEHOLDER` (remediation can stay shared).

daeunJe0ng (commented):
**Code review — TOPIC-020 handoff agent id check**

Verdict: **Approve with nits**

Reviewed the new check, the `get_dialog_components` client method, the CLI wiring, and the test module against the cardinal rules in `flightcheck/AGENTS.md` and `tests/AGENTS.md`.

What holds up well:
- `get_dialog_components(bot_id)` mirrors `get_knowledge_sources` exactly (same endpoint/headers/extraction) — reuses an already-registered API tier, no new grounding surface.
- Gating is correct and fails safe: returns `[]` when PVA is absent/unconfigured, bot id is missing, no enabled (`Active`) auto-handoff topic exists, or the client raises.
- `CheckResult` fields and the `Status` / `Priority` / `Role` enum values all match `runner.py`.
- Tests are cassette-grounded (validated tier, not the banned placeholder tier): every mock field traces to `island_gateway_botcomponents.yaml`, and only the two real captured values (`status`, `expressionText`) are varied. Gating / good / placeholder / blank / missing-SetVariable / multi-topic indexing are all covered.

One nit inline (non-blocking) on the FAILED-branch message. No blocking or should-fix bugs found.

daeunJe0ng:
Addressed the code-review nit in 8135710: the FAILED result text now branches on the failure mode. A blank `Topic.HandoffAgentId` is reported as blank; the placeholder case still names `AgentIdentifier`. Strengthened the blank-value test to assert the accurate wording. Targeted tests pass (12 passed).

apurvabanka:
Reviewed PR #218 — checked out the branch, read the check, the `pva_client` additions, the tests, the cassette fixture, and the CLI wiring; the 15 targeted tests pass locally. It's a solid, well-documented check. Wiring matches the existing `graph_connector_kb` pattern exactly, and identity-by-SetVariable (not schemaName) is the right call. Here are the gaps I'd flag, roughly by materiality:


**1. The `status == "Active"` gate has no automated guard (self-admitted).** Every "enabled" test manually flips the captured `Inactive`→`Active`; no cassette holds a real enabled topic. If Copilot Studio ever represents an enabled handoff topic differently (separate publish state, `status` absent on a draft component, etc.), `run_handoff_topic_checks` silently returns `[]` and #152 is never caught in production — a false negative with no test to catch the drift. The live E2E run mitigates this today, but nothing pins the behavior going forward. Worth a captured enabled cassette when one is available.


**2. Non-literal `HandoffAgentId` assignments → false-positive FAILED.** `_resolved_agent_id` only reads `expressionText`. The cassette itself shows sibling SetVariables using `variableReference` (e.g. `System.Activity.Text`) with no `expressionText`. If a maker sets the handoff id via a Power Fx variable/reference/formula instead of a string literal, `expressionText` is `""` → the check reports "blank (no target agent id set)" and **fails a correctly-configured topic**. No test covers a `variableReference`-style value. This is the most likely real-world false positive.


**3. `get_dialog_components` (43 new lines) has no automated test.** The endpoint + `botComponentChanges → component → $kind==DialogComponent` filtering is exercised only through the fake PVA in the check tests; no cassette-backed test asserts the parser reads the real fixture. It mirrors `get_knowledge_sources` (also untested), so no regression — but the new parsing is unverified.


**Minor:**



- **Placeholder match breadth:** `_is_placeholder_id` fails any id ending in `agentidentifier` (case-insensitive). Safe against prefixed-GUID ids, but technically over-broad. Low risk.

- **Duplicate POST:** `get_knowledge_sources` and `get_dialog_components` hit the identical botcomponents endpoint separately, so a full run makes two identical calls. Inefficiency, not correctness.

- **Doc gap:** the new `handoff` scope isn't added to the `cli.py` module docstring "Scopes:" list (argparse choices still work since they derive from `SCOPE_MAP`).

- **First-match only:** the recursive search returns the first `HandoffAgentId` SetVariable; conditional branches assigning it elsewhere aren't considered. Edge case.

CC: @daeunJe0ng

daeunJe0ng:
Thanks @apurvabanka, thorough review. Addressed the material items in cbe5a36:

**2 (non-literal `HandoffAgentId` -> false-positive FAILED): fixed.** `_resolved_agent_id` now returns `None` when the value is not a static quoted string literal (a `variableReference` or a Power Fx formula, per the sibling SetVariables in the cassette). Those topics now report **Manual**, not FAILED. Rationale: a reference is definitively not the shipped placeholder literal, but the kit cannot statically resolve the runtime value, so per `flightcheck/AGENTS.md` principle 1 it should not claim PASSED either. Manual does not fail readiness. New tests: `variableReference` -> Manual, non-literal formula -> Manual, and a MANUAL sibling not swept into a placeholder FAILED. The blank branch now fires only on the genuine shipped empty literal `""`.

**Doc gap (handoff scope missing from `cli.py` docstring): fixed.**

**1 (no captured enabled cassette): not fixed.** Cannot invent one under the cardinal rule; it needs a real capture of an enabled handoff topic. Left as the documented inference, validated live for now. Will capture when an enabled tenant is available.

**3 (`get_dialog_components` untested): deferred.** No regression (mirrors untested `get_knowledge_sources`, and no `pva_client` test harness exists yet). Worth a follow-up cassette-backed test against `island_gateway_botcomponents.yaml`.

**Minor (placeholder-match breadth, duplicate POST, first-match-only): acknowledged, not changing now** per your low/edge rating.

Tests: 18 targeted (`test_agent_handoff.py`), 1014 full flightcheck suite -- all passing; ruff clean on the changed files.

apurvabanka (approved):

PR #221, opened 2026-08-04
Automate ESS foundation setup

mukesh4139:
## Summary

This PR replaces the product-specific first-run `/setup` path with a reusable,
integration-neutral ESS foundation workflow.

The workflow:

- discovers, validates, and locks a Power Platform environment;
- verifies Dataverse, MCP, capacity, billing, and governance prerequisites;
- optionally configures a preferred unmanaged solution;
- installs or adopts supported ESS agents;
- verifies installation, binding, and baseline readiness;
- persists resumable progress and unlocks `/connect` only after setup completes.

Workday, ServiceNow, SAP SuccessFactors, integration authentication, extension
packs, and topic authoring remain outside foundation setup.

### Environment selection and validation

Makers can select a discovered environment, enter a known Dataverse URL, or
follow guidance to create a Dataverse-enabled environment.

Before locking the environment, setup verifies:

- **System Administrator** access, including team-inherited role assignments;
- environment identity and Dataverse provisioning;
- the documented Dataverse Allowed MCP Client entry for Microsoft GitHub
  Copilot.

The selected environment ID, name, type, endpoint, and verification timestamp
are persisted. Later steps reuse this locked context rather than repeatedly
listing or reconfirming environments.

### Prerequisites and ALM

Setup checks:

- Copilot Studio capacity and the selected billing model;
- DLP and outbound/firewall requirements;
- organization approval;
- Dataverse MCP enablement.

Preferred-solution configuration is optional. When selected, setup discovers
eligible customer-owned unmanaged solutions, sets the preferred solution only
when necessary, and verifies the final value before persisting its metadata.

### ESS product installation

The supported catalog now includes six independently installable products:

| Product ID | Experience | Product |
|---|---|---|
| `da.esshr` | Declarative agent | Employee Self-Service HR |
| `da.essit` | Declarative agent | Employee Self-Service IT |
| `da.esshub` | Declarative agent | Employee Self-Service Hub |
| `cea.esshr` | Custom engine agent | Employee Self-Service HR |
| `cea.essit` | Custom engine agent | Employee Self-Service IT |
| `cea.esshub` | Custom engine agent | Employee Self-Service Hub |

Live environment inventory separates installed agents from remaining
installation choices. Makers can install another available product or customize
an existing installed agent without reinstalling it.

Installation is catalog-driven and supports:

- automatic Power Platform installation with polling;
- manual-install fallback followed by verification;
- Microsoft 365 Self-Help connection discovery and selection;
- Dataverse connection-reference binding and reread verification;
- mandatory invoker-connection attestation where required;
- independent installation and readiness state for every selected product.

### Readiness and handoff

Each selected agent is checked for authoring access, reachable **Configure** and
**Topics** surfaces, and the expected starter footprint.

The final report is generated from persisted canonical state and includes the
locked environment, prerequisites, preferred solution, product status, and open
issues. `/connect` is enabled only when at least one selected product is
installed, bound, ready, and all required setup steps are complete.

## State and recovery model

`scripts/setup_state.py` provides a typed, versioned state model with these
canonical steps:

| Step | Purpose |
|---|---|
| `SETUP-01` | Lock the selected environment |
| `SETUP-02.1` | Verify environment access and Dataverse |
| `SETUP-02.2` | Verify MCP, capacity, billing, and governance |
| `SETUP-03` | Revalidate the locked environment |
| `SETUP-04` | Configure or explicitly skip the ALM baseline |
| `SETUP-05` | Install and bind selected ESS products |
| `SETUP-06` | Verify baseline readiness |
| `SETUP-07` | Finalize setup and hand off |

Each step stores only its status, concise result, mode, checkpoint where
applicable, and timestamp. Canonical environment, prerequisite, ALM, and product
facts are stored once rather than duplicated as validation evidence.

State mutations return concise deltas instead of printing the complete setup
document. Targeted views expose only the current step, environment, products,
report, or full state when explicitly requested.

Invalid, incompatible, or internally inconsistent state fails explicitly.
Finalization enforces locked-environment, selected-product, binding, readiness,
and step-result invariants.

## FlightCheck integration

Foundation setup reuses these checkpoints:

| Checkpoint | Usage |
|---|---|
| `ENV-001` | Validate the locked Power Platform environment |
| `ENV-002` | Verify Dataverse provisioning and access |
| `ENV-CAPACITY-001` | Check Copilot Studio capacity |
| `ENV-009` | Verify the preferred customization solution |
| `ESS-SOLN-001` | Verify ESS solution installation where required |

Single-checkpoint execution supports explicit environment context and quiet
authentication output. Exact-checkpoint runs now fail when they emit no target
row, and FlightCheck error results return a nonzero exit code.

Environment discovery emits reusable structured results so setup does not list
the same environments twice. A recently verified locked environment can be
reused within an explicit 15-minute validity window.

## Reliability and UX improvements

- Reuses locked environment and discovery results across setup transitions.
- Suppresses routine repeated authentication and discovery narration.
- Displays concise step results and the next required action instead of the full
  setup state.
- Refreshes cached Dataverse authentication once after HTTP 401.
- Uses user-scoped environment APIs for manual URL resolution.
- Runs capability telemetry without blocking the first setup prompt.
- Keeps DA and CEA products distinct and adds Hub support to both experiences.
- Removes legacy setup-state conversion and rejects unsupported schemas instead
  of silently migrating them.
- Removes the unrelated optional post-onboarding FlightCheck prompt.

## Validation

Focused regression coverage includes:

- setup state transitions, invariants, resume, and finalization;
- foundation and onboarding routing;
- environment discovery and role validation;
- Dataverse MCP configuration;
- preferred-solution selection and verification;
- all six ESS catalog products;
- installation, connection binding, and attestation;
- FlightCheck exact-checkpoint behavior and registry consistency;
- concise setup rendering and environment reuse.

Latest affected local suite:

```text
186 passed
```

## Scope boundaries

This PR intentionally does not perform:

- Workday, ServiceNow, or SAP SuccessFactors onboarding;
- external-system authentication;
- integration extension-pack installation;
- integration-specific connection configuration;
- topic creation or customization.

Those operations remain under `/connect` and `/create` after foundation setup.


solutions/ess-maker-skills/scripts/install_ess_agent.py
```diff
@@ -0,0 +1,723 @@
...
+    pp_admin.authenticate(include_flow=False)
+    environment_id = pp_admin.find_environment_id_by_dataverse_url(env_url)
+    if not environment_id:
+        raise RuntimeError(
+            "Could not resolve the selected environment's Power Platform ID."
+        )
+
+    validate_required_connection(
+        installation,
+        pp_admin.get_connections(environment_id),
```
saengland: **Blocking - reproducible crash on the permission-denied path.**

`PPAdminClient._get_all()` returns `{"_error": ..., "_status": 401|403}` on a permission failure rather than a list. Its own docstring states that *"every consumer of the list-returning methods (get_connections, get_flows, get_environments, get_dlp_policies) already guards with `isinstance(result, dict) and "_error" in result`."*

This call site does not guard, so `validate_required_connection()` iterates the dict and yields its string keys:

```
AttributeError: 'str' object has no attribute 'get'
```

Reproduced directly against this branch by passing `{"_error": "insufficient_permissions", "_status": 403}` into `validate_required_connection`. This fires whenever the signed-in maker lacks admin rights on the connections API, which is common rather than an edge case, and it is a gate the PR description calls mandatory for IT installs. The maker sees a stack trace instead of a permissions message.

`_list_packages()` at line 434 already does this correctly in this same file:

```python
packages = client.list_environment_application_packages(environment_id)
if isinstance(packages, dict) and packages.get("_error"):
    raise RuntimeError(
        "Your account cannot read Marketplace applications for this "
        "environment. Use a Power Platform or Dynamics 365 administrator "
        "account."
    )
```

Suggested fix here:

```python
connections = pp_admin.get_connections(environment_id)
if isinstance(connections, dict) and connections.get("_error"):
    raise RuntimeError(
        "Your account cannot read connections for this environment. "
        "Use a Power Platform administrator account, or ask an admin to "
        "confirm the required connection exists."
    )

validate_required_connection(installation, connections, connection_name)
```

Worth noting the current tests cannot catch this: `FakePPAdminClient.get_connections()` only ever returns a list, and the existing 403 test at `tests/scripts/test_install_ess_agent.py:414` covers `install_result` (the install POST), which is already guarded. A fake that returns the error dict would pin it.

solutions/ess-maker-skills/scripts/ess_connection_binding.py
```diff
@@ -0,0 +1,403 @@
...
+    client.authenticate(include_flow=False)
+    environment_id = derive_environment_id(env_url, "", client)
+    if not environment_id:
+        raise RuntimeError(
+            "Could not resolve the selected Dataverse URL to a Power Platform "
+            "environment ID."
+        )
+    return build_preflight_result(
+        installation,
+        client.get_connections(environment_id),
```
saengland: **Blocking - same defect as `install_ess_agent.py:561`, second instance.**

`client.get_connections(environment_id)` is passed straight into `build_preflight_result()` with no `_error` guard, so a 401/403 from the connections API produces the same crash:

```
AttributeError: 'str' object has no attribute 'get'
```

Reproduced on this branch by calling `build_preflight_result(installation, {"_error": "insufficient_permissions", "_status": 403}, "env-1")`.

Flagging separately because the bot review only caught the `install_ess_agent.py` instance - fixing that one alone leaves the connection-preflight path broken on the same permission failure.

Suggested fix:

```python
connections = client.get_connections(environment_id)
if isinstance(connections, dict) and connections.get("_error"):
    raise RuntimeError(
        "Your account cannot read connections for this environment. "
        "Use a Power Platform administrator account."
    )

return build_preflight_result(installation, connections, environment_id)
```

`tests/scripts/test_ess_connection_binding.py` has no `_error` coverage at all today, so this path is untested in both directions.

saengland (commented):
## [copilot-review] Automate ESS foundation setup

**Verdict: Request changes.** The design is sound and the state layer is genuinely
well built, but the branch is red (15 failed / 2 errors vs. 10 failed on main),
ships two reproducible `AttributeError` crashes on the permission-denied path,
regresses telemetry opt-out, and adds three unregistered `api.powerplatform.com`
endpoints (one of them mutating) without a tier-registry entry. None of this is
caught today because pytest does not run in CI.

**Why this PR exists (and is it the right change):** first-run `/setup` is
product-specific and not resumable, so an interrupted or partially-failed setup
leaves no durable evidence of what completed. Replacing it with an
integration-neutral foundation flow backed by a typed, resumable state document
is the right call, and `scripts/setup_state.py` is the right layer for it. The
objection is execution, not direction.

### Baseline used

| Branch | Result |
|---|---|
| `main` (cb9bfa2) | 10 failed, 1472 passed, 13 skipped |
| `pr/221` | **15 failed, 1572 passed, 13 skipped, 2 errors** |

The 10 installer failures are pre-existing on `main` and unrelated. The other
5 failures + 2 errors are new.

### Findings

**Blocking:**

1. **`get_connections()` error dict crashes two mandatory gates.**
   `_get_all()` returns `{"_error": ..., "_status": 401|403}` on a permission
   failure, and its own docstring states "every consumer ... already guards with
   `isinstance(result, dict) and "_error" in result`." This PR adds two consumers
   that do not:
   - `install_ess_agent.py:561` -> `validate_required_connection()`
   - `ess_connection_binding.py:153` -> `build_preflight_result()`

   Both reproduce identically:
   ```
   AttributeError: 'str' object has no attribute 'get'
   ```
   (iterating the dict yields its string keys). This fires whenever the maker
   lacks admin rights on the connections API, which is the common case, not an
   edge case. The maker sees a stack trace instead of "you need permission."
   `_list_packages()` at line 434 guards this correctly, so the fix is to apply
   the same pattern in both places.

   Note the test fake cannot catch this: `FakePPAdminClient.get_connections()`
   only ever returns a list. The existing 403 test covers `install_result`
   (the install POST), which is already guarded.

2. **Telemetry now fires for makers who opted out.** `emit_capability.py` moved
   from an in-process `emit_capability_use(...)` call to an unconditional
   `subprocess.Popen(...)`, with the consent check relocated *inside* the
   `--worker` branch. The parent no longer checks consent before spawning, so a
   maker who ran `python scripts/adk_telemetry.py off` still gets a Python
   process launched on every skill step. The module docstring still claims
   "Consent-aware ... so no extra gating is needed," which was only true while
   the emit was in-process. Gate the `Popen` on the same disabled check.

3. **Unreaped subprocess breaks the suite.** The `Popen` handle is discarded and
   never `wait()`ed, so `Popen.__del__` raises
   `ResourceWarning: subprocess N is still running`. `pyproject.toml` sets
   `filterwarnings = ["error"]`, turning that into 4 failures + 2 teardown errors
   in `tests/test_adk_telemetry.py`.

4. **Registry contract changed without updating its test.**
   `requires_config=True -> False` on `ENV-001`, `ENV-002`, `ENV-CAPACITY-001`.
   The intent is clear (run from `--environment-url` without `.local/config.json`)
   but `test_registry.py::test_env002_pulls_env001_prereq` pins the old contract
   and now fails. Separately: those specs keep `requires_dataverse_endpoint=True`
   while dropping `requires_config`. If config was the source of the endpoint,
   does a bare `--checkpoint ENV-002` still resolve one, or fail later and less
   clearly?

5. **Three unregistered API surfaces, one mutating.**
   `flightcheck/powerplatform_client.py` gains:
   ```
   GET  /environmentmanagement/environments
   GET  /appmanagement/environments/{id}/applicationPackages
   POST /appmanagement/environments/{id}/applicationPackages/{name}/install
   ```
   None are in `tests/fixtures/cassettes/INDEX.md`. The only registered
   `api.powerplatform.com` entry is the Licensing/Billing surface
   (`/licensing/billingPolicies...`, tier `documented`), a different path family.
   `INDEX.md` and `tests/mocks/` are both untouched, so there is no mock builder
   and no documented response contract. The registry rule is unconditional:
   "If you need to call an API that isn't in this registry, STOP ... the tier
   must be decided (and recorded here) before any check or test can use the API."

   The POST matters most: it writes to the customer tenant. `INDEX.md` currently
   describes INFRA-003 as "the only FlightCheck path that writes to the tenant,"
   with disposable-environment capture, a deterministic probe name, an orphan
   sweep, and SAS redaction. This adds a second tenant-write path through the
   same client family with none of those safeguards.

   Acknowledged nuance: these are called from setup scripts rather than
   `checks/*.py`, so the cardinal rule does not bind word-for-word. They still
   live in `scripts/flightcheck/` on a runner-instantiated client, and the
   registry sentence has no carve-out.

**Should-fix:**

6. **`clear_token_cache()` deletes the shared MSAL cache.** It `os.remove()`s
   `.local/.token_cache.bin`, which is shared by six clients (`auth.py`,
   `graph_client`, `pp_admin_client`, `powerplatform_client`, `azure_arm_client`,
   `pva_client`). A single Dataverse 401 from either call site (`auth.py:222`,
   `preferred_solution.py:277`) forces fresh interactive sign-in across Graph,
   BAP, Licensing, ARM, and Island Gateway. A 401 here often means the user has
   no access to that environment, so re-auth cannot fix it and the user lands on
   the same 401. Prefer MSAL `remove_account()` scoped to the Dataverse account.

7. **Structural state corruption leaks `TypeError`.** The PR says corrupt state
   "is surfaced as an error." `SetupWorkflow.validate()` handles semantic
   corruption well, but `from_dict` builds dataclasses via `**record` before
   validating, and `main()` only catches `SetupStateError`:

   | Corruption | Result |
   |---|---|
   | step record is a string | leaks `TypeError` |
   | step record has unknown key | leaks `TypeError` |
   | validations record is a list | leaks `TypeError` |
   | products record is a string | leaks `TypeError` |
   | products dropped / unknown product id / bad `selected_products` | clean `SetupStateError` |

   Through the real CLI:
   ```
   File "setup_state.py", line 213, in from_dict
       step_id: StepRecord(**record)
   TypeError: __main__.StepRecord() argument after ** must be a mapping, not str
   ```
   Wrapping the three comprehensions and re-raising as `SetupStateError` closes
   all four.

8. **Two open bot findings unaddressed** (filed 8/4):
   - `install_ess_agent.py:563` - the `get_connections` issue above (finding 1).
   - `list_environments.py:118` - `resolve_environment_for_user()` calls
     `discover_tenant(env_url)` outside its `try`, so an invalid URL produces a
     stack trace instead of the script's `ERROR:` + exit 1.

9. **23 new bare `except Exception` blocks** across the setup path. Large silent
   failure surface for a flow whose stated value is being "evidence-driven,
   fail-closed."

10. **Merge order vs. #224.** `git merge-tree` reports 4 files changed in both.
    Concretely this PR edits `onboarding/tasks.md` while #224 renames that exact
    file to `steps.md`. #224 is 60 lines and mechanical; this is 7,521. Land #224
    first.

**Nit / optional:**

11. `check_environment_roles.py` catches `(APIError, OSError, RuntimeError,
    ValueError)` while its siblings catch `(OSError, RuntimeError, ValueError)`.
    `APIError` subclasses `OSError` via `requests.RequestException`, so the extra
    entry is redundant but harmless. Worth aligning for consistency only.

### Product question (not an engineering defect)

`check_environment_roles.py` requires **both** Environment Maker **and**
Dataverse System Administrator, and `scope.md` makes it terminal: "When
`eligible` is false, do not lock the environment ... If the command fails, show
the exact error and stop." There is no override or attestation path.

Two consequences worth a deliberate decision. It raises the entry bar for a kit
the README positions as "no deep platform knowledge required," since most
enterprise makers hold Environment Maker but not System Administrator. And in
JIT-elevation tenants, reported role state and effective capability diverge, so a
maker who legitimately holds the role can still be blocked. Flagging rather than
prescribing.

### What holds up well

- `setup_state.py` persistence is correct: `tempfile.mkstemp` + `fsync` +
  `os.replace`, validating on both load and save. Better than most of this repo.
- `SetupWorkflow.validate()` catches every semantic corruption case I probed.
- No token leakage anywhere in the diff.
- `pp_admin.authenticate(include_flow=True)` default preserved and covered by a
  new test, so the second-token optimization is backwards-compatible.
- `_wait_for_install()` uses injectable `sleep`/`clock`, a real timeout, and a
  dedicated `InstallationTimeoutError`.
- `check_dataverse_mcp.py` escapes the OData filter correctly.
- ~2,600 lines of new tests (333 passing in `tests/setup` + `tests/scripts`).

### Meta

Worth flagging for the repo, not this PR: **pytest never runs in CI.** `ci.yml`
runs ruff, `compileall`, and a PowerShell installer harness; `validate-samples`
and `mtk-toolkit-ci` are both path-filtered away from `solutions/`. So 0 of ~1,482
tests execute on a maker-kit PR, which is why `main` has been sitting at 10
failures and why this branch's 5 new ones were invisible. #209 fixes the largest
part of this and is the highest-leverage PR in the queue right now. Separately,
`pip install -e .[test]` fails on `main` today ("Multiple top-level packages
discovered in a flat-layout"); #209 fixes that too.

solutions/ess-maker-skills/src/reference/solution-catalog.md
```diff
@@ -0,0 +1,106 @@
...
+|  |  |  | Name: `ServiceNow`<br>Logical name: `msdyn_copilotforemployeeselfserviceit.c3d5c4e2-5f11-4a7e-9f3b-9c57e1d78265.shared_service-now` | title: `ESS IT ServiceNow HRSD Common Orchestrator`<br>workflowid: `c3d5c4e2-5f11-4a7e-9f3b-9c57e1d78265`<br>runtime: `invoker` |
+|  |  |  | Name: `Microsoft Dataverse`<br>Logical name: `new_sharedcommondataserviceforapps_41c83` | title: `ESS IT ServiceNow HRSD Get HR Services With COEs for User`<br>workflowid: `7d92a3f4-1b4f-42e9-9d52-6a1c3eae0b98`<br>runtime: `embedded` |
+| `msdyn_CopilotForEmployeeSelfServiceIT` | `ServiceNow IT` | `msdyn_EssITServiceNowITSM` | Name: `ServiceNow`<br>Logical name: `msdyn_copilotforemployeeselfserviceit.cr.w2LCWZTZ` | title: `    ESS IT ServiceNow ITSM Common Orchestrator`<br>workflowid: `3f7a2b1c-8d6e-4a5b-9c0d-2e3f4a5b6c7d`<br>runtime: `invoker` |
+|  |  |  | Name: `Microsoft Dataverse`<br>Logical name: `new_sharedcommondataserviceforapps_41c83` | title: `    ESS IT ServiceNow ITSM Common Orchestrator`<br>workflowid: `3f7a2b1c-8d6e-4a5b-9c0d-2e3f4a5b6c7d`<br>runtime: `embedded` |
```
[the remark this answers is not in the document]
mukesh4139: that's intended for better formatting

mukesh4139:
Addressed the Python findings from this review in bfb5f93:

- Added explicit 401/403 error-dict handling for connection discovery in both installation and binding paths, while skipping the connections API entirely for HR/Hub products that do not require one.
- Restored telemetry opt-out before worker launch and added worker reaping; synchronous test mode no longer creates detached processes.
- Replaced shared-cache deletion with MSAL application-level account removal so the rejected account's access, refresh, and ID tokens are evicted without deleting the whole cache file.
- Normalized malformed step/validation/product state into `SetupStateError` and handled invalid environment URLs through the normal `ERROR:` path.
- Updated the ENV checkpoint registry contract test for explicit `--environment-url` execution.
- Added DA and CEA Employee Self-Service Hub installation choices end-to-end.

Validation now reports 1,589 passed and the same 10 pre-existing installer failures documented for `main`; no new PR-specific failures remain. The API tier-registry/documentation item and the product/merge-order questions are separate and are not claimed as resolved by this commit.

solutions/ess-maker-skills/scripts/list_environments.py
```diff
@@ -94,6 +97,61 @@ def get_dataverse_environments():
...
+            urlparse(instance_url.rstrip("/")).hostname or ""
+        ).casefold()
+        if instance_host == target_host:
+            return environment
+    return None
+
+
+def resolve_environment_for_user(env_url):
+    """Resolve one Dataverse URL through the user-scoped Power Platform API."""
+    client = PowerPlatformClient(discover_tenant(env_url))
```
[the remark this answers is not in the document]
mukesh4139: Fixed in bfb5f93. Tenant discovery now runs inside the guarded path, and invalid/non-HTTPS URLs produce the script's normal `ERROR:` output and exit code 1 instead of a traceback. Added regression coverage.

solutions/ess-maker-skills/scripts/install_ess_agent.py
```diff
@@ -0,0 +1,723 @@
...
+        connection_name,
+    )
```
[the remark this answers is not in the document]
mukesh4139: Fixed in bfb5f93. The installer now recognizes the structured `_error` response and returns a clear permissions error. It only calls `get_connections()` for products that actually declare a required connection, so HR and Hub installs are not blocked unnecessarily. Added 403 and no-connection regression tests.

solutions/ess-maker-skills/scripts/install_ess_agent.py
mukesh4139: Fixed in bfb5f93 with the suggested structured-error guard and regression coverage. The final implementation also avoids calling the connections API for HR/Hub products whose catalog entry has no required connection.

solutions/ess-maker-skills/scripts/ess_connection_binding.py
mukesh4139: Fixed in bfb5f93. Connection preflight now converts `_error` responses into a clear permissions `RuntimeError` before building the result, with dedicated regression coverage.

mukesh4139:
Comprehensive response to the review (current head `bfb5f93`):

1. **Connection error dictionaries — resolved.** Both installer and binding preflight now convert `_error` responses into clear permission errors. The installer only calls the connections API for products that declare a required connection, so HR/Hub are not blocked by connection-list permissions. Regression tests cover both paths.
2. **Telemetry opt-out — resolved.** Consent is checked in the parent before any worker is launched.
3. **Unreaped subprocess — resolved.** Detached workers are reaped, and synchronous test mode emits without creating a subprocess.
4. **Registry contract — resolved.** The test now expects `requires_config=False` and `requires_dataverse_endpoint=True`. `--checkpoint ENV-002 --environment-url <url>` runs without `.local/config.json`; a checkpoint invocation with neither config nor `--environment-url` fails immediately with the explicit missing-endpoint error rather than later in client initialization.
5. **Unregistered Power Platform API surfaces — still pending.** Agreed. The environment-list, package-list, and package-install routes are not yet registered in `tests/fixtures/cassettes/INDEX.md`, and the mutating install POST still needs its response contract/mock and tenant-write safeguards documented. This remains the outstanding engineering blocker from this review and is not claimed as fixed by `bfb5f93`.
6. **Shared MSAL cache deletion — resolved.** We now use `PublicClientApplication.remove_account()` for the rejected account and persist the remaining shared cache, rather than deleting `.local/.token_cache.bin`.
7. **Structural state corruption — resolved.** Malformed step, validation, and product records/containers are normalized to `SetupStateError`; regression coverage includes strings, unknown keys, lists, and invalid containers.
8. **Open bot findings — resolved.** Both the connection error path and invalid environment URL path have fixes and inline replies.
9. **Broad exception handling — rechecked.** On the current PR diff, the only newly added bare `except Exception` handlers are two boundaries in `emit_capability.py`; both are intentionally fail-open because telemetry must never break a maker flow. The user-facing setup/auth/environment paths use specific exception types. We are not treating errors as successful setup evidence.
10. **Merge order with #224 — pending sequencing.** Agreed with landing the smaller mechanical rename first. #224 is still open and mergeable; #221 should be rebased after it lands.
11. **Redundant `APIError` catch — optional and still present.** Confirmed harmless because `APIError` is already covered by `OSError`; this can be removed for consistency but is not behaviorally significant.

**Product decision — resolved.** The complete automated `/setup` path requires Environment Maker plus effective Dataverse System Administrator access in the selected environment because the mandatory MCP prerequisite directly reads the Allowed MCP Clients table. There is no manual attestation or bypass: a `403` remains blocking. The role assignment must be effective for the same signed-in account and environment; if the read is denied, verify account identity, environment-specific assignment/JIT activation, and the effective organization-level read privilege rather than merely reassigning the same role.

**CI meta item — tracked separately.** #209 remains open and mergeable. It is the appropriate place to add pytest execution/package-install fixes; it is not folded into #221.

Current validation is 1,589 passed with the same 10 pre-existing installer failures reported on `main`; the 5 failures + 2 errors introduced by the earlier PR head are gone.

mukesh4139:
Follow-up `2bef17d` resolves the two remaining items discussed above:

- **Power Platform API registry/contracts:** registered Environment Management and App Management as `documented` tier using the Microsoft Learn 2024-10-01 references; added grounded mock builders and direct client contract tests. The durable install POST now has explicit idempotency/write-safety documentation: package state is read first, installed/in-progress packages are never re-posted, the selected catalog product is persisted, and POST is excluded from automatic retries. Installation completion now polls the documented application-package collection, so the previously used undocumented operation-status endpoint has been removed. Environment pagination also handles the documented lowercase `@odata.nextlink` field.
- **Redundant `APIError` catch:** removed the redundant import/tuple entry from `check_environment_roles.py`. A regression test confirms `APIError`/`AuthExpiredError` remains caught through its `OSError` base class and returns the controlled `ERROR:` path.

Current validation: 1,596 passed, with only the same 10 pre-existing installer failures on `main`.

mukesh4139:
Product decision for the System Administrator question in [this review](https://github.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/pull/221#pullrequestreview-4866898543):

- **Yes, the current end-to-end `/setup` flow requires Environment Maker plus effective Dataverse System Administrator access in the selected environment.** System Administrator is needed because the mandatory MCP prerequisite performs a direct Dataverse read of the **Allowed MCP Clients** table.
- We will not add a manual attestation or bypass for that check. A `403` remains a blocking result because `/setup` cannot prove that Microsoft GitHub Copilot is allowed to use the Dataverse MCP server.
- The role name alone is not sufficient evidence. The assignment must be effective for the same signed-in account and selected environment. If the read still returns `403`, the handoff should verify the authenticated account, environment-specific role assignment/JIT activation, and the effective organization-level read privilege for Allowed MCP Clients rather than simply reassigning the same role.

This resolves the product-policy question: **Environment Maker alone is not sufficient for the complete automated setup path while Allowed MCP Client verification remains mandatory.**

apurvabanka:
## PR #221 review — "Automate ESS foundation setup"


**Verdict:** Solid, well-tested PR. All 358 new tests pass; the state-machine invariants, config drift protection, atomic persistence, and error paths are careful. No high-confidence functional bugs, security issues, or broken invariants. The gaps below are minor/provenance-level plus a few things worth confirming.


### Gaps worth addressing


**1. SETUP-06 readiness is always recorded as `--mode automated`** (`foundation-setup/readiness.md`)

The doc itself says to "require explicit **manual attestation** for each starter" (open agent, confirm Configure/Topics reachable), yet the persisted step-result hardcodes `--mode automated`. SETUP-05 correctly branches (`manual-attested` when invoker attestation was needed, else `automated`); SETUP-06 doesn't. It's cosmetic — no functional gate — but it mislabels provenance in the persisted state and the SETUP-07 handoff report. Fix: record `manual-attested` (or branch like SETUP-05).


**2. Hardcoded connection-reference GUID in `config.json`** (`da.it`, `cea.it`)

`referenceLogicalName` pins the suffix `…shared-alchemy-8262076a-…`, and `bind_connection` filters `connectionreferences` by *exact* logical name, hard-failing ("found 0") if it differs. This is packaged-stable within the managed solution, so likely fine — but it's a brittle literal that will silently break IT binding if the marketplace package ever re-versions that reference. Worth a comment noting it's package-pinned.


**3. `preferred_solution.list_candidates` is fail-closed and N+1**

One `get_publisher` call per solution, and a `RuntimeError` if *any single* unmanaged solution has incomplete publisher metadata blocks the entire picker from listing. Consider skipping/annotating bad entries instead of aborting.


**4. Dead `manual-required` branch in `install_ess_agent.main()`**

The callback `mapped_status` dict includes `"manual-required"`, but `install_agent` never emits it via the callback (the timeout path persists it directly). Harmless, but signals the manual-fallback path isn't driven where you'd expect.

CC: @mukesh4139

mukesh4139:
Addressed items **1** and **4** in `be95063`:

1. **SETUP-06 provenance:** readiness now records `manual-attested` when any selected starter required manual readiness attestation; fully automated verification remains `automated`. Added a structural regression test.
2. **Dead `manual-required` callback branch:** removed the unreachable mapping from `install_ess_agent.main()`. The timeout handler remains the single path that persists `manual-required` with the installation schema name.

Items 2 and 3 are intentionally unchanged.

apurvabanka (dismissed):

apurvabanka (dismissed):

apurvabanka (approved):

PR #225, opened 2026-08-05
flightcheck: keep GUIDs inside deep-link URLs unmasked in MANUAL reports

daeunJe0ng:
## Summary

`_mask_sensitive` (in `runner.py`) redacts operator-identifying values (email/UPN local parts, resource GUIDs) from `CheckResult.remediation` before a report is rendered, because readiness reports are often shared beyond the operator. It runs on the **MANUAL** render path (`_render_manual_checklist` -> `_mask_sensitive` -> `_md_links_to_html`).

Problem: it masked GUIDs that are **path segments of a deep link**. A MANUAL remediation deep link such as Copilot Studio `.../environments/<env_id>/bots/<bot_id>/overview` had both GUIDs rewritten to `1a2b3c4d-****-****-****-************`, so the resulting `href` no longer resolved. Every MANUAL check with a deep link was affected. Concretely, the publishing QA-* checks (`checks/publishing.py`) emit their Copilot Studio Analytics deep link as MANUAL, so those links were masked into broken hrefs.

## What changed

- `_mask_sensitive` now splits the text into `http(s)://` URL spans and prose, and masks **only the prose**. URL spans are kept verbatim, so a GUID that is a path segment of a deep link survives and the link resolves.
- Standalone GUIDs and emails/UPNs in prose are still masked exactly as before.
- The URL regex stops before a `)` so a markdown target `[label](url)` ends at its wrapping paren.

## Design rationale

The GUIDs that broke are `env_id` / `bot_id`, which the URL needs to be clickable. They already appear in **every** deep link the report emits (including FAILED/WARNING cards, which are never masked), so keeping them inside the URL leaks nothing the report doesn't already show. Masking them only broke the link without adding privacy. The fix is scoped to the single choke point (`_mask_sensitive`) so every MANUAL remediation inherits it, and it preserves the existing prose-masking behavior.

## Testing

- `tests/flightcheck/test_runner_report_layout.py`: added 3 tests - GUID inside a bare deep-link URL is preserved while a standalone prose GUID in the same string is still masked; GUIDs inside a markdown link target are preserved; end-to-end MANUAL check renders a deep-link `href` with real GUIDs while a UPN in the same remediation is still masked. Existing `test_mask_sensitive_redacts_upn_and_guid` and `test_manual_checklist_masks_sensitive_data` still pass.
- Full FlightCheck suite (`tests/flightcheck`): 999 passed.

## Scope

- Fix targets **URLs (links)** only. A GUID inside a non-URL context (e.g. a PowerShell `-EnvironmentName <guid>` snippet in a remediation) is still masked; that is a separate concern from clickable links and is left untouched here.
- Independent of #208 (bare-URL autolinking) and #209 (URL host hygiene gate). This bug pre-exists both on `main`: the affected publishing links are markdown links, already masked before #208.

## Risks and assumptions

- Change touches shared report rendering used by every MANUAL check. Mitigated by the full-suite run and by keeping prose masking identical.
- Assumes URL runs are delimited by whitespace or a wrapping `)`. A balanced-paren path segment after a `)` would fall outside the URL span, but GUIDs contain no parens, so masking behavior for GUIDs is unaffected.

### DIFF
--- a/solutions/ess-maker-skills/scripts/flightcheck/runner.py
+++ b/solutions/ess-maker-skills/scripts/flightcheck/runner.py
@@ -1510,20 +1510,42 @@ def _mask_sensitive(text: str) -> str:
     operator, so we mask the local part of addresses and the middle of
     GUIDs while keeping enough context (domain, first block) to stay
     actionable.
+
+    Text inside an ``http(s)://`` URL is left verbatim. A remediation deep
+    link (e.g. Copilot Studio ``.../environments/<env_id>/bots/<bot_id>``)
+    carries GUIDs as path segments; masking them rewrites the ``href`` and
+    the link stops resolving. Those ids already appear in every deep link
+    the report emits, so keeping them in the URL leaks nothing new while
+    keeping the link clickable. Standalone GUIDs / emails in prose are
+    still masked.
     """
     import re
-    # Email / UPN -> keep first char + full domain: j***@contoso.com
-    text = re.sub(
-        r'([A-Za-z0-9])[A-Za-z0-9._%+-]*(@[A-Za-z0-9.-]+\.[A-Za-z]{2,})',
-        r'\1***\2', text,
+    email_re = re.compile(
+        r'([A-Za-z0-9])[A-Za-z0-9._%+-]*(@[A-Za-z0-9.-]+\.[A-Za-z]{2,})'
     )
-    # GUID -> keep first block: 1a2b3c4d-****-****-****-************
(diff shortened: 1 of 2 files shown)


daeunJe0ng:
Folding this into #208 instead. The masking fix directly addresses review note 1 on that PR, so it makes sense to keep it in the same review rather than a separate one. The commit (keep GUIDs inside deep-link URLs unmasked) is now on the #208 branch.

PR #226, opened 2026-08-05
Add /test: self-serve runtime validation & debugging for topics and workflows

nkemms:
## Add `/test`: self-serve runtime validation & debugging for topics and workflows

Turns the `topics/test` stub into a working **drive-and-diagnose** capability behind a new **`/test`** command. A maker (or the agent) can drive a just-authored topic or workflow, confirm the reply is real, check it against the behaviour its evaluation cases intend, and localize a fault — without leaving the kit.

This fills the **L4 (self-serve validation & debugging)** gap. `/test` is component debug, not the graded evaluation gate (L5, a separate workstream).

### What's added

- **`/test` command** routing topic → `topics/test`, workflow → `workflows/test`. `/create` and `/update` auto-continue into it, scoped to the just-authored component.
- **The debug-and-validate loop** (`topics/test`): build a failure-first probe set → drive & capture → confirm real → check against intent → localize → re-drive.
- **Diagnostic core** (browser-free, unit-tested): `reply_signal.py` (classify a reply), `flow_run_inspect.py` (read-only per-action run history; surfaces the scope-vs-handler 500-masks-400 trap), `plant_debug.py` / `strip_debug.py` (byte-reversible DBG node), `topic_checker_capture.py` (surface the Topic checker for publish-time card/PowerFx defects).
- **Drive surface** (`drive_topic.py`, `cdp_driver.py`): launch/attach an InPrivate Edge on the test pane and drive turns, scoped to the requested agent.

### Key model

> [!NOTE]
> **Never mutate the eval cases.** They are the read-only intent guardrail: the loop confirms a fix doesn't break the intent the cases encode; it never edits them and never grades the topic.

Diagnosis is read-only; the DBG node is the one temporary, reversible mutation. `topics/test` covers a flow through the conversation; `workflows/test` covers it flow-first — they cross-link.

### Testing

- **146 unit tests** over the diagnostic core and drive surface (reply signals + precedence, flow-cascade summary, Topic-checker exit-code contract, DBG round-trips, page selection). ~5s.
- Offline structural validation (reference integrity, doc-vs-CLI flag audit, a no-env dry-run against the sample HR agent). The drive/capture path is live-verified against a real ESS agent.


amilandi (changes requested):
## Blocking: `flow_run_inspect._auth_headers` never sends the bearer token

**File:** `solutions/ess-maker-skills/scripts/flow_run_inspect.py:120-121`
**Severity:** Critical

`_auth_headers` returns a literal string with six asterisks instead of the bearer token — the `token` parameter is accepted but never interpolated:

```python
def _auth_headers(token: str) -> dict[str, str]:
    return {"Authorization": f"******"}
```

Every Flow Management API call in the module (`list_environments`, `resolve_environment_id`, `get_latest_run`, `get_run_by_id`, `get_run_actions`) routes through this and will be sent unauthenticated. `api.flow.microsoft.com` will 401 every request, so the "why did the flow fail" surface this PR advertises is non-functional in real use.

The existing tests only exercise the pure `summarize_actions` / `_extract_status_code` helpers — there's zero coverage of the HTTP layer, which is why CI didn't catch it.

**Fix:**
```python
def _auth_headers(token: str) -> dict[str, str]:
    return {"Authorization": f"Bearer {token}"}
```
Please also add at least one test that captures the outgoing `Authorization` header (e.g. monkeypatch `requests.get`) so this can't regress silently.

---

Otherwise the PR looked clean on the areas I reviewed carefully:

- `auth.py` refactor preserves the 0o600 cache file mode and the existing public-client/consent pattern; scopes unchanged.
- `cdp_driver.py` uses list-form Edge args (no shell), only talks to localhost, doesn't interpolate user input into evaluated JS, and `close()` releases the Playwright handle on error paths.
- `plant_debug` OData filter escapes single quotes, refuses to PATCH without an anchor, and rolls back the live topic if provenance write fails.
- `flow_run_inspect` validates env/flow/run IDs before URL interpolation, refuses non-HTTPS, and avoids stringifying SAS-fetch exceptions (no SAS token leakage in logs).
- `strip_debug` / `debug_plant` are idempotent on re-strip.

No other high-confidence issues found. Once the header fix and a covering test land, LGTM.

nkemms:
Thanks for the careful review. On the blocking item: the line on the branch is already the fix you suggested — `_auth_headers` returns `f"Bearer {token}"`, not a literal. The `******` in the "current" block is a secret-masking artifact from the file-reading tool (the `Bearer {token}` pattern gets masked on read); the raw bytes decode to `Bearer {token}`, and `git log -L` shows the function body was added once and never changed. So the API calls are authenticated in real use.

The second half was a real gap, though — there was zero HTTP-layer coverage. Added in cd37334:

- `test_auth_headers_interpolates_bearer_token` — asserts `Authorization == "Bearer " + token` (compared against a concatenated value, so it fails if the token is ever dropped or a template literal returned).
- `test_outgoing_request_carries_bearer_token` — monkeypatches `requests.get` and asserts the outgoing request from `get_latest_run` carries the bearer token.

20/20 pass, ruff clean. Could you re-review?

amilandi (approved):
You're absolutely right — apologies for the noise. I re-fetched the branch and base64-dumped the raw bytes of `_auth_headers`, and the source is `f"Bearer {token}"` as designed. What I read in my review was my tooling automatically masking the `Bearer` prefix on read (a known credential-header pattern gets redacted to `******` in file-view output). My `git show` pass hit the same masking, so I saw `f"******"` in both places and mistakenly took that as the truth of the file. The API calls have always been authenticated.

The two new tests are exactly the coverage that gap needed:

- `test_auth_headers_interpolates_bearer_token` asserts against `"Bearer " + token`, so a template-literal or dropped-token regression fails the assertion.
- `test_outgoing_request_carries_bearer_token` monkeypatches `requests.get` and checks the outgoing header from `get_latest_run`, which pins the actual HTTP request path.

Approving. Sorry again for the false alarm.

srideshpande (approved):
LGTM.

PR #227, opened 2026-08-05
flightcheck: apply Fluent 2 report redesign

daeunJe0ng:
## Summary

This applies a Microsoft Fluent 2 redesign to the FlightCheck HTML readiness report. The change is a re-skin plus markup refinements to match the approved design mockup. It touches only the report renderer and its layout tests; there are no changes to check logic, the data model, or the CLI.

## What changed

- Verdict header is now flat: a single brand title, `Label: value` meta rows, a state-specific SVG glyph inline with the headline, and the sub-line as a sibling `verdict-note` span.
- Readiness-at-a-glance synopsis drops the per-tile colored dot and the rollup summary line, keeping the per-status totals row.
- Action items now show the checkpoint ID at the top of each item, the title is no longer the link, and a separate `Details` link sits in the meta row.
- Warnings header reads "Review N warnings" and each warning lists its bold checkpoint ID followed by a short overview, linking to the full card.
- The two Expand-all and Collapse-all buttons are replaced with one self-relabeling toggle that stays in sync with section and filter state.
- The "How to read this report" guide gains guidance on the manual section steps and on value masking, and the stale "Fold all / Unfold all" wording is corrected.
- Removed four now-dead CSS selectors left behind by the markup changes.

## Design rationale

All rendering lives in `runner.py` as Python string builders plus one `_REPORT_CSS` constant, so the redesign is a swap of CSS tokens and small edits to the `_render_*` functions. Keeping the change inside the renderer avoids any risk to the check catalog or run results. The verdict header was flattened to match the mockup structure exactly, which required updating the verdict layout tests to anchor on the new sibling `verdict-note` span.

## Testing

- Full `tests/flightcheck` suite passes: 864 passed.
- `runner.py` compiles clean with `py_compile`.
- Rendered a representative sample report and verified the header, synopsis, action items, warnings, fold toggle, and how-to sections against the mockup.

## Risks and assumptions

- Multiline check results render as prose `dd.detail` with `<br>` rather than the mockup's `pre.err` evidence block, because the data model exposes a single `result` field. This is a deliberate divergence.
- `_render_header` still accepts a `verdict_icon` argument that is now unused; it mirrors the test-pinned `_verdict_text` 4-tuple and was left in place to avoid unnecessary churn.
- The mockup's Environment and Integration model meta rows and its "Grounding and honesty" paragraph are intentionally not included. Those values are illustrative in the mockup and have no backing data in `RunResult`; adding them would state unverified facts in a real report.

## Review fixes (commit d5a9159)

Addresses both issues from @amilandi's review.

- Issue A (manual checklist was dead markup): restored a delegated `change` handler in `_REPORT_SCRIPT` on `.cl-item input[type=checkbox]`. Ticking a box now recomputes `.cl-count` (turns green at 100%), toggles `.cl-item.done` strikethrough, and fills `.cl-bar > i` width, scoped per containing `.checklist`.
- Issue B (masking guidance was false): routed non-manual `res.result` (Detail) and `res.remediation` (Next step) through `_mask_sensitive` in `_render_check_card`, so UPNs and GUIDs are masked on every card, not just manual checklists. Corrected the how-to copy to state accurately what masking does (user emails and resource GUIDs redacted in prose; IDs inside links kept so links resolve) instead of overclaiming secrets and tenant domains.

Added two regression tests in `test_runner_report_layout.py`: one asserts the checklist `change` handler is wired, one asserts a non-manual (Warning) check masks its Detail and Next-step. Full `tests/flightcheck` suite passes: 1036 passed. `runner.py` compiles clean.


amilandi:
Reviewed the current head (`7eecca7`). Both issues from my earlier pass are still present.

### Issue A — Manual checklist counter / progress bar / strikethrough is broken (medium)

`runner.py` `_REPORT_SCRIPT` no longer contains the IIFE that listened for `change` on `.cl-item input` and updated `.cl-count`, `.cl-item.done`, and `.cl-bar > i` width. The CSS (lines ~699–718) and the HTML markup (lines ~1804–1820) are still emitted, but nothing wires them. Ticking a manual-check box currently does nothing visible:

- The `0 / N` counter never increments.
- The progress bar stays empty.
- Items don't get strikethrough / muted styling.
- The "Mark as verified" pill never turns green.

The new tests (`test_manual_check_renders_completion_checklist`) only assert the initial static `0 / 3` text — they never simulate a click, so this regression isn't caught.

**Fix:** restore the delegated `change` handler on `.cl-item input[type=checkbox]` that recomputes count, toggles `.cl-count.done` / `.cl-item.done`, and sets `.cl-bar > i` width for the containing `.checklist`.

### Issue B — "Sensitive values are masked" guidance is false (medium)

`_render_howto` (~lines 1553–1555) tells operators the report is safe to share because sensitive values are masked:

> **Sensitive values are masked** (secrets, IDs, tenant domains) so the report is safe to share and screenshot…

But `_mask_sensitive` is only called from `_render_manual_checklist` (lines 1791, 1799, 1811). Normal check cards render `res.result` via `_multiline_html` and non-manual `res.remediation` via `_md_links_to_html` **without masking** (`_render_check_card` at ~1497–1508). Any check that surfaces tenant IDs, subscription IDs, UPNs, object IDs, or tenant domains in its Detail or Next-step text will emit them verbatim, contradicting the guidance and creating a real risk that operators screenshot or share on the false promise the report is safe.

**Fix:** either route `res.result` and non-manual `res.remediation` through `_mask_sensitive` before rendering, or soften the copy so it says masking applies to manual-check step values only.

---

No other high-confidence issues found. The Python check-execution, bucket, and telemetry paths are untouched — this PR is presentation-only aside from the removed JS handler above.

daeunJe0ng:
Thanks for the re-review. Both fixed in d5a9159.

**Issue A:** Restored the delegated `change` handler in `_REPORT_SCRIPT`. It listens on `.cl-item input[type=checkbox]`, and per containing `.checklist` recomputes `.cl-count` (adds `.done` at 100%), toggles `.cl-item.done`, and sets `.cl-bar > i` width. Added `test_report_script_wires_completion_checklist_handler` so the handler can't be silently dropped again.

**Issue B:** Went with the first option and actually masked, rather than softening the copy. `res.result` (Detail) and non-manual `res.remediation` (Next step) now route through `_mask_sensitive` in `_render_check_card`, so UPNs and GUIDs are masked on every card, not just manual checklists. I also corrected the how-to copy to describe what masking really does (user emails and resource GUIDs redacted in prose; IDs inside links kept so links resolve) instead of overclaiming secrets and tenant domains. Added `test_non_manual_check_masks_detail_and_next_step`.

Full `tests/flightcheck` passes (1036).

amilandi (changes requested):
Thanks for the fast turnaround. The checklist handler is wired correctly (delegated `change` on document, per-checklist `refresh` covers count, `.cl-item.done`, `.cl-count.done`, and `.cl-bar > i` width), and the mask now correctly runs on non-manual `res.result` and `res.remediation`. Tests are the right shape.

One follow-up issue introduced by the mask ordering, though — I think this is a regression from the fix, medium severity.

## `_mask_sensitive` mangles GUIDs inside markdown link URLs, breaking maker-portal deep links

**File:** `solutions/ess-maker-skills/scripts/flightcheck/runner.py:1531, 1539` (and `_multiline_html` -> `_md_links_to_html`)

The ordering in `_render_check_card` is now:

```python
f'<dd class="detail">{_multiline_html(_mask_sensitive(res.result))}</dd>'
...
f'{_md_links_to_html(_mask_sensitive(res.remediation))}'
```

`_mask_sensitive` operates on the raw string with no awareness of markdown link syntax, and its GUID regex matches anywhere including inside `[label](url)` URLs. Multiple real check remediations already produce maker-portal deep links whose paths embed `env_id` (a GUID), for example `solutions/ess-maker-skills/scripts/flightcheck/checks/environment.py:887-891`:

```python
"[Power Automate → My flows]"
"(https://make.powerautomate.com/environments/{runner.env_id}/flows) "
...
"[solutions in this environment]"
"(https://make.powerapps.com/environments/{runner.env_id}/solutions) "
```

With this PR's ordering, the GUID inside the URL gets masked before `_md_links_to_html` sees it, so the emitted anchor becomes:

```html
<a href="https://make.powerautomate.com/environments/abc12345-****-****-****-************/flows" ...>Power Automate → My flows</a>
```

which navigates to an invalid environment ID. The updated how-to guide added in this same commit explicitly promises the opposite ("IDs inside links are kept so the links still resolve"), so this is a functional regression against the guide's contract and against the pre-PR behavior (which didn't mask non-manual text at all, so these links worked).

**Suggested fix (any of these):**

1. Parse markdown link URLs out before masking, mask only the outside-links portion, then reinsert the untouched links. Simplest structure:

    ```python
    def _mask_outside_links(text: str) -> str:
        import re
        parts = re.split(r'(\[[^\]]+\]\([^)]+\))', text)
        return ''.join(
            p if p.startswith('[') else _mask_sensitive(p) for p in parts
        )
    ```

    Then use `_mask_outside_links` in the two `_render_check_card` sites.

2. Do the mask *after* `_md_links_to_html` on the resulting HTML, and tighten the mask to skip GUIDs occurring inside `href="..."` — messier, more brittle.

3. Restrict the GUID regex not to match GUIDs preceded by URL-path characters (`[?&=/]`).

Option 1 is cleanest and keeps the mask function pure.

**Regression test to add** (mirrors the existing UPN test):

```python
def test_non_manual_link_url_with_guid_still_resolves(tmp_path):
    from flightcheck.runner import FlightCheckRunner, CheckResult, save_results
    env_id = "abc12345-1111-2222-3333-444455556666"
    runner = FlightCheckRunner(scope="test")
    runner.register("Environment", lambda _r: [
        CheckResult(
            checkpoint_id="ENV-004", category="Environment",
            priority="High", status="Warning",
            description="X",
            result="Y",
            remediation=f"Open [flows](https://make.powerautomate.com/environments/{env_id}/flows).",
        ),
    ])
    save_results(runner.run(), output_dir=str(tmp_path))
    html = (tmp_path / "report.html").read_text(encoding="utf-8")
    assert f"environments/{env_id}/flows" in html  # URL preserved
```

Everything else looks good. Once the link-URL case is preserved, LGTM.

amilandi (approved):
You're right — my "regression" reading was wrong. I re-fetched pr227 and looked at the raw bytes: `_mask_sensitive` already splits by `https?://[^\s)]+` URL runs and only masks prose outside URLs, so the env-id GUID in `[Power Automate → My flows](https://make.powerautomate.com/environments/<env_id>/flows)` survives verbatim and the `<a href>` is intact. The URL-aware logic was already in place at `d5a9159`.

Root cause of my confusion: I was reading `_mask_sensitive` off my *local checkout* (a different branch, `amilandin/adk-tenant-name-cache`) instead of `pr227`. The URL-splitting had never been present in that branch, so I read a stale version and posted the wrong review. My mistake — sorry.

The new tests are solid:
- `test_non_manual_link_url_with_guid_still_resolves` (552d897) locks the exact case I raised — non-manual `res.remediation` deep link with an env-id GUID survives masking and the href is preserved.
- `test_report_script_wires_action_item_collapsible` (204dc16) guards the `makeCollapsible` invocation you noticed was defined-but-never-called for `.ap-list > .ap-item` and `.ap-review ul > li`. Good separate catch.

Approving. Thanks for pushing back on the false alarm.

PR #229, opened 2026-08-07
flightcheck: add active Workday connector probe (WD-RUN-001) with shared consent gate

daeunJe0ng:
Title: flightcheck: add active Workday connector probe (WD-RUN-001) with shared consent gate

Summary

Adds WD-RUN-001, an active Workday connector health probe for FlightCheck, plus a consent gate for the mutating flow-probe it uses.
When the maker opts in with --runtime-reachability, the check briefly creates a transient Power Platform flow, runs one read-only Workday operation (default GetWorkerMe) through the agent's service-account connection, reads the connector HTTP status, then deletes the flow.
Without opt-in, the check falls back to the existing passive run-history assessment and creates nothing.
Failures are classified by layer (connector transport vs Workday business error vs indeterminate) with honest text when the probe cannot return a clear pass or fail.

Tracking work item: ADO 7500446 (https://dev.azure.com/O365Exchange/O365%20Core/_workitems/edit/7500446/).

Design rationale

The probe reuses the INFRA-003 live-egress primitive (live_egress_probe), so both checks share one transient-flow lifecycle and one consent prompt.
Consent is asked once per run at a single call site and shared by INFRA-003 and WD-RUN-001; one opt-in authorizes both transient flows, each self-cleaning.
The probe selects a non-invoker (service-account) Workday connection and validates the operation is read-only via an allowlist (get/list/read/raas_ prefixes; mutating prefixes rejected) before running, so it cannot mutate Workday data.
Cleanup runs in try/finally and also sweeps orphans on the next run, keyed off a stable flow display name.
Report text is written in maker-facing language. Internal terms (v2, active-probe fallback, layer names, plugin-tunneled SOAP) were removed so the output guides makers, not just the author.

Testing approach

Automated: 29 scenarios pass (12 probe branches + 17 consent-gate branches). Broader flightcheck suites 453 pass. ruff clean.
Gap branches confirmed: business-rule 409/422 -> FAILED, mutating-op -> passive, non-allowlisted op -> passive, prereq-missing -> SKIPPED, indeterminate-None.

Real-tenant (orged90ca58):
- No flag, non-TTY: consent gate reports it created nothing; WD-RUN-001 degrades to passive; no flow created.
- --runtime-reachability: transient flow created, read-only GetWorkerMe via service-account connection, HTTP 200 -> PASSED; orphan sweep clean.
- Forced nonexistent op: fails at activate -> indeterminate-None branch with honest text; no orphan left.
- --scope full: consent asked exactly once, shared by INFRA-003 and WD-RUN-001; both clean up.

Risks and assumptions

The active probe is reachable via --scope workday / workdayextension / full with --runtime-reachability, and (as of commit ed8bc5d) WD-RUN-001 is a registered checkpoint, so --checkpoint WD-RUN-001 now targets it directly instead of raising RegistryError.
A clean live invoke-time FAIL (egress OK plus Workday business error) is proven by the automated 409/422 test, not forced live, because that needs a valid op with failing params and risks repeated failed activations.
Passive-suffix accuracy: the passive fallback now claims readiness was assessed from run history only when the passive check reached a PASSED/FAILED verdict. When it could not read history (SKIPPED/NOT_CONFIGURED) the text says the live test did not run and history could not be assessed either. Fixed in the review follow-up below.

Reuse note

The consent gate and live-egress primitive are shared and will extend to ServiceNow, SAP SuccessFactors, and custom HTTP connectors. Those connectors do not yet have an active-probe check of their own; only the primitive is reused today.

Review follow-up (commit 36ebbaf)

- Replaced fragile substring reason-matching in WD-RUN-001 connection selection with an explicit not_configured flag, so rewording a reason string can no longer flip NOT_CONFIGURED vs the passive fallback.
- Dropped the unused layer value from the probe failure classifier; it now returns only the maker-facing cause (_workday_probe_failure_cause).
- Removed the internal 'v2' term from code comments and docstrings to match the maker-facing report copy.
- Added direct unit tests for interpret_connector_probe_response (succeeded / failed / skipped / error-only / missing-status / non-dict / bool-status / workday-fallback keys).
- Full flightcheck suite green after rebasing onto the latest branch head: 1067 passed.

Review follow-up (commit 05c5347)

- Passive WD-RUN-001 fallback no longer over-claims run history. The "assessed from recent Workday connector run history" wording is now used only when the passive check returned PASSED or FAILED. For SKIPPED / NOT_CONFIGURED it says the live test was not run and recent run history could not be assessed either, with the specific reason stated first.
- Added a unit test (test_passive_fallback_without_history_does_not_claim_history) asserting the non-assessed wording and that the history-assessed claim is absent.
- Full flightcheck suite green after rebase: 1085 passed.

Review follow-up (self-review, commits 319526f, f4fadb3, 5775a66, 019e117)

Addresses four findings from a dedicated ADK review of this PR.

- Consent naming (319526f): the Workday active probe binds a connection selected from the BAP connection list (pp.get_connections), which is independent of the .local/config.json connections map that drives the consent label. A Workday connection not recorded in that config would be probed without Workday appearing in the consent prompt (the PR #197 defect class). The consent gate now names Workday whenever the Workday active probe is in scope, on both the interactive-prompt and forced-flag paths. Two unit tests added in test_cli_consent.py.
- Probe failure causes (f4fadb3): only HTTP 200/400/500 were live-captured. The 401/403, 404/405 and 409/422 buckets in _workday_probe_failure_cause now attribute the likely cause to the HTTP status ("returned HTTP N, which typically means ...") instead of asserting the underlying fault as observed. No behavior change; the network, 400-indeterminate, and 500 buckets are unchanged, and existing assertions still hold.
- Dual key naming (5775a66): added comments documenting that the connector probe body emits every synchronous signal under both connector* and legacy workday* names, and that interpret_connector_probe_response reads connector* first with a workday* fallback that is load-bearing for replay of flightcheck_wd_connector_probe.yaml. Comment only, no behavior change.
- Recorder in-repo (019e117): the recorder that produced the validated cassette flightcheck_wd_connector_probe.yaml lived only on a spike branch and self-declared EXPERIMENTAL / UNVALIDATED, contradicting the INDEX registry tagging the cassette validated. Copied the recorder into tests/captures/ and reconciled its header to reflect the confirmed real capture; updated the INDEX row to point at the in-repo path (dropping the spike-branch reference).

Verification: pytest tests/flightcheck/checks/ tests/flightcheck/test_cli_consent.py -> 767 passed. ruff check on the five edited files clean. Not verified in this round: no live-tenant re-run of the recorder or the active probe.

Review follow-up (amilandi round 1, commit 4728fe7)

Addresses one question and three nits from the amilandi review.

- Q1 (400/BadRequest attribution): a probe HTTP 400 / BadRequest is now treated as indeterminate (_wd_probe_result_indeterminate) and degrades to the passive run-history signal instead of FAILING the connection. A wrong default operation and a genuine Workday business fault are indistinguishable at 400, so blaming a possibly-healthy connection was wrong. Added probe_reached plumbing to the passive context.
- S1 (stale doc): validation-matrix "only mutating path" corrected to "one of the kit's two mutating tenant paths"; added the WD-RUN-001 row.
- N2 (GUID vs display name): PASS/FAIL report rows now show properties.displayName, not the opaque BAP connection GUID.
- N3 (manual-fallback for Workday-only decline): suppressed the INFRA-style manual fallback text at the Workday-only-decline print sites in cli.py.

Verification: 36 targeted + 1091 full flightcheck passed; ruff clean.

Review follow-up (amilandi round 2, commit fab3ceb)

- Deterministic probe-connection selection: _select_workday_probe_connection sorts service-account candidates by connection GUID before picking [0], so BAP list order (undocumented) can no longer make the probe non-deterministic. Regression test added. (Note: issue #230 shows this candidate set is over-broad; the Gap 1 fix there supersedes this sort.)

Verification: 1092 full flightcheck passed (+1); ruff clean.

Review follow-up (apurvabanka, commit ed8bc5d)

Addresses two of four gaps from the apurvabanka review; the other two are tracked in #230.

- Gap 2 (registry): --checkpoint WD-RUN-001 raised RegistryError because the checkpoint was emitted but never registered. Added a fixed CheckpointSpec (clients PP_ADMIN+DATAVERSE, requires_dataverse_endpoint, WD-001 prereq to hydrate _workday_flows for the passive fallback) and added "WD-RUN" to OWNED_PREFIXES so the drift guard now enforces registration. Resolution test added.
- Gap 4 (concurrency): documented that cleanup_orphan_probe_flows sweeps by the shared probe-flow name, so concurrent runs against one environment can delete each other's in-flight flow. Same single-operator assumption INFRA-003 accepts; docstring notes the per-run-token fix. No behavior change.
- Deferred to #230: Gap 1 (the service-account filter reads runtimeSource from the BAP connection record, but runtimeSource is a flow-binding field absent from those records, so the filter is a no-op that fails open; needs positive ISU identification, not the naive default-flip) and Gap 3 (optional status_code guard on probe success).

Verification: python -m pytest tests/flightcheck/ -q -> 1093 passed (+1); ruff clean.


amilandi (dismissed):
Reviewed as amilandi (personal account). The WD-RUN-001-specific slice of this PR is thorough and I found no high-confidence bugs. Focused review on the ~15 files actually authored by this branch (skipped the ~17k lines of noise from #221/#226/#227 merged in).

**Verified correct:**
- Consent gate names Workday whenever `run_workday_checks` is in scope, on both forced (`--runtime-reachability`) and interactive prompt paths. `TestWorkdayScopeSurfacesConsent` asserts the label reaches the prompt/notice. PR #197 defect class closed.
- Read-only allowlist in `_workday_probe_config` lowercases first, rejects mutating prefixes, then requires `get/list/read/raas_` prefix. Rejection falls back to passive, never silently proceeds. Both explicit `--workday-op` and default `GetWorkerMe` are gated.
- Transient flow lifecycle: create→activate→callback→invoke wrapped in try/finally; pre+post orphan sweeps keyed on namespaced `flightcheck-wd-run-001-probe` with server-side `$filter`; every HTTP call carries a 60s timeout. `test_cleanup_on_invoke_error_deletes_created_flow` proves DELETE fires after a mid-invoke `ConnectionError`.
- Non-invoker selection: `_workday_runtime_source` treats anything other than `"invoker"` (including absent) as service-account. No-service-account → passive; no-Workday-connection-at-all → NOT_CONFIGURED.
- Failure-cause hedging: only live-verified 200/400/500 buckets speak declaratively; 401/403, 404/405, 409/422 all say "returned HTTP N, which typically means..." — matches commit f4fadb3's intent in bytes.
- Passive suffix only claims "assessed from recent Workday connector run history" for PASSED/FAILED; SKIPPED/NOT_CONFIGURED get "could not be assessed either". Covered by tests.
- Cassette scrubbed: GUIDs zeroed, SAS/Authorization/correlation IDs REDACTED, UPN/address mockified, hosts mockified. Dual `connector*`/`workday*` key fallback is load-bearing (cassette body uses `workday*`).

**Non-blocking observations (for future spec work, not this PR):**
- `service_account[0]` picks the first non-invoker Workday connection in BAP list order; BAP doesn't document stable ordering. On a tenant with two ISU connections, the "which one got probed" could silently drift between runs. Rare in practice.
- `x-ms-igw-tracking-id` header values in the cassette weren't scrubbed the way `x-ms-correlation-id` was. Trace IDs, not secrets — cosmetic only.
- Workday tenant path segment (`microsoft_dpt6`) is visible in captured response bodies. Not a credential, but worth confirming with the team if internal policy treats Workday tenant names as sensitive.

Approving. Ship it.

apurvabanka:
## Gaps


**1. Unknown/missing `runtimeSource` is treated as service-account (weakens the stated safety guarantee)**

`_select_workday_probe_connection` filters to non-invoker connections via `_workday_runtime_source(c) != "invoker"`, and `_workday_runtime_source` returns `""` when the field is absent (workday.py:34-46, 89-91). So any Workday connection where BAP doesn't emit `runtimeSource` is treated as service-account and probed. The PR's safety claim is "selects a non-invoker (service-account) connection," but the code fails *open*, not closed — an actual invoker/maker connection with an omitted field would be probed under the wrong identity. Impact is bounded (read-only `GetWorkerMe`), but the tests only cover explicit `"embedded"`/`"invoker"` values, not the missing-field case. Consider defaulting unknown→treat-as-invoker (skip) and adding that test.


**2. `--checkpoint WD-RUN-001` errors rather than "just doesn't target it"**

WD-RUN-001 is not in `registry.py` (confirmed). The PR frames this as "by design; not a regression." True that it pre-exists (the passive check was also unregistered), but `_resolve_or_raise` *raises* `RegistryError` for an unknown id, so `/flightcheck --checkpoint WD-RUN-001` gives a hard "does not resolve to any registered checkpoint" error. Since this PR promotes WD-RUN-001 to a first-class active check in SKILL.md and the validation matrix, the mismatch is now more visible/confusing. Worth a registry entry or an explicit note in user docs.


**3. Success is decided purely on action status, ignoring HTTP status**

`interpret_connector_probe_response` returns `succeeded=True` on `action_status == "Succeeded"` without checking `status_code` (live_egress_probe.py:276). The cassette only validates 200→Succeeded. If a connector ever marks the action Succeeded while surfacing a non-2xx `statusCode`, this would read as PASS. Low risk given connector behavior, but a defensive `status_code in 2xx` guard on the success branch would harden it.


**4. Concurrent runs can delete each other's in-flight probe flow**

`cleanup_orphan_probe_flows` sweeps by the stable display name `flightcheck-wd-run-001-probe`. Two overlapping FlightCheck runs in the same environment would each sweep the other's active flow. This is the same pattern INFRA-003 already uses, now duplicated for Workday — acceptable, but the orphan-sweep-by-name approach is not concurrency-safe and isn't called out in the risks section.

amilandi (approved):
Re-review of the three commits since my last approval (`4728fe7`, `fab3ceb`, `ed8bc5d`).

**All three land cleanly:**

1. `fab3ceb` — Deterministic connection selection. `_select_workday_probe_connection` now sorts the service-account candidates by `name` (immutable GUID) before picking `[0]`. Closes my earlier "BAP list order is undocumented" observation with a regression test that reverses the input and asserts the lexicographically-first GUID wins. Good.

2. `4728fe7` — HTTP 400 = indeterminate. Nicely reasoned: a synchronous 400 proves the probe reached Workday (egress, DNS, TLS, DLP, authz all worked) but a wrong default `GetWorkerMe` op and a genuine Workday validation fault are indistinguishable in the sync response, so scoring FAILED would blame the maker's connection for something the probe cannot establish. Routes to passive with `probe_reached=True` and honest "inconclusive" copy that does not claim "no Workday call was made". 401/403, 404/405, 409/422, 500 and no-HTTP still FAILED — correct. Also swaps the opaque BAP GUID for `properties.displayName` in maker-facing rows, and suppresses the INFRA-003 manual IP-allowlist block for Workday-only declines (matches SKILL.md).

3. `ed8bc5d` — Registers WD-RUN-001 in the registry so `--checkpoint WD-RUN-001` resolves instead of raising RegistryError; adds `"WD-RUN"` to `OWNED_PREFIXES` so the drift guard now enforces registration. Spec is right: clients = PP_ADMIN+DATAVERSE, `requires_dataverse_endpoint=True`, prereq WD-001 (hydrates `_workday_flows` for the passive fallback), category ordering test asserts External Systems runs before Workday. Also documents the shared-name orphan-sweep concurrency caveat on `cleanup_orphan_probe_flows` (same single-operator assumption INFRA-003 already accepts). No behavior change on the sweep — appropriately deferred.

**Acknowledged follow-ups (per commit body, non-blocking for this PR):**
- Gap 1 (fail-open identity filter): `runtimeSource` is a flow-binding attribute, not a BAP connection field, so `_workday_runtime_source` returning `None` for every BAP connection means the "service-account only" filter is effectively no-op today. The read-only allowlist bounds the blast radius (the probe cannot mutate Workday), but it does mean the probe may run as the maker's identity when only maker-invoker connections exist. Worth addressing in a follow-up before we lean on it for identity-scoped guarantees.
- Gap 3 (`status_code` guard): fine as a follow-up.

Re-approving.

daeunJe0ng:
Thanks @apurvabanka, all four are valid. Here is what I did with each.

**Gap 2 (registry) and Gap 4 (concurrency) - fixed in this PR (ed8bc5d).**
- `--checkpoint WD-RUN-001` was raising `RegistryError` because the checkpoint was emitted but never registered. Added a fixed `CheckpointSpec` (clients `PP_ADMIN`+`DATAVERSE`, `requires_dataverse_endpoint`, `WD-001` prereq to hydrate `_workday_flows` for the passive fallback) and added `"WD-RUN"` to `OWNED_PREFIXES`, so the drift guard now enforces registration. Resolution test added.
- Documented the `cleanup_orphan_probe_flows` concurrency caveat: it sweeps by the shared probe-flow name, so two runs against one environment can delete each other's in-flight flow. Same single-operator assumption INFRA-003 accepts; the docstring notes the per-run-token fix. No behavior change.

Verification: `python -m pytest tests/flightcheck/ -q` -> 1093 passed (+1); ruff clean.

**Gap 1 (fail-open service-account filter) - valid, and there is a deeper root cause. Tracked as follow-up #230.**

You are right that absent-source-defaults-to-service-account fails open. Digging in, the filter is actually a near no-op in production: `_workday_runtime_source` reads `runtimeSource` from the BAP `get_connections` record's `properties`, but `runtimeSource` is not a field on a connection record. It is a per-flow binding field that lives in the flow's `workflow.json` / `config.json` `requiredConnection` block. Evidence: the `flightcheck_pp_admin.yaml` cassette has zero `runtimeSource` occurrences, every production reader reads it off a flow binding (`config.json`, `ess_connection_binding.py`, `install_ess_agent.py`), and the author's own comment block above `_workday_runtime_source` documents it as living in `workflow.json`.

So for real connections `_workday_runtime_source` returns `""` and `!= "invoker"` keeps every connected Workday connection, including the OBO `ff0df` one. I did not apply the literal default-flip fix because it would over-correct: since the field is absent on virtually all real records, defaulting unknown to skip would disable the probe for nearly every real connection. The proper fix needs positive ISU identification (correlate to the flow's `connectionReferences`, or reuse the existing `_REF_SUFFIX_ROLES` fingerprint: `ff0df`=OBO, `0786a`/`d6081`=ISU). That work supersedes the fab3ceb determinism sort, which currently sorts an over-broad set.

**Gap 3 (`status_code` on probe success) - valid, low-risk, in #230.** Logic Apps only marks an action `Succeeded` on a 2xx connector response (documented in `live_egress_probe.py`), so it is defensive-only; if guarded it must preserve the `status_code is None` case.

Full detail and acceptance criteria in #230.

PR #235, opened 2026-08-13
flightcheck: fall back to cached tenant_name when Graph is unavailable

amilandi:
## Summary

FlightCheck telemetry emits the tenant display name (OII, privacy-approved) so dashboard series are human-readable. Both `cli.py` emit sites resolved `tenant_name` only via a live Microsoft Graph `/organization` call and left it as `""` whenever `graph` was `None` (infra-only scope, or Graph auth failed for want of `Organization.Read.All` consent) or the live lookup raised. The `.local/.tenant_name` cache that `adk_telemetry` already uses was never consulted.

## Evidence

Aria (prod, P30D, `essmakerkit_flightcheck_run`, `tenantClass == "customer"`):

- **483** runs across **7** tenants have `<blank>` `tenantName`.
- **197** runs across **8** tenants have a resolved `tenantName`.
- Tenant `935884d7-bdee-469b-a461-fcc530a3ac83` on ADK **0.4.24** shows **both** — 450 blank + 186 `EmployeeHub`, concurrent, latest `2026-08-12`. Same tenant, same version = code path split (not a version issue).

## Fix

At both `--scope` and single-`--checkpoint` emit sites in `flightcheck/cli.py`:

- If the live Graph lookup returned a name, proactively `cache_tenant_name(...)` (mirroring what `adk_telemetry.set_identity` already does), so any subsequent Graph-less run on this install keeps the label.
- If the live lookup returned `""` and we have a `tenant_id`, fall back to `telemetry.get_cached_tenant_name(tenant_id)`. The helper already enforces a same-tenant guard so a stale name is never leaked onto a different tenant.

Best-effort, never blocks the run.

## Test

- New regression test `test_tenant_name_falls_back_to_cache_when_graph_unavailable` seeds the cache, forces Graph auth to fail, and asserts the emit kwargs pick up the cached name.
- `pytest tests/flightcheck/test_cli.py tests/flightcheck/test_cli_single_checkpoint.py tests/flightcheck/test_telemetry.py tests/test_adk_telemetry.py` — **87 passed**.

## Non-goals / follow-ups

- Doesn't backfill blank rows already in Aria (retention <= 30 days makes them roll off).
- Doesn't add a Graph-consent-repair prompt — deliberately keeping this fully passive so telemetry never nags a maker.

### DIFF
--- a/solutions/ess-maker-skills/scripts/flightcheck/cli.py
+++ b/solutions/ess-maker-skills/scripts/flightcheck/cli.py
@@ -867,6 +867,12 @@ def _run_single_checkpoint(args):
         )
         # Best-effort tenant display name (OII; privacy-approved). Reuses the
         # already-authenticated Graph client when one was needed; never re-auths.
+        # Falls back to the persisted ``.local/.tenant_name`` cache when the
+        # live lookup is unavailable (e.g. infra-only scope where ``graph`` is
+        # None, or Graph auth failed for lack of ``Organization.Read.All``
+        # consent) so previously-resolved tenants keep their name on the event
+        # instead of emitting blank. Same-tenant guard is enforced inside the
+        # cache helper.
         tenant_name = ""
         try:
             if graph is not None:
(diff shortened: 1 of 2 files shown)


amilandi:
Superseded by #237 which combines this fix with the auth.py bootstrap + silent-Graph fallback.

PR #236, opened 2026-08-13
adk-telemetry: reduce blank tenant_name via reordered bootstrap + /me fallback

amilandi:
## Problem

Prod ADK telemetry emits **blank `tenant_name`** on roughly 74–80% of non-`adk.api.call` events (measured on `adk_session_start`, `adk_capability_use`, `adk_build_*`, `adk_agent_deploy` over the last 30 days). Dashboard trend lines and per-tenant filters degrade for those tenants because the label falls back to the raw Entra GUID or shows blank.

Data support (prod, single sample tenant `935884d7-…`, P30D):

| `hasNamedSession` | `ranFlightCheck` | instances |
|---|---|---|
| 0 | 0 | 18 |
| 0 | 1 | 11 |
| 1 | 1 |  9 |

The 18 no-name/no-FC installs never ran FlightCheck (so no interactive Graph consent) *and* the silent Graph resolver returned empty at `adk.session.start` time. The 11 no-name/did-FC installs ran FlightCheck *after* the first `session_start` emit, so early events went out blank until the FC-cached name kicked in for later events. Only 9 installs already emit names correctly.

## Root causes and fixes

**1. Bootstrap order in `auth.py`.** `adk_telemetry.start_session` fired *before* the silent Graph resolution ran, so the very first `adk.session.start` on a fresh install always went out with blank `tenant_name` — even when silent resolution *would* have succeeded on that same call. Reordered so the flow is now:

`\\\
resolve_tenant_display_name_silent -> set_identity(tenant_name=...) -> start_session
\\\`

Regression test in `tests/scripts/test_auth.py` pins the ordering (and asserts we still emit `start_session` when silent resolution returns blank).

**2. Silent Graph scope in `resolve_tenant_display_name_silent`.** The resolver requested `Organization.Read.All` only. That scope requires **admin consent**, which the majority of enterprise tenants have not pre-granted, so the FOCI shared refresh token from the Dataverse interactive sign-in cannot silently redeem it — the resolver returns `""` even though the user is fully signed in. Added a fallback path: on empty result, retry silently with `User.Read` and fetch `/me?$select=companyName`. `User.Read` is user-consent-only and is a *default static permission* granted at first-party sign-in, so the same shared RT typically redeems it silently. Admins routinely set the signed-in user's `companyName` attribute to the tenant display name, so this recovers a useful label for the many customers that would otherwise report as blank. Same OII classification we already have for `tenant_name`.

## Non-goals / follow-ups

- **Retention rolls off in ≤30 days** — this fix is forward-only. Existing blank rows are not backfilled.
- Not touching the `Organization.Read.All` scope requested at Dataverse sign-in itself — adding `extra_scopes_to_consent` risks breaking auth in tenants where admin consent for that scope is *required*, and the fallback covers most of the same ground silently.

## Tests

`\\\
pytest tests/flightcheck/test_telemetry.py tests/flightcheck/test_graph_client.py tests/scripts/test_auth.py tests/flightcheck/test_cli.py tests/flightcheck/test_cli_consent.py
83 passed in 4.11s
\\\`

New tests: ordering + fallback-scope + missing-companyName + both-silent-fail cases.

### DIFF
--- a/solutions/ess-maker-skills/scripts/auth.py
+++ b/solutions/ess-maker-skills/scripts/auth.py
@@ -267,15 +267,15 @@ def authenticate(env_url):
         claims = result.get("id_token_claims", {}) or {}
         tenant_id = claims.get("tid", "") or tenant
         adk_telemetry.maybe_print_notice()
-        adk_telemetry.start_session(
-            tenant_id=tenant_id,
-        )
-        # Best-effort: resolve the tenant's org display name via a SILENT-ONLY
-        # Graph token (never prompts) and record it so ADK telemetry carries
-        # tenant_name even when the maker never runs FlightCheck. The Dataverse
-        # sign-in above usually leaves a first-party (FOCI) refresh token that
-        # silently satisfies the read scope; set_identity caches the name for
-        # later ADK processes. Silent failure just leaves tenant_name empty.
+        # Resolve the tenant's display name via a SILENT-ONLY Graph token
+        # BEFORE emitting adk.session.start, so the very first ADK event on a
+        # fresh install carries tenant_name (instead of blank until FlightCheck
+        # is later run). The Dataverse sign-in above usually leaves a
+        # first-party (FOCI) refresh token that silently satisfies at least
+        # one of Organization.Read.All / User.Read; the graph_client helper
+        # tries both. Silent failure is fine — set_identity is only called on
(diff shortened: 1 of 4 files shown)


amilandi:
Superseded by #237 which combines this fix with the FlightCheck cli.py cache fallback.

PR #237, opened 2026-08-13
adk-telemetry: fix blank tenant identity (name + id) across auth, silent Graph, FlightCheck emit, and SKILL.md subprocesses

amilandi:
## Problem

Prod ADK telemetry emits **blank tenant identity** on the majority of non-`adk.api.call` events, breaking dashboards in two distinct ways:

**Blank `tenant_name`** (roughly 74-80% of `adk_session_start`, `adk_capability_use`, `adk_build_*`, `adk_agent_deploy` over P30D). FlightCheck too: `essmakerkit_flightcheck_run` shows **483** blank / **197** named. Per-tenant trend lines and filters degrade because the label falls back to the raw Entra GUID or is empty.

Data support (prod, sample tenant `935884d7-...`, P30D):

| hasNamedSession | ranFlightCheck | instances |
|---|---|---|
| 0 | 0 | 18 |
| 0 | 1 | 11 |
| 1 | 1 |  9 |

**Blank `tenant_id`** on capability events emitted from SKILL.md-driven skills. The External *Capability Usage by Type* chart (filter: `tenant_class == "customer"`) shows only `setup`, `backup_template_configs`, `restore_template_configs` — implying no customer ever runs anything else. In reality every capability lands, but under `tenant_class = "unknown"` because the shim runs in a fresh Python subprocess with no identity.

Prod P30D `adk_capability_use` by `(adk_capability, tenant_class)`:

| capability | customer | unknown |
|---|---:|---:|
| setup | 86 (10 tenants) | 113 |
| backup_template_configs | 46 (1) | 134 |
| restore_template_configs | 42 (1) | 134 |
| connect | 0 | 71 |
| topic_update | 0 | 28 |
| topic_create | 0 | 7 |
| workflow_update | 0 | 6 |
| evaluations | 0 | 4 |
| troubleshoot | 0 | 4 |
| topic_delete | 0 | 1 |
| (raw "unknown") | 0 | 106 |

This PR fixes both classes across all four affected code paths.

### 1. `auth.py` bootstrap order

`adk_telemetry.start_session` fired *before* silent Graph resolution, so the first `adk.session.start` on a fresh install always went out with blank `tenant_name` — even when silent resolution *would* have succeeded on that same call. Reordered:

```
resolve_tenant_display_name_silent -> set_identity(tenant_name=...) -> start_session
```

Regression test in `tests/scripts/test_auth.py` pins the ordering (and asserts we still emit `start_session` when silent resolution returns blank).

### 2. Silent Graph scope fallback in `resolve_tenant_display_name_silent`

The resolver requested `Organization.Read.All` only. That scope requires **admin consent**, which the majority of enterprise tenants have not pre-granted, so the FOCI shared refresh token from the Dataverse interactive sign-in cannot silently redeem it — the resolver returns `""` even though the user is fully signed in.

Added a fallback path: on empty result, retry silently with `User.Read` and fetch `/me?$select=companyName`. `User.Read` is user-consent-only and is a *default static permission* granted at first-party sign-in, so the same shared RT typically redeems it silently. Admins routinely set the signed-in user's `companyName` attribute to the tenant display name, so this recovers a useful label for many customers that would otherwise report blank. Same OII classification we already have for `tenant_name`.

### 3. FlightCheck `cli.py` cache fallback

Both `cli.py` emit sites (`--scope` and single-`--checkpoint`) resolved `tenant_name` only via a live Graph `/organization` call and left it as `""` whenever `graph` was `None` (infra-only scope, or Graph auth failed) or the live lookup raised. The `.local/.tenant_name` cache that `adk_telemetry` already uses was never consulted.

- If the live Graph lookup returned a name, proactively `cache_tenant_name(...)` (mirroring what `adk_telemetry.set_identity` already does) so subsequent Graph-less runs on this install keep the label.
- If the live lookup returned `""` and we have a `tenant_id`, fall back to `telemetry.get_cached_tenant_name(tenant_id)`. The helper already enforces a same-tenant guard so a stale name is never leaked onto a different tenant.

Best-effort, never blocks the run.

### 4. `tenant_id` persistence for fresh-subprocess emits

SKILL.md-driven skills (topic/workflow authoring, connect, cleanup, troubleshoot, evaluations) record capability usage by invoking `python scripts/emit_capability.py <cap>` as a fresh Python subprocess. That subprocess never calls `set_identity`, so its in-memory `_IDENTITY["tenant_id"]` is empty; `classify_tenant("")` returns `"unknown"`; the event never appears on the External dashboard. (The huge `unknown` bucket alongside the `customer` counts for `setup` / backup / restore in the table above has the same root cause on some code paths.)

Mirror the `.instance_id` persistence pattern for `tenant_id`:
- `flightcheck.telemetry.cache_tenant_id` / `get_cached_tenant_id` write/read `.local/.tenant_id` (best-effort, IO-swallowing).
- `adk_telemetry.set_identity` unconditionally caches the raw GUID (independent of `tenant_name`, which may not resolve when Graph is unavailable — otherwise this fallback would only work for tenants that already had a Graph-resolved name).
- `adk_telemetry.common_dimensions` falls back to the disk cache **only** when the caller passed no explicit `tenant_id` AND the in-memory identity is empty. An explicit empty kwarg still honors the caller's intent (anonymous emits stay anonymous); an explicit non-empty kwarg still wins over the cache (multi-tenant safety).

## Non-goals / follow-ups

- **Retention rolls off in <= 30 days** — forward-only fix. Existing blank rows are not backfilled.
- Not touching the interactive sign-in scope in `auth.py` — adding `extra_scopes_to_consent=["Organization.Read.All"]` risks breaking auth in tenants where admin consent for that scope is *required*, and the `/me` fallback covers most of the same ground silently.
- No Graph-consent-repair prompt — keeps telemetry fully passive so it never nags a maker.
- FlightCheck telemetry is not affected by the fix-4 gap specifically (single-process flow keeps a populated `_IDENTITY` for the whole run) but harmlessly benefits from the fallback if any future refactor ever re-enters FC emit standalone.

## Tests

```
pytest tests/flightcheck/test_telemetry.py tests/flightcheck/test_graph_client.py \
       tests/scripts/test_auth.py tests/flightcheck/test_cli.py \
       tests/flightcheck/test_cli_consent.py tests/flightcheck/test_cli_single_checkpoint.py \
       tests/test_adk_telemetry.py
```

New tests cover:
- **Fix 1**: auth bootstrap ordering + silent-blank degradation.
- **Fix 2**: `/me` fallback success + missing companyName + both-scopes-fail.
- **Fix 3**: FlightCheck cache-fallback-when-Graph-unavailable.
- **Fix 4**: subprocess scenario resolves to `tenant_class = "customer"`; explicit tenant_id kwarg wins over cache; explicit empty kwarg does NOT fall back (anonymous emits stay anonymous); `cache_tenant_id` round-trip + empty-input no-op + whitespace strip.

## Supersedes

Closes #235 and #236 — combined into one review.

### DIFF
--- a/solutions/ess-maker-skills/scripts/adk_telemetry.py
+++ b/solutions/ess-maker-skills/scripts/adk_telemetry.py
@@ -218,6 +218,13 @@ def set_identity(
     )
     _IDENTITY["tenant_id"] = tenant_id or ""
     _IDENTITY["tenant_name"] = tenant_name or ""
+    # Persist tenant_id to disk so a *later* Python subprocess (e.g. the
+    # emit_capability.py shim launched from a SKILL.md step) — which never
+    # calls set_identity itself — can still stamp the tenant on its events.
+    # Without this, subprocess-emitted events land with an empty tenant_id,
+    # classify as "unknown", and never show on the customer-filtered dashboard.
+    if _IDENTITY["tenant_id"]:
+        _fc.cache_tenant_id(_IDENTITY["tenant_id"])
     # When a Graph-capable flow (FlightCheck) resolves the org display name,
     # persist it so ADK events emitted later in a *different* process (which
     # only has a Dataverse/BAP token and can't resolve it) can reuse it. The
@@ -373,6 +380,14 @@ def common_dimensions(
 ) -> dict[str, Any]:
     """Build the dimensions present on every event (spec Common Dimensions)."""
     tid = _IDENTITY["tenant_id"] if tenant_id is None else tenant_id
+    # Fall back to the disk-persisted tenant_id when this process never
+    # called set_identity (subprocess launched by emit_capability.py from a
+    # SKILL.md step). Without this, capability events emitted by the shim
+    # carry an empty tenant_id, classify as "unknown" via classify_tenant(""),
+    # and never appear on the customer-filtered External dashboard even
+    # though the maker's earlier auth flow knew the tenant.
+    if not tid and tenant_id is None:
+        tid = _fc.get_cached_tenant_id()
     tname = _IDENTITY["tenant_name"] if tenant_name is None else tenant_name
     # Fall back to the org display name a prior Graph-capable run (FlightCheck)
     # cached for THIS tenant, so pure-ADK events (session/build/deploy/
--- a/solutions/ess-maker-skills/scripts/adk_telemetry.py
+++ b/solutions/ess-maker-skills/scripts/adk_telemetry.py
@@ -198,8 +198,38 @@ _SESSION_LOCK = threading.Lock()
 
 
 # --- Identity model (spec) ------------------------------------------------
+# Canonical Entra tenant GUID: 8-4-4-4-12 lowercase-hex, dashes at fixed
+# offsets. Enforced *only* on tenant_id (the raw OII we ship to Aria) so
+# obvious test-fixture placeholders like ``"tenant-id"`` don't slip into
+# the prod stream and inflate the customer bucket. Also applied when
+# reading back from the on-disk ``.tenant_id`` cache, so a torn / legacy /
+# hand-edited file can never bypass the check.
+_GUID_RE = re.compile(
+    r"^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$"
+)
+
+
+def _sanitize_tenant_id(tenant_id: str) -> str:
+    """Return ``tenant_id`` iff it looks like a canonical GUID, else ``""``.
+
+    Never raises. Empty input stays empty. Whitespace/case are tolerated so
+    a caller that passes ``"  ABCDEF01-2345-6789-ABCD-EF0123456789  "`` still
+    gets a valid identity (normalized to lowercase). Anything else
+    (``"tenant-id"``, ``"unknown"``, truncated GUIDs, org display names
+    accidentally routed here) normalizes to ``""`` so downstream
+    ``classify_tenant`` labels the event as ``"unknown"`` rather than as
+    a fake customer.
+    """
+    if not tenant_id:
+        return ""
+    v = tenant_id.strip().lower()
+    return v if _GUID_RE.match(v) else ""
+
+
 def set_identity(
-    tenant_id: str = "", instance_id: str | None = None, tenant_name: str = ""
+    tenant_id: str | None = None,
+    instance_id: str | None = None,
+    tenant_name: str | None = None,
 ) -> dict[str, str]:
     """Record the install + tenant identity for all subsequent events.
 
@@ -212,25 +242,53 @@ def set_identity(
     organization display name (OII identifying the enterprise tenant, not a
     person; privacy-approved without a Data Profile update) — best-effort,
     stored as ``""`` when unavailable. No developer/user id is collected.
+
+    All three parameters use a ``None`` sentinel to mean "don't touch". This
+    lets a later call like ``set_identity(tenant_id=<guid>)`` add / refresh
+    the tenant id without clobbering a tenant_name a previous call already
+    resolved (the reason ``start_session`` used to blank out ``tenant_name``
+    when it re-bootstrapped identity right after auth). Passing ``""`` for
+    any field explicitly clears it.
+
+    When ``tenant_id`` changes to a genuinely different GUID and the caller
+    does not simultaneously supply a matching ``tenant_name``, the stored
+    ``tenant_name`` is dropped — a name resolved for tenant A must never be
+    stamped on tenant B's events.
+
+    Any ``tenant_id`` that is not a well-formed Entra tenant GUID is
+    silently normalized to ``""`` (see ``_sanitize_tenant_id``).
     """
-    _IDENTITY["instance_id"] = (
-        _fc.get_instance_id() if instance_id is None else (instance_id or "")
-    )
-    _IDENTITY["tenant_id"] = tenant_id or ""
-    _IDENTITY["tenant_name"] = tenant_name or ""
-    # Persist tenant_id to disk so a *later* Python subprocess (e.g. the
+    if instance_id is not None:
+        _IDENTITY["instance_id"] = instance_id or _fc.get_instance_id()
+    elif not _IDENTITY["instance_id"]:
+        _IDENTITY["instance_id"] = _fc.get_instance_id()
+
+    if tenant_id is not None:
+        new_tid = _sanitize_tenant_id(tenant_id)
+        if (
+            new_tid
+            and _IDENTITY["tenant_id"]
+            and new_tid != _IDENTITY["tenant_id"]
+            and tenant_name is None
+        ):
+            # Tenant switch with no fresh name supplied: drop the stale one
+            # rather than stamp tenant A's name onto tenant B's events.
+            _IDENTITY["tenant_name"] = ""
+        _IDENTITY["tenant_id"] = new_tid
+
+    if tenant_name is not None:
+        _IDENTITY["tenant_name"] = tenant_name or ""
+
+    # Persist tenant_id so a *later* Python subprocess (e.g. the
     # emit_capability.py shim launched from a SKILL.md step) — which never
     # calls set_identity itself — can still stamp the tenant on its events.
-    # Without this, subprocess-emitted events land with an empty tenant_id,
-    # classify as "unknown", and never show on the customer-filtered dashboard.
     if _IDENTITY["tenant_id"]:
         _fc.cache_tenant_id(_IDENTITY["tenant_id"])
-    # When a Graph-capable flow (FlightCheck) resolves the org display name,
-    # persist it so ADK events emitted later in a *different* process (which
-    # only has a Dataverse/BAP token and can't resolve it) can reuse it. The
-    # cache is keyed by tenant_id in flightcheck.telemetry.cache_tenant_name.
-    if tenant_name:
-        _fc.cache_tenant_name(_IDENTITY["tenant_id"], tenant_name)
+    # Persist the tenant name so ADK events emitted later in a *different*
+    # process (which only has a Dataverse/BAP token and can't resolve it)
+    # can reuse it. Only cache when both fields are present.
+    if _IDENTITY["tenant_name"] and _IDENTITY["tenant_id"]:
+        _fc.cache_tenant_name(_IDENTITY["tenant_id"], _IDENTITY["tenant_name"])
     return dict(_IDENTITY)
 
 
@@ -385,9 +443,11 @@ def common_dimensions(
     # SKILL.md step). Without this, capability events emitted by the shim
     # carry an empty tenant_id, classify as "unknown" via classify_tenant(""),
     # and never appear on the customer-filtered External dashboard even
-    # though the maker's earlier auth flow knew the tenant.
+    # though the maker's earlier auth flow knew the tenant. Re-run the cache
+    # value through the same GUID sanitizer used in ``set_identity`` so a
+    # torn / legacy / hand-edited file can never bypass the check.
     if not tid and tenant_id is None:
-        tid = _fc.get_cached_tenant_id()
+        tid = _sanitize_tenant_id(_fc.get_cached_tenant_id())
     tname = _IDENTITY["tenant_name"] if tenant_name is None else tenant_name
     # Fall back to the org display name a prior Graph-capable run (FlightCheck)
     # cached for THIS tenant, so pure-ADK events (session/build/deploy/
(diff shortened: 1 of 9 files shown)


daeunJe0ng:
[copilot-review]

## Verdict: Approve with nits (no blocking issues)

A well-scoped, well-tested telemetry-quality fix for a real, data-supported prod problem (74-80% blank `tenant_name`, large `unknown` `tenant_class` bucket). The identity model is handled carefully: None-sentinels, tenant-switch safety, GUID sanitization, atomic writes for the new cache. Remaining items are an auth-path latency question, a privacy classification confirmation, and small consistency/doc nits.

## Design verdict
Sound and appropriately conservative. The four fixes attack four independent root causes, all forward-only (no backfill, retention <=30d, correctly called out as non-goal). Reusing the existing `.local/` cache pattern (`.instance_id` -> `.tenant_id`) is the right, minimal design. The `None`-sentinel refactor of `set_identity` correctly resolves the "`start_session` re-bootstrap wipes the name auth.py just resolved" hazard, and multi-tenant safety (drop stale name on tenant switch, same-tenant guard on the name cache, GUID sanitizer blocking placeholders from the customer bucket) is genuinely thoughtful. Telemetry stays fail-open throughout. No architectural objection.

## Findings

### Blocking
None.

### Should-fix

**1. Added synchronous Graph round-trip on the interactive auth critical path for the majority (admin-consent-gap) tenants** (`auth.py:270-288`, `graph_client.py:_try_me_company_name_lookup`)
- What: `authenticate()` calls `resolve_tenant_display_name_silent()` synchronously before returning the access token. Previously, in the common "Organization.Read.All not admin-consented" case, `acquire_token_silent(_ORG_READ_SCOPE)` returned `None` and the resolver bailed with zero network I/O. Now that same case falls through to `_try_me_company_name_lookup`, which performs a live `GET /me?$select=companyName` (15s timeout, `_SESSION` retries on 429/5xx with `backoff_factor=1`). That is a new blocking network call on the sign-in hot path for exactly the tenant segment this PR targets as the majority.
- Why it matters: it delays returning the token to the maker on every fresh sign-in in the common path; a slow/redirecting Graph endpoint can add seconds (worst case tens of seconds with retries) to interactive auth. It is wrapped so it cannot break auth, but it does slow it.
- Suggested fix: either (a) resolve the name off the critical path (detached worker like `emit_capability.py`, or after the token is returned), or (b) cap total resolution wall-time (short deadline / single attempt / no retry on the `/me` probe). At minimum, confirm the added latency is acceptable given how common this path is claimed to be.

### Question

**2. Privacy classification of the new `/me?$select=companyName` data source** (`graph_client.py:_try_me_company_name_lookup`)
- The PR asserts `companyName` is "same OII classification we already have for `tenant_name`." That is plausible (org-level attribute, and only `companyName` is `$select`ed, not name/UPN/oid), but this introduces a new Graph endpoint (`/me`, a user resource) and a new profile attribute into the telemetry stream. Has the privacy/Data-Profile owner explicitly signed off on reading `/me` companyName as OII (vs EUPI), given `companyName` is a per-user, admin-editable free-text field that a tenant could populate with non-org text? A one-line confirmation (or ADO link) in the PR would close this out.

**3. `cache_tenant_name` write is non-atomic while `cache_tenant_id` is atomic** (`telemetry.py:cache_tenant_name` vs `cache_tenant_id`)
- The PR (rightly) made `cache_tenant_id` atomic (`tempfile` + `os.replace`) and versioned, citing torn-read safety. But the sibling `cache_tenant_name`, which this PR also modifies (adds `source` + read-modify-write ranking), still does a plain `open(path, "w")` after a read. A concurrent reader can observe a truncated `.tenant_name`. It degrades safely (`get_cached_tenant_name` swallows `ValueError`/`OSError` -> `""`), so it is not a leak, but it is an inconsistency the PR's own rationale argues against. Intentional (out of scope), or worth mirroring the atomic pattern here too?

### Nit (non-blocking)

**4. `get_cached_tenant_id` legacy branch returns the raw string unvalidated, contradicting its own docstring** (`telemetry.py:362+`)
- The docstring says "Any other content (truncated GUID, JSON at a future schema version, garbage) is discarded," but the legacy raw-string branch does `return raw` with no GUID check. In practice the only caller (`common_dimensions`) re-runs it through `_sanitize_tenant_id`, so it is safe, but the safety lives in the caller, not here as the docstring implies. Either tighten the docstring or validate in the reader so the guarantee is local.

**5. Duplicated cache-fallback block across the two cli.py emit sites** (`cli.py` `_run_single_checkpoint` ~line 885 and `main` ~line 1359)
- The `if not tenant_name and tenant_id: ... elif tenant_name and tenant_id: cache...` block is duplicated (with a cosmetic `(tenant_id or "")` vs `tenant_id` difference). Minor; a shared `_resolve_or_cache_tenant_name(telemetry, tenant_id, tenant_name)` helper would remove the drift risk. Respecting the intentional separation of the two call sites, this is optional.

**6. Behavior change: explicit `instance_id=""` now falls back to `get_instance_id()`** (`adk_telemetry.set_identity`)
- Old: `instance_id=""` stored `""`. New: `"" or _fc.get_instance_id()` -> fetched id. No known caller passes an explicit empty instance_id, so harmless, but it is an undocumented semantic shift in the `""`-clears convention the docstring otherwise establishes for the other two fields.

## What's good (genuine)
- Root-cause rigor: each fix is backed by prod Aria numbers and tied to a specific mechanism; the PR description is exemplary.
- Multi-tenant safety is handled correctly end-to-end: GUID sanitizer keeps placeholders (`"tenant-id"`) out of the customer bucket; tenant-switch drops a stale name; the name cache's `tenant_id` guard prevents cross-tenant leakage; explicit-empty kwarg stays anonymous. Tests pin each.
- The `/organization` vs `/me` `source`-ranking so a per-user `me` label can't silently downgrade the authoritative org label is a nice touch, and the transport-ok-vs-auth-fail distinction (`_try_organization_lookup` returning `(name, transport_ok)`) correctly avoids a pointless second call on a dead network.
- `cache_tenant_id` is atomic + versioned with a legacy-read shim: good forward/backward-compat hygiene.
- Tests are strong: they assert on real OneCollector envelopes (not just call order), and the end-to-end subprocess test actually spawns `emit_capability.py` in a fresh interpreter with a broken proxy and reads the buffered event, which genuinely proves the cross-process bridge.

## Verification
- Cloned branch `amilandin/adk-blank-tenant-name-fix`; read all 9 changed files in full against the surrounding source (`adk_telemetry.set_identity`/`common_dimensions`/`start_session`, `graph_client` resolver, `telemetry` cache helpers, `auth.authenticate`, both `cli.py` emit sites, `emit_capability.py`, `classify_tenant`).
- Ran the PR's full referenced test set: `pytest tests/flightcheck/test_telemetry.py test_graph_client.py test_cli.py test_cli_consent.py test_cli_single_checkpoint.py tests/scripts/test_auth.py tests/test_adk_telemetry.py` -> 149 passed in ~8s (Python 3.11).
- Did not run: full repo test suite, linters/type-checkers, or any live Graph/network calls (no `--run-live`). The latency concern (#1) is a code-path analysis, not a measured timing.
- Gaps: did not measure real-world `/me` latency, and could not independently verify the privacy classification of `companyName` (#2); that needs the data-profile owner.

## Summary table

| Item | Severity | Area | File | Issue |
|------|----------|------|------|-------|
| 1 | Should-fix | Performance/UX | auth.py / graph_client.py | New synchronous `/me` Graph GET on the interactive auth hot path for the common admin-consent-gap tenants |
| 2 | Question | Privacy | graph_client.py | Confirm `/me?$select=companyName` OII sign-off (new user endpoint/attribute) |
| 3 | Question | Correctness/consistency | telemetry.py | `cache_tenant_name` non-atomic while sibling `cache_tenant_id` is atomic |
| 4 | Nit | Docs/robustness | telemetry.py | `get_cached_tenant_id` legacy branch returns raw string; docstring claims it is discarded |
| 5 | Nit | Complexity | cli.py | Duplicated cache-fallback block across both emit sites |
| 6 | Nit | Backward-compat | adk_telemetry.py | Explicit `instance_id=""` now falls back to fetched id (semantic shift) |

daeunJe0ng:
[copilot-review] Addendum to the review above (design alternative for #1 and the `unknown` bucket)

Following up on Should-fix #1 with a design option I should have called out in the original review. This is a suggestion, not a blocker.

**`tenant_class` may not need a Graph call at all; the tenant GUID is already in the token.**

The auth path in this PR already extracts the tenant GUID from the token claims:

```python
claims = result.get("id_token_claims", {}) or {}
tenant_id = claims.get("tid", "") or tenant
```

Since `classify_tenant` buckets on the GUID (not the display name), classification is fully determined by the `tid` claim you already hold after a successful sign-in, with zero network I/O and no admin consent. If `tid` reliably reaches the emit site, the large `unknown` `tenant_class` bucket should be solvable from the token alone. In that case the remaining `unknown` events are more likely an identity-propagation or unauthenticated-emit problem than a lookup problem, which is a different fix than adding a Graph call.

This reframes #1: the inline `/me?$select=companyName` call is needed only for the human-readable `tenant_name`, never for `tenant_class`. Since the name is a best-effort telemetry label and the pipeline is fail-open, resolving it off the sign-in critical path (or with a hard deadline) costs nothing in correctness and removes the added interactive-auth latency.

Net suggestion, unchanged verdict (Approve with nits):
- Confirm `tid` propagates all the way to the emit site so `tenant_class` is driven by the token claim, not by Graph.
- Keep the name lookup, but move it off the auth hot path so no sign-in ever waits on a telemetry-only field.

Caveat: I verified the `tid`-claim extraction exists in the auth diff, but I did not trace end-to-end that the same GUID always reaches every emit site; that is worth a quick confirmation.

amilandi:
Round-2 review addressed in [`0a01196`](https://github.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/commit/0a01196):

**Should-fix #1 (auth-path latency)** — the /organization and /me silent lookups now use a bounded 3s socket timeout via plain `requests.get` (not the module-level `_SESSION`, whose 3× 429/5xx retry with `backoff_factor=1` would compound to ~7s per lookup on a throttled tenant). Worst-case `authenticate()` overhead is now bounded to ~6s instead of ~15–40s in the admin-consent-gap case. Your addendum's core point holds: `tenant_class` is driven by the `tid` claim already extracted from the token, so a timeout here only degrades `tenant_name` to blank (recoverable on the next FlightCheck run) — never a misclassification.

I stopped short of moving name resolution fully off the auth hot path because the design goal of this PR is to make the *first* `adk.session.start` carry `tenant_name` (currently blank 74–80% of the time). A bounded synchronous call keeps that guarantee in the majority of tenants where silent redemption succeeds fast, while the 3s ceiling makes the worst case acceptable. Happy to move to a fully async path in a follow-up if you'd prefer.

**Inline nits** — all three addressed in the same commit:
- `common_dimensions` explicit `tenant_id=` kwarg now flows through `_sanitize_tenant_id` (single choke point).
- `cache_tenant_name` now uses `tempfile` + `os.replace` (atomic write, matching `cache_tenant_id`).
- `get_cached_tenant_id` legacy raw-string branch now validates against `_GUID_RE` before returning.

Regression tests added for the kwarg-sanitization and legacy-string paths.

daeunJe0ng:
[copilot-review] Re-review of round-2 commit `0a01196`

Thanks for the fast turnaround. I fetched `0a01196`, read the diff, and ran the touched tests. All items I raised are addressed. Verdict stays **Approve** (nits only), now with the Should-fix resolved.

**Should-fix #1 (auth-path latency): resolved.** Both silent lookups now use plain `requests.get` with a 3s `_SILENT_LOOKUP_TIMEOUT` instead of the retrying `_SESSION`, so the worst-case sign-in overhead is bounded rather than compounding through 429/5xx backoff. Agreed with your reasoning that a timeout here only degrades `tenant_name` to blank (recoverable next run) and never causes a misclassification, since `tenant_class` rides the `tid` claim already extracted from the token. Keeping it bounded-synchronous to preserve the first-`adk.session.start` name guarantee is a reasonable call; a fully async path as a follow-up is fine by me, not required for this PR.

**Question #3 (non-atomic name cache): resolved.** `cache_tenant_name` now writes via `tempfile` + `os.replace`, matching `cache_tenant_id`, so a concurrent subprocess reader can't see a truncated file.

**Nit #4 (legacy branch vs docstring): resolved.** The legacy raw-string branch now validates against `_GUID_RE` before returning, so the guarantee is local and matches the docstring.

**Inline kwarg nit: resolved.** `common_dimensions` now routes the explicit `tenant_id=` kwarg through `_sanitize_tenant_id`, closing the last path by which an `emit_*` caller could push a non-GUID into the customer bucket. The two added regression tests (`test_common_dimensions_sanitizes_explicit_tenant_id_kwarg`, `test_get_cached_tenant_id_rejects_non_guid_legacy_string`) pin both fixes.

**Verification.** Fetched PR head `0a01196` into a detached worktree and ran `pytest tests/flightcheck/test_telemetry.py tests/flightcheck/test_graph_client.py tests/test_adk_telemetry.py tests/scripts/test_auth.py tests/flightcheck/test_cli.py` -> **118 passed in ~8.6s** (Python 3.11). Did not run linters/type-checkers, the full suite, or any live Graph call.

**Minor, non-blocking:** `requests` applies `timeout` to the connect and read phases separately, so a pathological connect-then-read stall can reach ~2x `_SILENT_LOOKUP_TIMEOUT` per call, making the rare both-lookups worst case somewhat above the "2 x 3s = 6s" stated in the comment. Not worth changing; flagging only so the stated ceiling isn't read as a hard total.

**Still open (owner, not code):** Question #2 (privacy sign-off that reading `/me?$select=companyName` is OII, not EUPI) is unchanged. That needs the data-profile owner's explicit confirmation or an ADO link; it's the one item I'd want closed before merge. Nits #5 (duplicated cache-fallback block across the two cli.py emit sites) and #6 (explicit `instance_id=""` now falls back to a fetched id) remain optional.

PR #242, opened 2026-08-18
adk-telemetry: fail-closed test-suite guard + tenant_id GUID sanitizer

amilandi:
## Summary

Two layers of defense so the test suite can never ship real Aria events to prod, and even a non-test code path can't stamp events with a bad `tenant_id`.

## Motivation

Investigation of the External dashboard's "Capability Usage by Type" gap surfaced a test-fixture leak polluting the prod Aria stream. In the past 30 days, tests that exercised `main()` end-to-end with a monkey-patched `discover_tenant` returning the fixture string `"tenant-id"` shipped:

- **391** `adk.api.call` events
- **88** `adk.capability.use` events (46 `backup_template_configs` + 42 `restore_template_configs`)
- **1** `adk.session.start` event

…all stamped `tenant_id="tenant-id"` across 23 fake instance_ids over 7 days, inflating the customer bucket on the External dashboard and creating the appearance of high customer usage of two capabilities that are essentially never invoked by real customers.

### Layer 1 (primary) — autouse `_disable_adk_telemetry` in `tests/conftest.py`

Sets `ESS_ADK_TELEMETRY=off` (+ `ESS_ADK_TELEMETRY_SYNC=1`) for every test. Every public emit short-circuits at `telemetry_enabled()` before touching the transport. Belt-and-braces sync-emit ensures any future accidental leak is inspectable in-process instead of racing off in a daemon thread past test teardown.

### Layer 2 (paranoia) — `_sanitize_tenant_id` in `adk_telemetry.py`

`set_identity` normalizes `tenant_id` through a canonical Entra GUID check (8-4-4-4-12 lowercase hex, whitespace + case tolerated). Anything else (placeholders, org display names accidentally routed here, truncated GUIDs) → `""`. The event still emits (fail-open telemetry, per SDK contract) but `classify_tenant` labels it `"unknown"` — it's excluded from the customer bucket instead of masquerading as one.

## Regression tests

- **`tests/test_conftest_telemetry_guard.py`** (new) — positive-observation module with no local `_isolate` fixture. Asserts the env vars are set, `telemetry_enabled()` is false, and every public emit path completes without invoking `_fc._post`.
- **`tests/test_adk_telemetry.py`** — new tests for `_sanitize_tenant_id` accept/reject/normalize behavior + the end-to-end effect that `set_identity(tenant_id="tenant-id")` yields empty in `common_dimensions` and `classify_tenant` returns `"unknown"`. Existing fixture strings (`"tenant-Z"`, `"tenant-Y"`) migrated to canonical GUIDs so they survive the sanitizer.

## Validation

- **209/209** telemetry-adjacent tests pass (`tests/test_adk_telemetry.py`, `tests/test_conftest_telemetry_guard.py`, `tests/scripts/test_auth.py`, `tests/scripts/test_discover.py`, `tests/flightcheck/test_cli.py`, `tests/test_push.py`).
- Full suite: 14 unrelated pre-existing installer/setup failures on `main` are untouched by this PR.

## Follow-ups (not in this PR)

- Dashboard-side redaction filter on the External tiles to also hide the historical 179 `tenant_id="tenant-id"` events already in the P30D window.

## Related

- Companion PR #237 (blank tenant identity fix for the `emit_capability.py` subprocess shim).

### DIFF
--- a/solutions/ess-maker-skills/scripts/adk_telemetry.py
+++ b/solutions/ess-maker-skills/scripts/adk_telemetry.py
@@ -218,6 +218,13 @@ def set_identity(
     )
     _IDENTITY["tenant_id"] = tenant_id or ""
     _IDENTITY["tenant_name"] = tenant_name or ""
+    # Persist tenant_id to disk so a *later* Python subprocess (e.g. the
+    # emit_capability.py shim launched from a SKILL.md step) — which never
+    # calls set_identity itself — can still stamp the tenant on its events.
+    # Without this, subprocess-emitted events land with an empty tenant_id,
+    # classify as "unknown", and never show on the customer-filtered dashboard.
+    if _IDENTITY["tenant_id"]:
+        _fc.cache_tenant_id(_IDENTITY["tenant_id"])
     # When a Graph-capable flow (FlightCheck) resolves the org display name,
     # persist it so ADK events emitted later in a *different* process (which
     # only has a Dataverse/BAP token and can't resolve it) can reuse it. The
@@ -373,6 +380,14 @@ def common_dimensions(
 ) -> dict[str, Any]:
     """Build the dimensions present on every event (spec Common Dimensions)."""
     tid = _IDENTITY["tenant_id"] if tenant_id is None else tenant_id
+    # Fall back to the disk-persisted tenant_id when this process never
+    # called set_identity (subprocess launched by emit_capability.py from a
+    # SKILL.md step). Without this, capability events emitted by the shim
+    # carry an empty tenant_id, classify as "unknown" via classify_tenant(""),
+    # and never appear on the customer-filtered External dashboard even
+    # though the maker's earlier auth flow knew the tenant.
+    if not tid and tenant_id is None:
+        tid = _fc.get_cached_tenant_id()
     tname = _IDENTITY["tenant_name"] if tenant_name is None else tenant_name
     # Fall back to the org display name a prior Graph-capable run (FlightCheck)
     # cached for THIS tenant, so pure-ADK events (session/build/deploy/
--- a/solutions/ess-maker-skills/scripts/adk_telemetry.py
+++ b/solutions/ess-maker-skills/scripts/adk_telemetry.py
@@ -198,8 +198,38 @@ _SESSION_LOCK = threading.Lock()
 
 
 # --- Identity model (spec) ------------------------------------------------
+# Canonical Entra tenant GUID: 8-4-4-4-12 lowercase-hex, dashes at fixed
+# offsets. Enforced *only* on tenant_id (the raw OII we ship to Aria) so
+# obvious test-fixture placeholders like ``"tenant-id"`` don't slip into
+# the prod stream and inflate the customer bucket. Also applied when
+# reading back from the on-disk ``.tenant_id`` cache, so a torn / legacy /
+# hand-edited file can never bypass the check.
+_GUID_RE = re.compile(
+    r"^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$"
+)
+
+
+def _sanitize_tenant_id(tenant_id: str) -> str:
+    """Return ``tenant_id`` iff it looks like a canonical GUID, else ``""``.
+
+    Never raises. Empty input stays empty. Whitespace/case are tolerated so
+    a caller that passes ``"  ABCDEF01-2345-6789-ABCD-EF0123456789  "`` still
+    gets a valid identity (normalized to lowercase). Anything else
+    (``"tenant-id"``, ``"unknown"``, truncated GUIDs, org display names
+    accidentally routed here) normalizes to ``""`` so downstream
+    ``classify_tenant`` labels the event as ``"unknown"`` rather than as
+    a fake customer.
+    """
+    if not tenant_id:
+        return ""
+    v = tenant_id.strip().lower()
+    return v if _GUID_RE.match(v) else ""
+
+
 def set_identity(
-    tenant_id: str = "", instance_id: str | None = None, tenant_name: str = ""
+    tenant_id: str | None = None,
+    instance_id: str | None = None,
+    tenant_name: str | None = None,
 ) -> dict[str, str]:
     """Record the install + tenant identity for all subsequent events.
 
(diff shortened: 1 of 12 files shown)


solutions/ess-maker-skills/scripts/adk_telemetry.py
```diff
@@ -373,6 +438,16 @@ def common_dimensions(
 ) -> dict[str, Any]:
     """Build the dimensions present on every event (spec Common Dimensions)."""
     tid = _IDENTITY["tenant_id"] if tenant_id is None else tenant_id
```
daeunJe0ng: [copilot-review] The GUID sanitizer is applied to the cache fallback (just below) and in `set_identity`, but an explicitly passed `tenant_id=` kwarg is emitted raw here (`... else tenant_id`). Since `classify_tenant` labels any non-empty, non-internal value as `customer`, a non-GUID passed explicitly to `common_dimensions` / `emit_*` still lands in the customer bucket, which is the exact failure this layer is meant to prevent.

The deeper point: sanitize once at the single choke point where all `tenant_id` sources converge, rather than per-source. Suggest sanitizing the explicit branch too:
```python
tid = _sanitize_tenant_id(tenant_id) if tenant_id is not None else _IDENTITY["tenant_id"]
```
`""` still maps to `""` (preserves the "explicit empty stays empty" tests) and a real GUID is unchanged. There is currently no test for a non-GUID explicit kwarg; worth adding if you take this.

daeunJe0ng (commented):
[copilot-review]

**Verdict: Approve with nits.** Well-motivated, data-backed telemetry data-quality fix; root cause verified against prod (391 `adk.api.call` + 88 `adk.capability.use` events stamped `tenant_id="tenant-id"`). No blocking defects. The two-layer approach is the right shape: fail-closed autouse conftest guard as the primary test-boundary fix, GUID sanitizer as defense-in-depth. Inline comments below are non-blocking.

**Design note (non-blocking, for a possible follow-up):**
- The emit control could be inverted to opt-in / fail-closed at the sink: emit only when an explicit prod signal (e.g. an env flag set by the prod deployment) is present, with the autouse conftest guard kept as Layer 2. An env gate at the emit boundary protects every path (ad-hoc scripts, notebooks, a new test file that bypasses `conftest.py`), whereas the autouse fixture only protects processes that load `conftest.py`.
- `classify_tenant`'s `non-internal => customer` default is optimistic. Consider mapping values that fail GUID validation to `unknown` instead of `customer`, as a second safety net for the External dashboard even when a bad value slips upstream.

**Question:** the 391 + 88 events already in prod aren't corrected by this forward fix. Is there a plan to exclude known-placeholder / non-GUID `tenant_id`s from the External dashboard query (or a backfill filter) so the customer bucket is accurate now, not just going forward?

**Verification:** reviewed the full 5-commit diff and head-of-branch source; confirmed statically that `import re` is present, `_GUID_RE` matches the migrated fixtures, and `classify_tenant` semantics make the first inline comment material. Did not run the suite (no local checkout).

solutions/ess-maker-skills/scripts/flightcheck/telemetry.py
```diff
@@ -266,18 +270,131 @@ def cache_tenant_name(
...
+            and _rank.get(existing.get("source", "organization"), 0) > new_rank
+        ):
+            return
+    except (OSError, ValueError):
+        pass
     try:
         os.makedirs(local_dir, exist_ok=True)
         with open(path, "w", encoding="utf-8") as f:
-            json.dump({"tenant_id": tenant_id, "tenant_name": tenant_name}, f)
+            json.dump(
```
daeunJe0ng: [copilot-review] `cache_tenant_name` still writes in place (`open(path, "w")` + `json.dump`), while `cache_tenant_id` in this same PR was made atomic (`tempfile` + `os.replace`). Both files are read cross-process by the subprocess / SKILL.md scenario this PR targets, so a torn read of `.tenant_name` yields a blank label. Lower impact than `tenant_id` (blank name vs misclassification), but suggest applying the same atomic temp-file + `os.replace` pattern here for consistency.

solutions/ess-maker-skills/scripts/flightcheck/telemetry.py
```diff
@@ -266,18 +270,131 @@ def cache_tenant_name(
...
+            obj = json.loads(raw)
+        except ValueError:
+            return ""
+        if not isinstance(obj, dict) or obj.get("version") != 1:
+            return ""
+        return str(obj.get("tenant_id", "")).strip()
+    # Legacy raw-string format from pre-versioned ADK builds.
+    return raw
+
+
```
daeunJe0ng, line 444: [copilot-review] The docstring above says the legacy raw-string path recognizes "a single line whose contents look like a GUID" and discards "any other content (truncated GUID ...)", but this branch just does `return raw` with no GUID validation. It is safe today only because the sole caller re-sanitizes, but `get_cached_tenant_id` is public / importable. Suggest either validating `raw` against `_GUID_RE` before returning (to match the docstring) or softening the docstring so claim and code agree.

solutions/ess-maker-skills/scripts/adk_telemetry.py
amilandi: Addressed in [`0a01196`](https://github.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/commit/0a01196) on PR #237. Sanitizer now applies to every `tenant_id` source (explicit kwarg, in-memory identity, disk cache) at the single choke point in `common_dimensions`. New regression test `test_common_dimensions_sanitizes_explicit_tenant_id_kwarg` locks it down.

solutions/ess-maker-skills/scripts/flightcheck/telemetry.py
amilandi: Addressed in [`0a01196`](https://github.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/commit/0a01196) on PR #237. `cache_tenant_name` now uses tempfile + `os.replace` for atomicity, matching `cache_tenant_id`.

solutions/ess-maker-skills/scripts/flightcheck/telemetry.py
amilandi: Addressed in [`0a01196`](https://github.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/commit/0a01196) on PR #237. The legacy raw-string branch now validates against `_GUID_RE` before returning (case-insensitive) and normalizes to lowercase. Docstring and code now agree; regression test `test_get_cached_tenant_id_rejects_non_guid_legacy_string` added.

amilandi:
Design note addressed in [`83ab94e`](https://github.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/commit/83ab94e): `classify_tenant` now maps any non-empty, non-internal, non-GUID value to `unknown` instead of `customer`. This is Layer 3 of the defense against fixture / placeholder / garbage `tenant_id` values reaching the External dashboard — on top of the autouse `_disable_adk_telemetry` fixture (Layer 1) and `_sanitize_tenant_id` at `set_identity` ingress (Layer 2). Regression tests added for the non-GUID → `unknown` mapping and for the lock-step between `adk_telemetry._GUID_RE` and `flightcheck.telemetry._GUID_RE`.

On your **question** about the 391 + 88 already-in-prod fixture events: this PR is intentionally forward-only. For the historical data I've been filtering the External Aria dashboards with `tenant_id NOT IN (<known bad values>)` on a per-tile basis, but a durable fix (dashboard-level query filter or an OneCollector processor rule) is worth doing separately. I'll open a follow-up work item to track that.

The **fail-closed-at-sink design alternative** (env-gated emit boundary as Layer 1, autouse fixture as Layer 2) is a cleaner shape and I agree with the direction — I'll spin it out as a separate PR so this one stays focused on the fixture-leak-specific fix. The current autouse guard is sufficient for the "test suite pollutes prod" hazard we hit; the env-gate is the right defense against ad-hoc scripts / notebooks that don't load `conftest.py`.

The inline nits on PR #237 (single-choke-point sanitizer, atomic `cache_tenant_name`, `_GUID_RE` validation on legacy raw-string) are addressed in [`0a01196`](https://github.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/commit/0a01196) on that PR.

daeunJe0ng (approved):
[copilot-review] Re-review at `83ab94e` — Approve.

Thanks for the thorough turnaround. Confirmed each item is addressed in code and covered by a regression test (verified by reading the source at head, not just the replies):

- **Explicit `tenant_id=` kwarg bypass (was Should-fix):** `common_dimensions` now routes every source through `_sanitize_tenant_id` at a single choke point. `test_common_dimensions_sanitizes_explicit_tenant_id_kwarg` pins `"tenant-id" -> "" / unknown` and GUID-preserve-lowercase. Resolved.
- **`cache_tenant_name` atomicity (was Should-fix):** now `tempfile.mkstemp` + `os.replace`, matching `cache_tenant_id`. Resolved.
- **`get_cached_tenant_id` legacy path (was Should-fix):** legacy raw-string branch now validates against `_GUID_RE`; `test_get_cached_tenant_id_rejects_non_guid_legacy_string` covers it. Docstring and code agree. Resolved.
- **Classifier default (design note):** the new Layer-3 `classify_tenant` non-GUID -> `unknown` mapping plus the `_GUID_RE` lock-step test are a good defense-in-depth addition.
- **Historical 391 + 88 events / fail-closed-at-sink:** both reasonably deferred to follow-ups; forward-only scope for this PR is fine.

Also noted the `_SILENT_LOOKUP_TIMEOUT` (3s, no-retry `requests.get`) change on the auth hot path — sensible tradeoff, and `tenant_class` still comes from the token `tid` claim so a timeout only degrades `tenant_name`, never classification.

Not run: I did not execute the suite locally, so the above is static verification of code + test presence/correctness. LGTM once CI is green.

PR #245, opened 2026-08-20
planner: /planner skill with WeveNova MCP + role management

jainharsheet77:
## Summary

Adds the **`/planner` skill** to the ESS Maker Kit and integrates it with the **WeveNova plan MCP server** (`weve-open-mcp` 3.1.0), including full **role management** (role registry, plan-scoped attestation, role-aware/caller task filtering, write-time role validation).

The planner authors a structured **Plan** for an ESS rollout — grounded on Microsoft Learn, interviewed for the rest — as atomic Tasks each owned by a **role** and a **person**, with a ledger of what each Task produced. The plan persists locally **or in WeveNova** via `--store mcp`.

## What's in this PR

- **Planner skill** (`src/skills/planner/*`): research → interview → model → assign → evaluate(preview) → capture phases, plus Flow 2 ("what am I assigned?").
- **WeveNova MCP backend** (`scripts/planner/plan_store.py`, `mcp_client.py`): reads/writes the WeveNova project plan; bound to the multi-plan 3.x tool surface (project/plan/tenant ids); threads the required `If-Match` **etag** through update/set-state/delete and re-reads a fresh etag between chained mutations.
- **Role management**:
  - `roles.py` — WeveNova role registry (internal authority + attestable External/Entra/PowerPlatform roles). Ids are emitted **verbatim** (ordinal, case-sensitive; no slugify/lowercase) to match the backend exactly.
  - `weve_mapping.py` — `AssignedToType`/`AssignedToId`/`AssignedToRoleId` task shape.
  - `attest.py` — `AttestationClient` (attest / list / get / revoke / caller-tasks) with local validation mirroring the server's `ValidateAttestationRequest`.
  - `cli.py` — `roles`, `attest`, `assignments`, `revoke`, `caller-tasks` commands + global `--project-id`/`--plan-id`.
- **Docs** (`SKILL/model/assign/mytasks/interview`): teach verbatim role ids, attestation (Phase 4) and server-resolved Flow 2.

## Role wire format (critical)

Role ids are matched ordinally and case-sensitively by the backend with **no normalization** — so the skill emits them exactly:

- Internal authority: `AgentOwner`, `AgentEditor`, `AgentAnnotator`, `AgentViewer`
- External (compact id, not display): `WorkdayAdmin`, `ServiceNowAdmin`, `ServiceNowKnowledgeManager`
- Entra / PowerPlatform (id == display): `Global Administrator`, `Network Administrator`, `User Administrator`, `Power Platform Administrator`, `Environment Maker`

## Testing

`python -m pytest tests/planner -q` → **150 passed, 2 skipped** (the 2 are live-network tests). Store tests use an etag-enforcing fake client to validate the If-Match wiring. No live-server mutations were performed.


amilandi:
## Telemetry gap: `/planner` doesn't emit ADK capability telemetry

Reviewed the branch for telemetry coverage. The new skill will **not** appear on the ADK Capability Usage donut (or any ADK cube). Every other maker-facing skill in `solutions/ess-maker-skills/src/skills/` either emits `capability.use` in-process (setup, connect, discover, evaluate, backup/restore template configs, push, publish) or via a `python scripts/emit_capability.py <cap>` step in its `SKILL.md` (topics/{create,update,delete,review,test}, workflows/{create,update,delete,test}, cleanup, troubleshoot, onboarding, evaluations/{create,update,delete}). `/planner` does neither — I grepped the whole feature branch for `emit_capability`, `adk_telemetry`, and `adk_capability` and found zero hits across all 23 planner source files and 12 test files.

Consequence: when this merges as-is, every `/planner` run will be silently invisible on the dashboards used to track ADK adoption and capability usage. It won't even land in the `unknown` bucket — the emit call has to happen for that.

### Recommended fix

Match the `/discover` pattern from PR #238 (Apurva's `TestCapabilityTelemetry` class is the model). It's a one-place, in-process emit right after argparse succeeds:

1. **Add `"planner"` to `ADK_CAPABILITIES`** in `solutions/ess-maker-skills/scripts/adk_telemetry.py` (and to the comment header block). Note that PR #259 already has this pending — you may want to coordinate with that PR or with @amilandi to avoid a merge conflict on `ADK_CAPABILITIES`.

2. **Emit once per invocation in `scripts/planner/cli.py::main`**, right after `parser.parse_args(argv)`:

   ```python
   def main(argv: list[str] | None = None) -> int:
       _configure_io()
       parser = build_parser()
       args = parser.parse_args(argv)

       # Anonymous capability telemetry, emitted here — right after argparse
       # succeeds, so `--help` and bad arguments don't count — rather than at
       # the end. The "Capability Usage by Type" donut measures that a maker
       # *used* /planner, not whether the command happened to succeed; gating
       # the emit on a clean exit would silently undercount precisely the
       # runs worth looking into.
       try:
           import adk_telemetry
           # block=True: short-lived CLI process, emit synchronously so the
           # event isn't dropped when the interpreter exits and kills a
           # daemon thread.
           adk_telemetry.emit_capability_use("planner", block=True)
       except Exception:  # noqa: BLE001 — telemetry must never break the flow
           pass

       try:
           return args.func(args)
       # ...
   ```

3. **Do NOT also add a `emit_capability.py planner` step to `SKILL.md`.** `emit_capability_use()` doesn't dedupe, so wiring both would double-count every run. This is called out in Apurva's SKILL.md guardrail for `/discover` — same reasoning applies here. `scripts/planner/cli.py` is always the entry point, so the in-process emit is sufficient.

4. **Add a test file `tests/planner/test_cli_telemetry.py`** modeled on Apurva's `TestCapabilityTelemetry` from `tests/scripts/test_discover_inventory.py`:
   - `test_planner_is_canonical_so_it_cannot_bucket_to_unknown` — asserts `"planner" in ADK_CAPABILITIES` and `normalize_capability("planner") == "planner"`.
   - `test_a_run_emits_the_capability_exactly_once` — spies `emit_capability_use`, runs `cli.main(...)`, asserts exactly one `("planner", ...)` call.
   - `test_the_emit_is_synchronous` — asserts `block=True`.
   - `test_bad_arguments_do_not_count_as_usage` — `main(["nonexistent-cmd"])` raises `SystemExit`; assert no emit.
   - `test_a_telemetry_failure_never_breaks_the_flow` — monkeypatch `emit_capability_use` to raise; assert `main(...)` still returns 0.

### Note on subcommand granularity

`/planner` has many subcommands (init, add-scenario, add-task, assign, claim, capture-setup, pin-output, research, summary, validate, sync, etc.). Emitting a single `"planner"` capability for all of them is the right first cut — it matches how the rest of the taxonomy works today (one value per maker-facing skill/command, not per sub-action). If PMs later want finer breakdown (e.g., "how many plans got past `add-scenario`?"), that's an additive change: add `adk_capability` as an already-present dimension and stamp the subcommand into a new `planner_subcommand` dimension on a follow-up telemetry event, without breaking the donut.

Happy to help review the fix.

PR #247, opened 2026-08-24
feat(ess-maker-skills): add /harden instruction review skill

amilandi:
ADO: https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7774542

Adds a `/harden` capability to the ESS Maker Kit: a review of an agent's **system instructions** (the `instructions:` block in `agent.mcs.yml`) targeting the failures where an agent answers confidently from something other than its knowledge sources, or offers actions nothing authorized.

Draft for review. The flow has been executed end to end twice against a real extracted agent with a scripted maker on the other side — once with a specific complaint, once without. See the validation comment below.

## Why this, and why it looks the way it does

**Hardening can regress an agent.** An agent that declines questions its sources fully answer is also a failure, and it is the one a hardening pass is most likely to introduce. So the skill asks what the maker has actually observed *before* analyzing, and when nothing specific is wrong it reports findings and proposes only contradiction fixes rather than tightening pre-emptively. Part 4 of the rule pack is dedicated to catching over-restriction in the skill's own proposals.

The contradiction pass runs **regardless** of what the maker reports. A rule contradicted elsewhere is not in force however firmly it is written, and the author usually has no idea.

## What is here

| File | Purpose |
|---|---|
| `.github/prompts/harden.prompt.md` | Setup-state gate + delegation, mirroring `review.prompt.md` |
| `src/skills/instructions/harden/SKILL.md` | The ten-step flow |
| `src/reference/ess-docs/hardening/instruction-rules.md` | Generic rule pack — contradictions, grounding gaps, over-commitment, over-restriction, rewriting principles |
| `scripts/check_instruction_budget.py` | Deterministic character-budget probe |
| `scripts/list_agent_capabilities.py` | Deterministic inventory of the agent's topics, workflows and acting steps |
| `tests/scripts/` | 29 tests across the two probes and the prompt/skill wiring |

Plus registration: `ADK_CAPABILITIES`, the `wired` set in `test_adk_telemetry.py`, `/menu`, and the README.

**The budget probe** exists because hardening usually lengthens instructions, and instructions that exceed the ceiling can be silently truncated — which can drop the guardrail the pass just added. Asking the model to count characters was not reliable, so the skill is told the probe's verdict outranks its own estimate. Its error paths report `unknown` rather than a falsely clean measurement.

**No changes to the push path were needed.** `agent.mcs.yml` is already collected by `collect_files()`, classified as `botcomponent` by `classify_path()`, mapped to componenttype 15 in `.component-map.json`, and covered by the existing checkpoint/rollback.

## Review notes

- **The live push is unverified.** `agent.mcs.yml` is pushable by inspection of the code path, but whether a type-15 update takes effect against Copilot Studio has not been tested against a real agent. Worth validating before this is announced.
- **The 8000-character limit is a working assumption**, not a verified platform constant. It is the default and is overridable with `--limit` so a maker who knows their real ceiling is not blocked by ours.
- **The rule pack's claims about the shipped ESS default were checked against the default itself**, not assumed. The variant I read grounds well (it forbids using general knowledge outright) but says nothing about offering unsupported actions, external referrals, or jurisdiction scope, and it applies standing helpfulness pressure. The skill tells the maker when a finding is inherited rather than something they wrote, but only on an exact marker match — template text varies.
- **Telemetry follow-up:** `harden` needs adding to the Aria cube dimension value list per the note in `adk_telemetry.py` (ADO #7532631). I cannot do that from here.
- The suite has 14 pre-existing failures on `main` (installer and foundation-setup tests), unrelated to this change. Everything else passes.


amilandi:
## Validation

I could not drive Copilot Chat directly, so I executed the skill end to end twice against a real extracted agent (`employee-self-service-preview`: 36 topics, 12 workflows), with a scripted maker on the other side. Both runs were asked for a blunt defect report rather than a verdict.

**Branch A — maker reports two specific bad responses.** The skill traced both to the instructions, proposed a change, and the proposal prevented both behaviours without prohibiting anything the agent is actually configured to do.

**Branch B — maker reports nothing specific.** The skill checked for contradictions, found none worth reporting, and concluded the agent looks fine, proposing no changes. This is the design's central claim: with no complaint, `/harden` must not manufacture findings to justify itself. Hardening on a healthy agent costs real capability.

The two runs surfaced 15 defects, all fixed in 95e50be. The ones worth calling out:

- **Capability checks had no source.** `read the agent's tool inventory` pointed at a path that does not exist, and reading 36 topic files is not practical, so every capability claim was guesswork. Fixed with `list_agent_capabilities.py`, a probe that reports topics, workflows and acting steps. It discloses topics it could not read, because the caller is about to write prohibitions from that picture and silence would be worse than the gap.
- **Copilot Studio topic YAML is not valid YAML.** It contains unquoted `@type` keys, and `yaml.safe_load` raises on exactly the Workday/ServiceNow integration topics whose descriptions matter most. The probe uses a line-based key extractor instead. Covered by `test_description_survives_non_yaml_topic`.
- **`--agent` rejected the documented invocation.** `config.json` stores `agent.folder` as a path, and the skill prepended a directory to it, producing a doubled path.
- **Two rules fought each other.** `quote the exact line` versus `an absence is a valid finding` made models either fabricate a guilty line or suppress real findings. Findings are now anchored at sentence level, and a missing safeguard is explicitly reportable without a quote.
- **No contradiction threshold**, so branch B classified ordinary tone-versus-grounding guidance as a contradiction in order to have something to report. Reporting one now requires naming a request where both rules cannot be satisfied.
- **The rule pack described a template that does not exist.** I checked its claims against the instructions the template actually ships; two of four asserted defect classes were absent. Provenance is now only claimable on an exact marker match.

Still unverified, and flagged in the description: whether pushing an `agent.mcs.yml` instruction change takes effect in Copilot Studio, and whether 8,000 characters is the real ceiling (hence `--limit`).

amilandi:
Three fixes from running the skill against a live agent (0025160).

**The intake question got answered with a menu.** Copilot offered the maker a list — bad Q&A, general concern, or just check contradictions — and picking `general concern` ended the interview. A picked category cannot be anchored to a change, so the run went straight to drafting generic hardening. Step 2 now forbids offering options while eliciting, and requires a prose follow-up when the answer names a category rather than a behaviour.

**Runs ended on the diff.** Step 10 was written as though the apply path were the only path, and Step 8 told the model to `stop` on the others, so nothing pointed the maker at `/test` or `/evaluate`. Step 10 now runs on every path that reached the analysis, with separate wording for applied, declined, and nothing-proposed. The Step 1 gates are the only exception.

**Contradictions were only checked against the maker's existing text.** Hardening adds prohibitions to a document that already has rules, which is exactly how contradictions get created — a new `never offer a next step` sits perfectly well next to an existing `always end by offering further help`, and the skill would have shipped it. Step 6 now re-runs the contradiction pass against the whole candidate, and when the new text collides with a rule that stays behind, that rule has to be amended in the same proposal. Stacking a stricter rule on top and relying on it winning is the failure this skill exists to catch.

A re-run confirms all three: prose question, follow-up on the vague answer, handoff to `/test` and `/evaluate`. 99 tests, including regression tests for each.

johnguy0:
We should have evals to show:
- the skill is working as expected
- the skill works better than not having the skill
- regression prevention

johnguy0:
Are there guardrails in place for any parts of our instruction that we consider immutable?

PR #248, opened 2026-08-25
adk-telemetry: app-only report_client_events bridge (P5-1-ADK)

deepaksharmacse12:
Implements `P5-1-ADK` from [`60-phase-5.md`](https://dev.azure.com/O365Exchange/Weve_DA/_git/Jabberwocky?path=/Client/Vorpal/Docs/Plans/vorpal-implementation-plan/60-phase-5.md&_a=preview) — the ADK side of the short-term Vorpal telemetry bridge.

Vorpal owns the event contract and batches widget telemetry through an app-only MCP tool. ADK validates the envelope and emits accepted events through its **existing** anonymized 1DS path, so opt-out, buffering, tenant context, and the approved Data Profile all continue to apply unchanged. No new transport, no Aria logger, no residency bootstrap.

## Contract

Matches `Client/Vorpal/Telemetry/types.ts` at `schemaVersion: 1`:

```
{ schemaVersion, correlationId, mountId, appName,
  buildEnvironment, buildNumber, toolCallId?, events }
```

The tool is registered with `_meta.ui.visibility: ["app"]`, so the host keeps it out of the model-visible tool list. The result is always an explicit accepted/rejected verdict carrying no customer content.

## Two decisions worth reviewer attention

**The MCP tool parameters are deliberately untyped.** FastMCP validates the tool signature with Pydantic *before* the body runs, so a narrowly-typed signature turns an out-of-contract envelope into a `ToolError` with no `structuredContent`. Vorpal reads a result without a `status` as a transient `invalid_result` and may retry it — meaning a permanently malformed batch would retry forever. Widening the signature routes every envelope through the bridge validator instead. Verified through the real `call_tool` path:

| envelope | result |
|---|---|
| `events: [1, 2]` | `rejected` / `invalid_event_shape` |
| `correlationId: 123` | `rejected` / `invalid_correlation_id` |
| `mountId: 123` | `rejected` / `invalid_mount_id` |
| `toolCallId: 123` | `rejected` / `invalid_tool_call_id` |
| `schemaVersion: "2"` | `rejected` / `unsupported_schema_version` |
| `buildNumber` missing | `rejected` / `invalid_event_shape` |
| `events` missing | `rejected` / `empty_batch` |

**An internal ADK fault resolves to `accepted` for the whole batch.** Vorpal rejects an acknowledgement that doesn't account for every sent event, so a partial or zero count would also drive a retry loop. ADK's own buffer already owns delivery retries, which matches the module's fail-open rule ("never break a caller"). The same reasoning applies when the maker has opted out: nothing is transmitted, but the batch is still acknowledged in full. The echoed count is clamped to `CLIENT_EVENTS_MAX_BATCH_EVENTS`, since this path runs when validation itself faulted and the batch cap has not necessarily been applied.

## Privacy: no free-text tunnel

Every field that reaches Aria is constrained, by one of two strategies:

| Field | Strategy |
|---|---|
| `correlationId`, `mountId`, `toolCallId` | **Rejected** unless identifier-shaped: prefix (`corr-`/`mount-`/`tool-`) + `^[A-Za-z0-9._-]{1,64}$` |
| `eventName` | **Rejected** unless identifier-shaped: `^[A-Za-z][A-Za-z0-9_.-]{0,63}$` |
| Property keys | **Rejected** unless `^[A-Za-z][A-Za-z0-9_]{0,63}$` |
| Property values, `appName`, `buildEnvironment`, `buildNumber`, `locale` | **Scrubbed** via the module's existing `_scrub()` |

Scrubbing turns customer content into placeholders:

```
"C:\Users\jdoe\secret.docx see https://contoso.sharepoint.com/x"  ->  "<path> see <url>"
"jdoe@contoso.com"                                                ->  "<email>"
"6f7c8f9c-1234-4abc-9def-0123456789ab"                            ->  "<guid>"
```

The correlation identifiers are **rejected rather than scrubbed** on purpose: a redacted correlation id would still be accepted but would silently stop stitching a mount's events together, which is worse than refusing the batch outright.

ADK keeps **no** Vorpal event-name allowlist — Vorpal owns and versions the contract, so new bounded events don't need an ADK release. Property values are limited to bounded primitives and arrays of primitives, and unknown envelope/event fields are rejected. No raw 1E request IDs or user-linkable identifiers are emitted; tenant/time-window correlation continues to come from ADK's existing context.

## Rejection reasons — coordination note for the Vorpal side

Each identifier field reports its own reason, so a failure is attributed to the field that actually failed rather than all three sharing `invalid_correlation_id`:

```
correlationId -> invalid_correlation_id
mountId       -> invalid_mount_id
toolCallId    -> invalid_tool_call_id
```

`invalid_mount_id` and `invalid_tool_call_id` **extend Vorpal's bounded `BridgeRejectedReason` union**, so until the matching `types.ts` / `TelemetryBridge.ts` change ships, `isRejectedReason` won't recognize them and `getTelemetryBridgeOutcome` maps them to `unrecognized_reason`:

| | before | this PR, pre-Vorpal | after Vorpal ships |
|---|---|---|---|
| bad `mountId` | `invalid_correlation_id` (wrong field) | `unrecognized_reason` (no field) | `invalid_mount_id` |

**No functional risk in the gap:** an unrecognized reason is still classified as a *rejection*, so the batch is dropped rather than retried — identical behavior to today. The transitional trade is an honest "unknown" in place of a confidently wrong attribution.

⚠️ If any Vorpal dashboard or alert already buckets on `invalid_correlation_id`, it will lose `mountId`/`toolCallId` traffic into `unrecognized_reason` until the union update lands. That change is self-contained in `fabf84f` and can be dropped from this PR so the two ship together — flag it in review if preferred.

## Session state

`get_session()` mutates `~/.adk/session.json` (it mints a fresh id once the inactivity window lapses and stamps `last` on every call), so it runs only after the whole batch passes validation — a rejected batch leaves no side effect. An internal fault *after* validation can still touch the session file before the fail-open acknowledgement.

Whether widget telemetry should use its own surface/session model instead of `SURFACE_CLI`, rather than extending the maker's CLI session window, is **tracked separately in ADO** and intentionally out of scope here.

## Testing

104 tests pass across `tests/test_adk_telemetry.py` and `tests/test_adk_mcp_report_client_events.py`, covering the happy path, opt-out, every rejection reason, scrubbing, identifier charset gating (including a positive case proving the gate still accepts the shapes Vorpal mints), per-field rejection reasons and the bounded value-list guard, non-ASCII event names, oversized property strings, non-finite numbers, fail-open and its clamped echo, app-only metadata, and the malformed-envelope regression through the real `call_tool` path.

Full suite: 1984 passed. The 10 failures in `tests/scripts/test_installer_*.py` are pre-existing and unrelated — confirmed identical on the base commit with these changes stashed.

## Review follow-ups

- ✅ `correlationId`/`mountId`/`toolCallId` free-text tunnel — fixed in `28b3498`.
- ✅ Unbounded fail-open echo — clamped in both the bridge and the MCP wrapper.
- ✅ Test gaps for identifier content, non-ASCII event names, and oversized property strings.
- ✅ Per-field rejection reasons — `fabf84f`; see the coordination note above.
- ⏭️ `_THREADS` thread-safety is pre-existing (predates this PR; `_emit_many` mirrors the existing `_emit` pattern) and is better fixed across all async emitters separately.

## Notes

- Based on `feature/landing-page-config` (commit `3dda156`), not `main`.
- Opened from a fork: my corporate account is an Enterprise Managed User and cannot fork or be granted access to the public `microsoft` org. Commits are authored as `Deepak Sharma <dsharm@microsoft.com>`.
- Pairs with `P5-1-JW` on the Vorpal side. No Vorpal change is required for this PR to merge; the `BridgeRejectedReason` union update is a follow-up, not a blocker.


solutions/ess-maker-skills/scripts/adk_telemetry.py
```diff
@@ -534,6 +586,249 @@ def _emit(event_name: str, data: dict[str, Any], *, block: bool = False) -> dict
...
+    # only after the whole batch has passed validation — a rejected batch must
+    # leave no side effect on ADK session state.
+    sid, _ = get_session(SURFACE_CLI)
+    for event in events:
+        locale = event.get("locale")
+        row = common_dimensions(SURFACE_CLI, session_id=sid)
+        row.update(
+            {
+                "client_event_name": event["eventName"],
+                "client_correlation_id": envelope["correlationId"],
```
amilandi, line 901: **🔴 High — free-text tunnel via `correlationId` / `mountId` / `toolCallId`**

The PR description says *"ADK still refuses to act as a free-text tunnel: event names must be identifier-shaped… every emitted string goes through `_scrub()`"*. That's enforced for `eventName` (regex-guarded), property keys/values (regex + `_scrub_client_scalar`), and `appName`/`buildEnvironment`/`buildNumber`/`locale` (scrubbed at 782-784, 787). But `client_correlation_id` (this line), `client_mount_id` (781), and `client_tool_call_id` (786) are written **verbatim** — no `_scrub_client_scalar` call — because their validator is only `_valid_client_identifier` (line 657):

```python
def _valid_client_identifier(value, prefix):
    return _valid_bounded_string(value) and value.startswith(prefix)
```

`_valid_bounded_string` requires non-empty `str` ≤ 200 chars. **No character class check.** So this envelope validates cleanly:

```json
{ "correlationId": "corr-jdoe@contoso.com C:\\Users\\jdoe\\secret.docx https://contoso.sharepoint.com/x",
  "mountId":      "mount-<any 195 chars of arbitrary UTF-8>",
  "toolCallId":   "tool-<any 195 chars of arbitrary UTF-8>" }
```

Each field gives a buggy widget (or compromised app-only caller) ~195 characters of unrestricted UTF-8 to smuggle into Aria — UPNs, paths, URLs, GUIDs, object IDs. `_fc._post` is a passthrough, so the raw string reaches 1DS.

The test `test_report_client_events_scrubs_paths_urls_emails_and_guids` only exercises property values; the only negative test for `correlationId` uses `"request-123"`, which happens to fail the prefix check. That's why the gap wasn't caught.

**Fix — pick one:**
- **(a) Tighten** `_valid_client_identifier` to require an identifier-shaped suffix, e.g. `^(corr|mount|tool)-[A-Za-z0-9._-]{1,64}$`. Matches the PR's stated posture and rejects the smuggling attempt outright.
- **(b) Scrub** all three fields with `_scrub_client_scalar` the same way `appName`/`buildEnvironment`/`buildNumber` are, at 780/781/786.

I'd take (a) — it matches the "refuses to act as a free-text tunnel" phrasing and rejects rather than silently redacts. Please add a targeted test where `correlationId="corr-jdoe@contoso.com"` (or with a `<guid>`/path) is rejected.

solutions/ess-maker-skills/scripts/adk_telemetry.py
```diff
@@ -534,6 +586,249 @@ def _emit(event_name: str, data: dict[str, Any], *, block: bool = False) -> dict
...
+        return CLIENT_EVENTS_REJECTED_INVALID_CORRELATION_ID, []
+    if not _valid_bounded_string(envelope.get("appName")):
+        return CLIENT_EVENTS_REJECTED_INVALID_EVENT_SHAPE, []
+    if not _valid_bounded_string(envelope.get("buildEnvironment")):
+        return CLIENT_EVENTS_REJECTED_INVALID_EVENT_SHAPE, []
+    if not _valid_bounded_string(envelope.get("buildNumber"), allow_empty=True):
+        return CLIENT_EVENTS_REJECTED_INVALID_EVENT_SHAPE, []
+    tool_call_id = envelope.get("toolCallId")
+    if tool_call_id is not None and not _valid_client_identifier(tool_call_id, "tool-"):
+        return CLIENT_EVENTS_REJECTED_INVALID_CORRELATION_ID, []
```
amilandi: **🟡 Low — `INVALID_CORRELATION_ID` reason is reused for `mountId` and `toolCallId` failures**

Lines 734 and 743 both return `CLIENT_EVENTS_REJECTED_INVALID_CORRELATION_ID` when the failing field is actually `mountId` / `toolCallId`. Vorpal-side dashboards / alerts on `rejectedReason` will attribute those to the wrong field.

The value list is bounded by design so this is a spec/wire-contract decision more than a bug — but if `rejectedReason` will be surfaced anywhere for triage, consider either one reason per field (`invalid_mount_id`, `invalid_tool_call_id`) or a single unified `invalid_identifier`. Either is a coordinated change with the Vorpal side.

solutions/ess-maker-skills/scripts/adk_telemetry.py
```diff
@@ -534,6 +586,249 @@ def _emit(event_name: str, data: dict[str, Any], *, block: bool = False) -> dict
...
+    try:
+        reason, rows = _validate_client_events_envelope(envelope)
+        if reason:
+            return _rejection(reason)
+        _emit_many(EVENT_CLIENT, rows, block=block)
+        return {"status": "accepted", "acceptedEventCount": len(rows)}
+    except Exception:  # noqa: BLE001 — the bridge must never break the widget
+        return {
+            "status": "accepted",
+            "acceptedEventCount": _client_events_batch_size(envelope),
```
amilandi, line 942: **🟡 Low (defense-in-depth) — fail-open echoes an unbounded `len(events)`**

The batch-level fail-open echo relies on the pre-validation `len(events)`. `_client_events_batch_size` (below) reads it back off the envelope without applying the `CLIENT_EVENTS_MAX_BATCH_EVENTS` cap. Same pattern in `src/mcp/adk/server.py:573`.

Normally fine — the *validator* caps at 25 events and 64KB serialized. But `_meta.ui.visibility: ["app"]` is advisory (not a standard MCP-spec field the model necessarily honors), so if the host doesn't filter this tool from the model-visible list, a misbehaving caller could pass `events=[…100000 dicts…]` and get `acceptedEventCount: 100000` back — nothing is actually emitted, but the number Vorpal / the model sees is unbounded.

**Suggested fix:** cap the echoed count at the batch max, e.g. `min(_client_events_batch_size(envelope), CLIENT_EVENTS_MAX_BATCH_EVENTS)` here and `min(len(events), CLIENT_EVENTS_MAX_BATCH_EVENTS) if isinstance(events, list) else 0` in `server.py`. Vorpal's own batcher shouldn't exceed the cap anyway, so this is a no-op in the healthy case.

amilandi (commented):
Solid PR overall — the two headline design decisions (untyped kwargs so malformed envelopes get a `rejected/<reason>` verdict instead of a FastMCP `ToolError`, and batch-level fail-open `accepted`) both hold up under the real `call_tool` path and the tests exercise them well. Session-state deferral is also correctly implemented and regression-tested.

One **High** finding that directly contradicts the "no free-text tunnel" claim in the PR description, plus three minor items — inline below.

## What I verified
- **Untyped-kwargs path**: every malformed envelope shape I traced produces `structuredContent` with `status:"rejected"` + bounded `rejectedReason`; the parametric test exercises the real `mcp.call_tool` boundary (not a direct Python call), so the Pydantic-before-body concern is genuinely covered.
- **Fail-open batch-level ack**: both bridge and MCP wrapper trap all exceptions and return `accepted` with the pre-validation batch size; no stack-trace leakage; `_emit_many_sync` buffers on failure for the existing retry path.
- **Session side-effect**: `get_session()` is called strictly *after* every per-event validator return — regression test covers it. Minor: an internal fault *after* validation still touches the session file before the fail-open `accepted`, which is consistent with the rule but the PR description phrasing implies otherwise. Worth a one-line clarification.
- **Property-value type gate**: correctly rejects nested dicts, NaN/±inf, bools-as-numbers; arrays length-capped at 10 with per-item recheck. Nested-dict smuggling is not possible.
- **Envelope size cap**: 64KB serialized cap runs before per-event work; non-serializable envelope returns MAX+1 → `batch_too_large`.
- **`eventName` regex**: correctly rejects spaces/punctuation/emoji/oversize.

## Findings (details inline)
1. 🔴 **High** — `correlationId` / `mountId` / `toolCallId` are a free-text tunnel (no charset check, no `_scrub`). ~195 chars of arbitrary UTF-8 per field reach Aria unmodified.
2. 🟡 **Low** — `INVALID_CORRELATION_ID` reason is reused for `mountId` / `toolCallId` rejections.
3. 🟡 **Low (defense-in-depth)** — fail-open echoes an unbounded `len(events)` back to the caller when the outer try/except fires before size caps apply.
4. 🟡 **Low** — test coverage gaps for the specific attack shapes the PR summary claims to defend against.

Happy to help with follow-ups once you've decided on the direction for #1 (tighter regex vs. scrub).

tests/test_adk_telemetry.py
```diff
@@ -451,6 +483,213 @@ def test_buffer_oversize_drops_oldest_and_accepts_newest(monkeypatch):
...
+
+    assert result == {"status": "accepted", "acceptedEventCount": 2}
+    assert len(captured_post) == 1
+    _ikey, envelopes = captured_post[0]
+    assert [envelope["name"] for envelope in envelopes] == [
+        "adk.client.event",
+        "adk.client.event",
+    ]
+    first = envelopes[0]["data"]
+    assert first["client_event_name"] == "WidgetReady"
```
amilandi, line 638: **🟡 Low — test gaps for the specific attack shapes the PR summary defends against**

The tests cover schema/empty/oversized/property-type/event-name regex, and scrubbing of property *values*. Not covered:

- Free-text content in `correlationId` / `mountId` / `toolCallId` (the primary gap — see the High finding on `adk_telemetry.py`).
- Non-ASCII / emoji in `eventName` (regex correctly rejects; not asserted).
- Per-string-length rejection for a property string > `CLIENT_EVENTS_MAX_STRING_LENGTH` (validated but not tested).
- Concurrent `report_client_events` calls (`_THREADS[:] = ...` list mutation at `adk_telemetry.py:584` and `:644` is not thread-safe — pre-existing, but this PR increases traffic through the path).
- Opt-out flipping mid-batch, and buffered-replay of an `adk.client.event` envelope through the existing retry path.

Not blockers on their own; worth adding once the High finding is resolved.

solutions/ess-maker-skills/scripts/adk_telemetry.py
deepaksharmacse12: Confirmed and reproduced — thanks, this is a real gap, and my PR description overclaimed. `_valid_client_identifier` only checked the prefix and length, so ~195 chars of arbitrary UTF-8 rode through unscrubbed while identical content in `appName` was redacted.

Went with **(a)**. Added `_CLIENT_EVENTS_IDENTIFIER_SUFFIX_RE = ^[A-Za-z0-9._-]{1,64}$`, applied to the suffix after the `corr-`/`mount-`/`tool-` prefix.

One note on why (a) over (b) beyond matching the stated posture: scrubbing a correlation id would silently break event stitching for that mount — the batch would still be accepted, but the rows would no longer be correlatable. Failing loudly beats quietly-wrong telemetry.

Your exact payload now:

```
corr-jdoe@contoso.com C:\Users\jdoe\secret.docx https://contoso.sharepoint.com/x
  -> rejected / invalid_correlation_id   (0 rows emitted)
```

Tests added — 18 parametrized cases covering 6 free-text payloads (UPN, Windows path, URL, GUID-in-a-sentence, a bare space, 65 chars) × all 3 identifier fields. I also added a **positive** test, since an over-tight regex would silently drop *all* telemetry, which is the worse failure mode:

```
corr-6f7c8f9c-1234-4abc-9def-0123456789ab   -> accepted   (UUID suffix)
mount-1a2b_3c.4d  /  tool-Call.42           -> accepted   (dot/underscore/dash)
```

PR description corrected to describe the actual posture. Fixed in 28b3498.

solutions/ess-maker-skills/scripts/adk_telemetry.py
deepaksharmacse12: Taking the fix — cheap and correct, applied in both `adk_telemetry.py` and `server.py`.

One clarification for the record, since the title and the example differ slightly. On the **normal** path that batch is already bounded — the cap at the `len(events) > CLIENT_EVENTS_MAX_BATCH_EVENTS` check fires and returns `rejected/batch_too_large`. I could only reproduce `acceptedEventCount: 100000` by *also* forcing an exception inside validation (`_serialized_size` raising), which is exactly the "before size caps apply" case your title names. So it's genuinely defense-in-depth for the fault path rather than something reachable on its own.

After the fix, with validation faulting:

```
events=[…100000…]  ->  {"status": "accepted", "acceptedEventCount": 25}
```

I pinned both halves so the distinction can't silently regress: one test induces the validator fault and asserts the clamp, and a companion asserts the same oversized batch still returns `batch_too_large` on the healthy path. Same clamp and an equivalent test on the MCP wrapper. Fixed in 28b3498.

tests/test_adk_telemetry.py
deepaksharmacse12: Added the first three; deferring the last two with reasoning.

**Added** (+26 tests, 98 passing in the two bridge files):
- Free-text content in `correlationId` / `mountId` / `toolCallId` — 18 parametrized cases, plus a positive test that the gate still accepts the shapes Vorpal actually mints. You were right that the old `"request-123"` case passed for the wrong reason: it failed on the *prefix*, so it never exercised content at all.
- Non-ASCII / emoji in `eventName` — already rejected, now asserted.
- Property string > `CLIENT_EVENTS_MAX_STRING_LENGTH` — already rejected, now asserted, so the 64KB envelope cap isn't the only thing bounding a single value.

**Deferring, and why:**

*Concurrency.* Agreed `_THREADS[:] = …` / `.append` isn't thread-safe, but those sites (`:584-585`, `:848`) predate this PR and `_emit_many` only mirrors the existing `_emit` dispatch pattern — it doesn't add a new hazard. A correct fix means putting `_THREADS` behind a lock, which touches every async emitter in the module (`emit_capability_use`, `emit_api_call`, `flush`, …), well outside a telemetry-contract PR. Happy to file it separately; if the increased traffic through this path makes you want it landed together, say so and I'll do it here.

*Opt-out mid-batch and buffered replay.* Both exercise the pre-existing opt-out and buffer machinery rather than bridge logic — `adk.client.event` envelopes go through the same `_buffer_append` / `_buffer_flush` path as every other `adk.*` event, which already has coverage. Worth adding, but they read more naturally as tests of that shared path than of the bridge.

Fixed in 28b3498.

solutions/ess-maker-skills/scripts/adk_telemetry.py
deepaksharmacse12: Done — went with one reason per field.

```
correlationId -> invalid_correlation_id
mountId       -> invalid_mount_id      (was: invalid_correlation_id)
toolCallId    -> invalid_tool_call_id  (was: invalid_correlation_id)
```

Both new values are registered in `_CLIENT_EVENTS_REJECTED_REASONS`. That registration is load-bearing rather than bookkeeping: `_rejection()` coerces anything outside that set to `invalid_event_shape`, so a new reason that wasn't registered would have been silently swallowed instead of reported. There's a test pinning that guard so the next reason added can't regress it.

**On the coordination you flagged** — this is the part worth a second opinion. These extend Vorpal's bounded `BridgeRejectedReason` union, so until the matching `types.ts` / `TelemetryBridge.ts` change ships, `isRejectedReason` won't recognize them and `getTelemetryBridgeOutcome` maps them to `unrecognized_reason`:

| | before | after this PR, pre-Vorpal | after Vorpal ships |
|---|---|---|---|
| bad `mountId` | `invalid_correlation_id` (wrong field) | `unrecognized_reason` (no field) | `invalid_mount_id` |

No functional risk in the gap: an unrecognized reason is still classified as a **rejection**, so the batch is dropped rather than retried — identical behavior to today. The transitional trade is an honest "unknown" in place of a confidently wrong attribution, which I think is the better failure mode for triage.

The one thing I can't see from this side: **if any Vorpal dashboard or alert already buckets on `invalid_correlation_id`**, it'll lose the `mountId`/`toolCallId` traffic into `unrecognized_reason` until the union update lands. If that matters, this commit (fabf84f) is self-contained and easy to drop from the PR so the two can ship together — happy to do that instead. Otherwise I'll file the Vorpal-side union change as a follow-up.

Tests: +6 covering per-field reasons, distinctness, the bounded value-list guard, and `mountId`/`toolCallId` through the real `call_tool` path.

deepaksharmacse12:
> Deepak Sharma (@deepaksharmacse12) please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

deepaksharmacse12:

amilandi (dismissed):

asherjh:
## Overall

The bridge should pass through what Vorpal sends except where something concrete breaks, and most of the validation here doesn't clear that bar. The failure mode is uniformly harsh: nearly every check rejects the whole 25-event batch, and Vorpal drops a rejection without retry (`LoggerQueue.ts:108-112`), so one out-of-shape property costs 25 unrelated events permanently, with no field attribution and no signal beyond a counter nothing reads yet.

Two motives are worth separating. Strict shape validation in this repo buys no security — the validator is public source, the tool is unauthenticated on the maker's own machine, and the caller is a first-party app. The legitimate motive is Data Profile compliance, but that's an *accident* concern, and ADK's answer to accidents everywhere else is `_scrub` (`adk_telemetry.py:398`) or shed-oldest (`:167`), never refusal — `_emit` posts whatever it's handed. Reject only where scrubbing or dropping destroys the field's purpose.

The other constraint driving this: we want Vorpal to add events and properties without an ADK release. Several checks make that false today.

## Blocking

**Closed event key set.** `set(event) - _CLIENT_EVENTS_EVENT_KEYS` → reject is reachable, unlike its envelope-level twin — `events` is typed `Any`, so Pydantic passes nested keys through untouched. The day Vorpal adds any optional `BridgeEvent` field, every batch is rejected and silently dropped until an ADK release ships. `Logger.ts:75` already computes an info/error `level` and discards it at `:88`, so that distinction reaches Aria nowhere today and is the obvious next field; `locale` is in the type and never set, same story.

This also cuts against the PR's own design statement. Keeping no event-name allowlist makes new event *names* releaseless, but a closed key set makes new event *fields* hard-lockstep in both directions — and bumping `schemaVersion` to signal the addition is itself rejected.

## Structural: one JSON field instead of per-property columns

`build_event` puts custom fields flat in `data` (`flightcheck/telemetry.py:334`), so every `client_prop_<key>` becomes a new Aria column on `adk.client.event`, created the first time a widget logs that key. Three consequences none of the validations address:

- **Unbounded column growth.** The ≤25 cap is per occurrence, not per schema — 40 events at 5 keys each is 200 columns.
- **Type conflict.** `_valid_property_value` is type-agnostic per key, so `stage: "loaded"` and `stage: 3` both pass, and Aria fixes a column's type from what it first sees. Key *charset* is gated with batch-death as the penalty, an axis Aria doesn't care about; per-key *type* isn't gated at all.
- **Naming.** Vorpal keys are camelCase, so `client_prop_durationMs` lands beside `duration_ms` in the same tenant.

Emitting one scrubbed `client_properties` JSON string fixes all three, keeps the emitted field set fixed so `SCHEMA_VERSION` stays meaningful, and removes the justification for key-charset gating entirely. `parse_json` on the KQL side is cheap.

## Changes

**Delete — batch-killers on the additive path**

| Check | Why |
|---|---|
| Closed event key set | Above. Highest priority |
| `eventName` charset + 64 chars | It's a value in `client_event_name`, not a column name, so no schema risk. Scrub it. `StarterPrompts.Category.Reordered` is already 33 of the 64, and the two length rules disagree — `_valid_bounded_string` allows 200, the regex caps at 64 |
| Property key charset | Keys live inside a JSON string now. Bound key length only, so one key can't eat the blob budget |
| ≤25 properties, ≤10 array items | No precedent in the module, no stated rationale |
| ≤200 rejections on property values, `appName`, `buildEnvironment`, `buildNumber`, `locale` | Each refuses a string `_scrub` truncates to 200 one line later |
| ≤25 events | Exactly Vorpal's `TELEMETRY_MAX_BATCH_SIZE`, so raising the cap on the JW side kills every batch. Replace with a high OOM guard (~1000) so `batch_too_large` keeps a purpose |

**Delete — dead or misdirected**

| Check | Why |
|---|---|
| Closed envelope key set | Unreachable. FastMCP's arg model is `extra='ignore'`, so unknown top-level args are dropped before the body runs |
| 64KB serialized envelope | Measures the inbound envelope rather than the outbound records, so it enforces no real collector limit. A 413 would buffer and retry through the existing path — strictly better than dropping |
| `corr-`/`mount-`/`tool-` prefixes | The charset does the privacy work. The prefix hard-codes `getRandomId("corr")` into the OSS repo for no benefit, and the ids are already distinguished by field name. Apply the charset to the whole value instead of the suffix, and collapse the two disagreeing length bounds (`≤200` never binds against a `{1,64}` suffix) into one |
| Fail-open clamp to 25 | Unreachable — Vorpal splices `min(sameAppCount, 25)`, so `min(len, 25) == len` always. Cleanup, not a live bug |
| Duplicate `try/except` in `server.py` | `adk_telemetry.report_client_events` already catches everything |

**Downgrade — degrade the field, never the batch**

| Case | Behavior |
|---|---|
| Non-finite `timeSinceAppStart` | `-1`. Negative is unambiguously a sentinel since `performance.now()` is non-negative, and `0` is a real value for a bootloader event |
| `properties` not a dict | Drop the bag, emit the event |
| Non-finite number | Drop that property |
| Dict, nested list, other | Drop that property |
| `toolCallId` fails charset | **Omit the field, accept the batch.** The host mints this, so ADK is asserting a format it has no authority over — an id containing `:` or `/` would kill every batch. Unlike `correlationId` this is supplementary rather than what stitches a mount together |

**Add**

- Accept and emit `client_level`, so Vorpal can start sending it without another ADK release.
- Bump `SCHEMA_VERSION` from `1.1.0` for the new `client_*` field set, per the convention at `adk_telemetry.py:72` and `flightcheck/telemetry.py:97` ("Bump when the emitted field set changes so dashboards can version-gate"). One-time — with the blob the field set stops moving.
- `client_prop_dropped_count`, a bounded integer. Restores the diagnostic the rejections were providing, in the table we're already querying, at zero event loss.

**Rejection reasons**

`invalid_property_type` and `invalid_tool_call_id` both become unreachable. Delete them from ADK, from `BridgeRejectedReason`, and from `REJECTED_REASONS`. That leaves `invalid_mount_id` as the only union addition, which retires the coordination note in the description.

## GUID redaction on property values

`_scrub` rewrites any GUID to `<guid>`, which destroys joinability for exactly the values a correlation property exists to carry. The rule is miscalibrated for this position: `adk_telemetry.py:406` says it was written for exception text ("Dataverse exceptions can echo a user's UPN or object id") — unbounded free text ADK never chose. A property value is a literal a Vorpal call site deliberately set. Same regex, different risk, and it can't distinguish a resource id from a user object id in either direction.

Proposal: property values get a reduced scrub — newline flatten, `<path>`, `<url>`, `<email>`, truncate — with GUID redaction dropped. Full `_scrub` stays on error-message paths. Which specific identifiers are permissible to emit is a privacy call rather than something a regex settles; I'll confirm the set and put the constraint on `BridgePropertyValue` in `types.ts`, since Vorpal is where it's enforceable. `adk_telemetry.py:31` currently claims no developer/user identifier is ever emitted, so that line needs to match whatever we land on.

## Implementation notes

- **Scrub per value, then serialize — never scrub the serialized blob.** `_scrub`'s `(?<!\w)/[^\s]+` → `<path>` runs to the next whitespace, and compact JSON has none, so over `{"a":"/x","b":"y"}` it eats through to the closing brace.
- **Filter non-finite, then `json.dumps(..., allow_nan=False, separators=(",", ":"))`.** `json.dumps` emits bare `NaN`/`Infinity`, which is invalid JSON — inside the blob that makes it unparseable by `parse_json`. Drop the bag for that event if serialization still raises.
- **Don't scrub property keys.** They're call-site literals; rewriting them corrupts the field names analysts query.
- **Blob size cap** is the one size limit that's genuinely justified, against Aria's per-field string limit. Shed properties until it fits rather than truncating the string, which would leave unparseable JSON.

## What's right as-is

- **Reject-not-scrub for `correlationId`/`mountId`.** `_scrub` would rewrite `corr-<uuid>` to `corr-<guid>`, which is accepted and useless — every mount collapses to the same literal and stitching dies silently. Refusing is louder and honest, and it's cheap to be strict here specifically because Vorpal mints both at a single call site it controls, so a failure is a genuine contract violation. Keep the charset and the rejection on both.
- **`get_session()` only after validation.** Correct, and the reasoning in the description is the right one.
- **Untyped tool parameters.** The ToolError analysis holds — a narrow signature turns an out-of-contract envelope into a result with no `status`, which `getTelemetryBridgeOutcome` reads as transient.
- **Fail-open acceptance** and echoing the full sent count, for the same reason.

## Open

Aria's per-field string limit, which sets the blob cap.

deepaksharmacse12:
Thanks — this reframed the problem correctly. "Pass through what Vorpal sends except where something concrete breaks" is the right bar, and separating the security motive from the Data Profile one is what made the rest obvious: the validator is public source and the caller is first-party, so strictness bought no safety, while the accident concern already has an answer everywhere else in the module (`_scrub`, shed-oldest) that I didn't reach for.

Pushed as six commits, split so the behavior changes are isolated from the deletions.

| Commit | Covers |
|---|---|
| `26cd7c2` | Dead/misdirected: closed envelope key set, 64KB cap, fail-open clamp, duplicate `try/except` |
| `4644abc` | Prefix requirement dropped; charset applied to the whole value |
| `1c036ee` | **Blocking** — closed event key set removed |
| `91bd0f5` | **Structural** — one scrubbed `client_properties` JSON string |
| `f0564fc` | Batch-killers on the additive path |
| `398b7b8` | Degrade the field, never the batch |

Also in: `client_level`, `client_prop_dropped_count`, `SCHEMA_VERSION` → `1.2.0`. `invalid_property_type` and `invalid_tool_call_id` are gone, leaving `invalid_mount_id` as the only `BridgeRejectedReason` addition — which retires the coordination note from the description.

The implementation notes were all load-bearing, and two were genuine landmines I'd have hit. `_scrub` over compact JSON gives `{"a":"<path>` — the `(?<!\w)/[^\s]+` rule eats to end-of-string, so scrubbing per value before serializing isn't a style preference. And `json.dumps` emits bare `NaN`, which `parse_json` can't read. Both are covered by tests now.

On the closed event key set: you were right that it cut against the PR's own design statement, and worse, my test for the envelope-level twin was passing only because it called the bridge in-process and bypassed FastMCP. It was asserting a guarantee the transport never provided. Replaced with one that pins the real behavior.

**Two things still open, both yours:**

1. **Aria's per-field string limit.** `CLIENT_EVENTS_MAX_PROPERTIES_BYTES` is currently `8 * 1024` — a placeholder, not a derived number. It's the one value in the stack resting on an assumption, so I'd rather flag it than let it look settled.

2. **GUID redaction.** Left fully on: property values still emit `<guid>`. Your analysis is right that `_scrub` is miscalibrated for this position — it was written for exception text, not for literals a call site deliberately set — but relaxing redaction is a privacy call rather than an engineering one, and it contradicts the "no developer/user identifier" claim at `adk_telemetry.py:31`. Happy to land the reduced scrub the moment you confirm the permissible set, and I'll update that docstring in the same commit so the code and the claim can't drift. Agreed the durable constraint belongs on `BridgePropertyValue` in `types.ts`.

**One heads-up and one question:**

`CLIENT_EVENTS_MAX_BATCH_EVENTS` moved 25 → 1000. Worth surfacing because @amilandi specifically asked for exactly 25 to match `TELEMETRY_MAX_BATCH_SIZE`, which is the coupling you identified as the footgun. 1000 sits far enough above any plausible batcher max that it can't kill batches in either direction, so I think it resolves cleanly — but it is a reversal of earlier review guidance rather than a gap, and you two may want to agree explicitly.

That cap does mean a worst case of 1000 events × 8KB of properties ≈ 8MB in one call. Nothing in the bridge bounds the total payload now that the envelope cap is gone, which I think is correct per your 413-buffers-and-retries argument — but flagging it in case the OOM guard was meant to bound bytes rather than cardinality.

Deliberately not touched: `_THREADS` thread-safety. It predates this PR and `_emit_many` only mirrors the existing `_emit` dispatch, so a real fix means putting `_THREADS` behind a lock across every async emitter. Happy to file separately.

Suite is green: 109 tests across the two bridge files, 1989 in full, with the 10 pre-existing `test_installer_*` failures unchanged from the base commit.

asherjh (dismissed):
Re-reviewed at `f465122`. The rework is thorough and the commit split makes it easy to see which argument landed where. I verified the behavior independently rather than from the description — unknown event fields survive the real `call_tool` path, the blob scrubs per value before serializing, `NaN` is filtered out, non-dict bags and nested lists cost the property rather than the batch, and `toolCallId` omits instead of rejecting. 112 bridge tests pass in a clean checkout.

Two things worth calling out as good rather than just accepting. Moving the tool to the landing-page server (`3c4de35`) is a catch I missed entirely: MCP Apps binds a widget to the connection it loaded from, so the original placement would have failed every call with "Tool not found" and the Vorpal queue would have swallowed it — telemetry silently never arriving, with nothing surfaced anywhere. And catching that the envelope-key test only passed because it bypassed FastMCP is a better find than the item that prompted it; that test was asserting a guarantee the transport never made.

Approving. Everything below is non-blocking.

**[suggestion]** `solutions/ess-maker-skills/scripts/adk_telemetry.py:72-75` — the 1.2.0 changelog line describes a migration that never happened. There was no client bridge in 1.1.0, so nothing moved and no `client_prop_<key>` column was ever emitted in shipped data. The version gate exists so a dashboard owner can reason about what changed between versions, and this points them at a column set that never existed. Worth restating as what 1.2.0 adds — the `adk.client.event` field set.

**[suggestion]** Source comments narrate the review rather than the code — `:96-98` ("This used to be exactly Vorpal's `TELEMETRY_MAX_BATCH_SIZE`"), `:106-107` ("the charset rule that used to reject the batch"), `:159-160` ("replaces the old pairing of a <=200 check"), `:679` ("would have been trimmed anyway"), `:762` ("replaces the old per-key"), `:862` ("instead of costing 25 unrelated events").

Each defends the current shape against an alternative that only existed in this PR's earlier commits. This repo is public, so every future reader arrives without that context and has to reconstruct a design they never saw in order to parse a comment about the design that exists. The reasoning is good and worth keeping — it just wants to state the constraint rather than the delta. `:96` works as: "An OOM guard, not a contract — deliberately far above any client batcher max, so a client-side cap change cannot reject a well-behaved batch." Same information, no phantom. The before/after belongs in the PR description and commit messages, where it already is.

**[suggestion]** `adk_telemetry.py:31` — "enums only, no free text" no longer describes what the module emits, independent of the GUID question below. `client_event_name` and every property value are free text now; scrubbed, but free text. Worth correcting in whichever commit next touches that docstring.

**[nit]** `adk_telemetry.py:152` — `_CLIENT_EVENTS_IDENTIFIER_RE` is compiled identically at `:152` and `:161`, and the second shadows the first. `:152` is the leftover from `4644abc`.

**Both open items are mine, and neither should hold the merge.** I'll confirm the permissible identifier set before we relax GUID redaction — leaving it fully on is the right default, and I'm not going to assert that an agent id or environment id is safe to emit when either could be customer content. On `CLIENT_EVENTS_MAX_PROPERTIES_BYTES`, the research is a real answer: no documented per-field limit, 2.5 MB per event, ~3.15 MB per request. 8 KB is conservative by roughly 300x, and the comment says so plainly rather than dressing a guess as a derivation. Fine to ship; I'll follow up if anything tighter turns out to exist.

@amilandi — flagging the `CLIENT_EVENTS_MAX_BATCH_EVENTS` move from 25 to 1000, since it reverses your earlier ask. My read is that the request was for consistency with `TELEMETRY_MAX_BATCH_SIZE`, but matching it exactly is the coupling that turns a JS-side cap raise into total batch loss on this side: the cap is an OOM guard, not a contract, and 1000 satisfies the intent without the lockstep. Say so here if you read it differently.

Agreed on filing `_THREADS` separately — it predates this PR and a real fix means putting it behind a lock across every async emitter, which shouldn't ride along here.

solutions/ess-maker-skills/scripts/adk_telemetry.py
```diff
@@ -534,6 +606,339 @@ def _emit(event_name: str, data: dict[str, Any], *, block: bool = False) -> dict
...
+                "client_time_since_app_start_ms": _client_time_since_app_start(event),
+            }
+        )
+        if tool_call_id is not None:
+            row["client_tool_call_id"] = tool_call_id
+        if locale is not None:
+            row["client_locale"] = _scrub_client_scalar(locale)
+        level = event.get("level")
+        if level is not None:
+            row["client_level"] = _scrub_client_scalar(level)
```
rebova-microsoft, line 915: Assuming  level  is intended to be a string, could we guard that here and omit other shapes while keeping the batch accepted? Since nested events arrive as  Any , an object-valued level passes through  _scrub_client_scalar  unchanged. In an offline MCP repro, a synthetic email in that object remained in the serialized telemetry payload, while the same value in  properties  was scrubbed. I have not confirmed that the current logger produces this shape; this seems like a useful guard for the new field.

solutions/ess-maker-skills/scripts/adk_telemetry.py
```diff
@@ -534,6 +606,339 @@ def _emit(event_name: str, data: dict[str, Any], *, block: bool = False) -> dict
...
+def _scrub_client_scalar(value: Any) -> Any:
+    """Redact paths / URLs / emails / GUIDs from a bridge string value.
+
+    Bounded primitives still reach Aria verbatim otherwise, so a widget that
+    puts a UPN, a local path or an object id into a property value would
+    contradict the approved Data Profile. Non-strings pass through unchanged.
+    """
+    if isinstance(value, str):
+        return _scrub(value, limit=CLIENT_EVENTS_MAX_STRING_LENGTH)
+    return value
```
rebova-microsoft, line 758: Would it be worth bounding the work done by  _scrub  on long strings here? In an offline MCP repro, a valid property containing 32 KB of alphanumeric text without  @  took around 1.5 seconds before reaching a mocked emitter, despite async delivery being enabled. The email regex appears to retry across the string before the 200-character truncation happens. I am not sure how long we expect producer values to be, but avoiding those repeated scans would help keep telemetry from delaying other MCP calls while preserving redaction-before-truncation and batch acceptance.

rebova-microsoft (dismissed):

solutions/ess-maker-skills/scripts/adk_telemetry.py
asherjh: [note/non-blocking] Keeping the shared scrubber unchanged in this PR; this follow-up is limited to the optional `level` type guard.

asherjh (approved):

PR #251, opened 2026-08-25
Add a WeveNova planner MCP server sharing the agentconfig client core

jainharsheet77:
## What

Add a second MCP server, **`ess-planner`**, so the planner skill can drive the WeveNova **AgentConfiguration beta** surface — projects, plans, tasks, and plan role attestation — through repo-native code instead of an external MCP.

The change is **additive** and shares the landing-page client core rather than duplicating it. The two servers (`ess-landing-page-config` and `ess-planner`) sign in the same way and reuse the same HTTP/auth plumbing; only the endpoint surface differs.

## Layout

New files (all planner/role code lives here):

| File | Role |
| --- | --- |
| `_odata.py` | Dependency-free OData/JWT helpers shared by both endpoint surfaces (kept import-cycle-free) |
| `roles.py` | `RolesMixin` — tenant-sharded plan role attestation |
| `planner.py` | `PlannerMixin` — projects / plans / tasks on the beta base |
| `planner_client.py` | `PlannerClient = PlannerMixin + RolesMixin + AgentConfigClient` |
| `server_planner.py` | `ess-planner` FastMCP server exposing the tools |

**The only change to a landing-page-owned file** is a backward-compatible `transform_payload: bool = True` flag on `AgentConfigClient._request` (default preserves existing behavior; the planner/role surfaces pass `False` so responses carrying user-authored keys are returned verbatim). **`server.py` is unchanged** from the landing-page baseline and does not appear in the diff.

## Auth (unchanged)

`PlannerClient` inherits the landing-page client's MSAL delegated sign-in as-is:

- **Tenant** comes from the access token's `tid` claim (existing behavior).
- **Caller** comes from the token's `oid` claim (for "tasks assigned to me").

No `userName` / `aadId` / `tenantId` / `callerId` identity arguments are exposed on any tool — identity and tenant always come from the token.

## Tools (on `ess-planner`)

**Projects** — `list_agent_configuration_projects`, `get_agent_configuration_project`, `create_agent_configuration_project`, `archive_agent_configuration_project`

**Plans** — `list_project_plans`, `get_project_plan`, `create_project_plan`, `update_project_plan`, `archive_project_plan`

**Tasks** — `list_project_plan_tasks`, `list_project_plan_tasks_for_caller`, `get_project_plan_task`, `create_project_plan_task`, `create_role_assigned_project_plan_task`, `update_project_plan_task`, `set_project_plan_task_state`, `complete_project_plan_task`, `delete_project_plan_task`

**Role attestation** — `list_attestable_roles`, `list_plan_role_assignments`, `get_role_assignment`, `attest_plan_role`, `revoke_role_assignment`

## Design notes

- **Beta base** (`https://substrate.office.com/weveb2/api/beta`) addressed with absolute URLs; httpx uses them verbatim while still applying the shared bearer/Accept headers. Overridable via `AGENTCONFIG_PROJECTS_BASE_URL`.
- Routes: `me/agentConfigurationProjects('id')/agentPlans('id')/agentPlanTasks('id')`; role assignments are tenant-sharded at `tenants('{tid}')/agentRoleAssignments`.
- **camelCase** request bodies, returned **untransformed** (`transform_payload=False`).
- Optional **`If-Match`** (from an `etag` arg) and **`Idempotency-Key`** headers on mutations.
- Projects/plans have no DELETE route → archive via PATCH (`state:"Archived"` for projects, `status:"Archived"` for plans); only tasks have a real DELETE.
- Task completion normalizes outputs to the `PlanArtifact` shape (dedupes keys, enforces the Environment `environmentId` rule).
- Role attestation validates against the three provider-owned attestable roles (`WorkdayAdmin`, `ServiceNowAdmin`, `ServiceNowKnowledgeManager`); provider is always `External`.

## Validation

- `python -m py_compile` clean on all seven files (`client.py`, `server.py`, and the five new modules).
- `ruff check` clean on the five new modules, the `client.py` hook, and all three new test files (repo rules `E4,E7,E9,F`).
- **A committed test suite** was added under `tests/mcp/agentconfig/`, mirroring the existing agentconfig test conventions (flat `sys.path` import, fake `tid`/`oid` JWT, `httpx.MockTransport`, AST + import-based contract checks). `python -m pytest tests/mcp/agentconfig` → **60 passed** (20 existing + 40 new).

### Test coverage (per endpoint)

- **`test_planner_client.py`** — request-construction contract for **every** `PlannerClient` method: exact route, HTTP verb, camelCase body, headers (`If-Match` / `Idempotency-Key` / bearer / Accept), OData key encoding, `$filter` / `$orderby` queries, `oid`-derived caller scoping, validation errors (bad state/role, Environment `environmentId` rule, lifecycle-field rejection, duplicate output keys), `transform_payload=False` passthrough, and the `ATTESTABLE_ROLES` constant.
- **`test_planner_server_contract.py`** — static AST contract (no import) asserting the exact tool→args map (23 tools), the per-tool annotation constant, a docstring on every tool, and that no identity/tenant argument leaks into any signature.
- **`test_planner_server_protocol.py`** — imports the server and asserts `name` / instructions, the full tool set, each tool's `inputSchema` (`properties` == args, `required` == args without defaults), non-empty descriptions carrying the key safety phrases, the annotation hint triples per category, and invocation wiring via `call_tool` against a recording client fake.

## Follow-up

`server_planner.py` still needs to be registered in the planner skill's MCP config (replacing the `weve-plan` prototype tunnel) to be usable end-to-end — that wiring belongs to the planner-skill branch and is intentionally out of scope for this client PR.


solutions/ess-maker-skills/src/mcp/agentconfig/planner.py
```diff
@@ -0,0 +1,488 @@
...
+            )
+
+        try:
+            return await _perform(etag)
+        except AgentConfigApiError as error:
+            if error.http_status == 412:
+                entity = await refetch()
+                fresh = _entity_scalar(entity, "ETag", "@odata.etag")
+                if fresh and _normalize_etag(fresh) != _normalize_etag(etag):
+                    return await _perform(fresh)
```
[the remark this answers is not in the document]
jainharsheet77: Fixed in c71d7e6. `_mutate_with_etag_recovery` no longer replays on a 412: when the re-read shows the ETag advanced it raises an actionable 412 ("the entity changed since you read it (ETag advanced from X to Y); re-read and reapply"), preserving the `If-Match` lost-update guard rather than clobbering a concurrent edit. The MCP tool docstrings that advertised the old auto-retry were reconciled too.

Verified live against the WeveNova beta surface: a stale `If-Match` PATCH returns 412 and the prior value is preserved (no replay/clobber).

solutions/ess-maker-skills/src/mcp/agentconfig/planner.py
```diff
@@ -0,0 +1,488 @@
...
+            headers=_mutation_headers(idempotency_key=idempotency_key),
+            transform_payload=False,
+        )
+
+    async def update_project_plan(
+        self, project_id: str, plan_id: str, patch: dict[str, Any], etag: str
+    ) -> Any:
+        if not isinstance(patch, dict) or not patch:
+            raise ValueError("patch must be a non-empty object")
+        return await self._mutate_with_etag_recovery(
```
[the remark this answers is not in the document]
jainharsheet77: Fixed in c71d7e6. `update_project_plan` now rejects `{"status": "Archived"}` (case/whitespace-insensitive) with a ValueError that directs callers to `archive_project_plan`.

Verified live that archiving is genuinely destructive: after a raw archive PATCH the plan's task list came back empty (its tasks were cancelled), so routing this through the tool that carries the destructive annotations is warranted.

solutions/ess-maker-skills/src/mcp/agentconfig/client.py
```diff
@@ -336,8 +336,22 @@ def _agent_path(self, title_id: str) -> str:
...
 
-    async def _request(self, method: str, path: str, **kwargs: Any) -> Any:
-        """Execute a request with bounded retry for transient responses."""
+    async def _request(
+        self,
+        method: str,
+        path: str,
+        *,
+        transform_payload: bool = True,
+        **kwargs: Any,
```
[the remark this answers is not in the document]
jainharsheet77: Fixed in c71d7e6. `_request` now gates retries on idempotency: an ambiguous gateway failure (502/503/504) or a network `RequestError` on a non-idempotent request (an unkeyed create) is surfaced instead of replayed, so a committed-but-unacknowledged POST can't be duplicated. Reads, requests carrying `If-Match`/`Idempotency-Key`, and the explicitly-safe agent PATCH/DELETE keep their bounded retry; a 429 always retries (the service rejects it before doing work). This takes the "suppress retries for unsafe methods when no key is present" option you suggested.

solutions/ess-maker-skills/src/mcp/agentconfig/planner.py
```diff
@@ -0,0 +1,488 @@
...
+        plan_id: str,
+        task_id: str,
+        patch: dict[str, Any],
+        etag: str,
+    ) -> Any:
+        if not isinstance(patch, dict) or not patch:
+            raise ValueError("patch must be a non-empty object")
+        allowed = {field.lower() for field in _TASK_UPDATE_FIELDS}
+        for key in patch:
+            if key.lower() not in allowed:
```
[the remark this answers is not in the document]
jainharsheet77: Addressed in c71d7e6, with a caveat from live testing. I verified against the WeveNova beta surface that this endpoint deserializes these field names case-insensitively: a PATCH sending `{"Title": ...}` applied identically to `{"title": ...}`. So the premise that wrong casing is "rejected or ignored" does not hold on the live surface.

I still normalize each key to its canonical camelCase spelling before sending (and now reject a field named twice under different casing), so the emitted body is deterministic and stays correct if the backend ever tightens to case-sensitive parsing — but I updated the inline comment so it no longer claims the service requires camelCase. Happy to switch to strict exact-key validation instead if you prefer.

jainharsheet77:
## Planner MCP docstrings — live contract verification

Went through **every** tool docstring in `server_planner.py` (projects, plans, tasks, **and** role attestation) with a fine-tooth comb and live-verified each claim against the WeveNova AgentConfiguration **beta** surface (real API, ESS project).

### One inaccuracy found → fixed
`create_agent_configuration_project` advertised the **unsupported** project name `"Employee Self Serve"`. The planner agent's first get-or-create call therefore returned **400 ValidationError**, and every downstream step cascaded — this was the root cause of the "so many errors" run.

- Corrected to the supported name **`"Employee Self-Service"`** and appended the exact API error text so the agent can self-correct.
- Live check: `POST {name:"Employee Self-Service"}` → **200** (idempotent get-or-create, no duplicate); `POST {name:"Employee Self Serve"}` → **400** *"name must be one of the supported configuration experiences: Employee Self-Service, Workforce Insights"*.

### Everything else live-verified ACCURATE
- **Plans:** unknown fields → 400 (`Could not find member … on PlanCreateRequest`); lifecycle fields can't be combined → 400 (`Combining lifecycle fields … is not supported`); new plans start **Draft**; `configuringAgentName` required + 4-value enum.
- **Tasks:** `title` required → 400; initial state **`NotStarted`**; `NotStarted→Completed` blocked → 400 (`Transition 'NotStarted' -> 'Completed' is not allowed`); update accepts only `title/description/assignedToId/produces/consumes`; task is the **only** resource with a DELETE route (→ 204).
- **Complete:** output `kind` ∈ `Custom|Environment|Connection|KnowledgeSource`; `Environment` requires a non-empty `environmentId` → 400 (`artifact … of kind 'Environment' requires…`).
- **Roles (exact match):** the live attest enum is *"Workday Administrator (**WorkdayAdmin**), ServiceNow Administrator (**ServiceNowAdmin**), ServiceNow Knowledge Manager (**ServiceNowKnowledgeManager**)"* — identical to `ATTESTABLE_ROLES`; all three attest 200/201; provider **External** only (`Internal` → 400); `list` is plan-scoped with `Active|Revoked`; revoke is a soft-revoke to `Status=Revoked` (→ 204).

### Gates
`ruff` clean; **51/51** `tests/mcp/agentconfig` pass.

> MCP-surface change — mirrored identically across **#251** and **#253**; no functional/behavioral change, docstrings only.

solutions/ess-maker-skills/src/mcp/agentconfig/client.py
```diff
@@ -460,12 +500,14 @@ async def update_agent_config(
...
             self._agent_path(title_id),
             json=_to_api_payload(config),
+            idempotent=True,
         )
 
     async def delete_agent_config(self, title_id: str) -> dict[str, Any]:
         return await self._request(
             "DELETE",
             self._agent_path(title_id),
+            idempotent=True,
```
asherjh: update_agent_config, delete_agent_config, and search_agents all got idempotent=True, but create_agent_config (line 471) didn't — so it's the one existing landing-page method whose behavior actually changes here. Verified against a 503-then-200 mock transport: on this branch it raises on the first attempt, on the base branch it retries and succeeds on attempt 3. Same for a network RequestError at line 434. No existing test covers create-retry, which is why the suite stays green.

If dropping retry on an unkeyed create is deliberate, that's a defensible call — but it needs to be a stated decision with a test pinning it, not a side effect, and the PR description currently says the opposite ("default preserves existing behavior").

solutions/ess-maker-skills/src/mcp/agentconfig/client.py
```diff
@@ -336,8 +336,40 @@ def _agent_path(self, title_id: str) -> str:
...
+        ``idempotent`` gates whether an *ambiguous* transient failure — a 502/
+        503/504 gateway error or a network ``RequestError`` that may have landed
+        server-side after committing — is safe to replay. When ``None`` it is
+        inferred: safe for read methods and for any mutation carrying an
+        ``If-Match`` or ``Idempotency-Key`` header, unsafe otherwise. An unsafe
+        create is surfaced instead of retried so a committed-but-unacknowledged
+        POST is never silently duplicated. A 429 is always retried because the
+        service rejects it before doing any work.
+        """
+        if idempotent is None:
```
asherjh: I'd rather this not change the shared default at all. Inferring retry-safety from request headers means the next landing-page POST someone adds silently gets no-retry unless they know to pass idempotent=True — the failure mode is invisible until a gateway blips in production.

Can you invert it? Default idempotent=True to preserve today's semantics, and have the planner/role call sites opt out explicitly where a replay would genuinely duplicate. That keeps the safety win where you want it without making it a trap on my surface.

solutions/ess-maker-skills/src/mcp/agentconfig/planner_client.py
```diff
@@ -0,0 +1,50 @@
...
+from planner import PlannerMixin
+from roles import RolesMixin
+from _odata import (
+    DEFAULT_AGENTCONFIG_PROJECTS_BASE_URL,
+    _decode_object_id_from_jwt,
+    _validate_https_base_url,
+)
+
+
+class PlannerClient(PlannerMixin, RolesMixin, AgentConfigClient):
```
asherjh: This is one root cause showing up in three places, so I'm raising it once here.

AgentConfigClient is our landing-page client, not a shared core. Inheriting it wholesale gives PlannerClient list_agent_configs, search_agents, create_agent_config, get/update/delete_agent_config, view_agent_icon, the three open_* widget methods, _collection_path, _agent_path, and _unwrap_collection — 15 members it never calls, live on the planner surface. It also inherits the logger name from client.py:35, so planner retry warnings log under "ess-landing-page-config".

The same root cause is why _odata.py has to re-implement things that already exist: _require_odata_id is _validate_title_id + _encode_odata_key (client.py:58-76) minus the length cap, _validate_https_base_url is the inline block from AgentConfigClient.__init__ (client.py:279-291) extracted verbatim, and _decode_object_id_from_jwt is _decode_tenant_id_from_jwt (client.py:240) differing only in which claim it reads. None of the originals were changed to call the new helpers, so each of those is now two copies.

And it's why server_planner.py sits in agentconfig/ rather than its own directory. Every other MCP in the kit is one directory per server (adk/, servicenow/, workday/, dataverse/), but there's no __init__.py anywhere under src/mcp/ and each server launches with cwd set to its own folder on a flat sys.path, so a cross-directory import doesn't work without a path hack. Keeping it here was the right call under that constraint — I just don't think it's where this should land.

What's genuinely shared is about 60% of client.py: _CLIENT_ID/_SCOPE/_AUTHORITY, _resolve_token, the whole MSAL block, the JWT claim decode, AgentConfigApiError, _ensure_client, aclose, and _request. What's landing-page-specific is the v1.1 base URL, the PascalCase transform, the title-id helpers, and the EmployeeAgents methods. Pull the first group into a neutral base — src/mcp/weve/weve_client.py or similar — and have both AgentConfigClient and PlannerClient inherit it, each with its own logger name. Then the planner moves to src/mcp/planner/ and matches the rest of the kit.

One thing worth preserving either way: _TOKEN_CACHE_PATH is derived from client.py's own __file__, so both servers currently share one MSAL cache and one sign-in. A copy-pasted client in a separate folder would give users two caches and two browser prompts — the shared base keeps that working.

Happy for the extraction to be a follow-up rather than growing this PR, as long as the folder move rides along with it. What I'd like to avoid is the misnamed inheritance settling in as the permanent shape.

solutions/ess-maker-skills/src/mcp/agentconfig_core/_odata.py
```diff
@@ -0,0 +1,167 @@
...
+    if not isinstance(value, str) or not value or value != value.strip():
+        raise ValueError(
+            f"{name} must be a non-empty string without surrounding whitespace"
+        )
+    if any(ord(character) < 0x20 or ord(character) == 0x7F for character in value):
+        raise ValueError(f"{name} must not contain control characters")
+    return urllib.parse.quote(value.replace("'", "''"), safe="")
+
+
+def _escape_odata_literal(value: str, name: str) -> str:
```
asherjh, line 68: _escape_odata_literal and _require_odata_id open with the same six-line validation block. Worth pulling into a _validate_odata_string(value, name) that both call, since the only real difference is whether the result gets URL-quoted.

solutions/ess-maker-skills/src/mcp/agentconfig/roles.py
```diff
@@ -0,0 +1,138 @@
...
+ATTESTABLE_ROLES = ("WorkdayAdmin", "ServiceNowAdmin", "ServiceNowKnowledgeManager")
+_ROLE_ASSIGNMENT_STATUSES = ("Active", "Revoked")
+_ROLE_ASSIGNMENT_ORDERBY = ("createdAt asc", "createdAt desc")
+_ATTESTATION_PROVIDER = "External"
+
+
+class RolesMixin:
+    """Plan role-attestation methods (tenant-sharded on the token's ``tid``)."""
+
+    # Provided by the assembled client (AgentConfigClient / PlannerClient).
```
asherjh: "Provided by the assembled client (AgentConfigClient / PlannerClient)" — projects_base_url only exists on PlannerClient; AgentConfigClient never sets it. drop the first name.

solutions/ess-maker-skills/src/mcp/agentconfig/_odata.py
```diff
@@ -0,0 +1,167 @@
...
+from __future__ import annotations
+
+import base64
+import binascii
+import json
+import urllib.parse
+from typing import Any, Optional
+
+
+DEFAULT_AGENTCONFIG_PROJECTS_BASE_URL = "https://substrate.office.com/weveb2/api/beta"
```
asherjh: DEFAULT_AGENTCONFIG_PROJECTS_BASE_URL is public while every other symbol in this module is underscore-prefixed, yet planner.py and roles.py import the underscored ones across module boundaries anyway. Either convention is fine, but pick one — right now the prefix doesn't signal anything.

solutions/ess-maker-skills/src/mcp/agentconfig/planner.py
```diff
@@ -0,0 +1,531 @@
...
+        self, project_id: str, plan_id: str, query: Optional[dict[str, Any]] = None
+    ) -> Any:
+        caller_id = self._caller_object_id
+        if not caller_id:
+            raise AgentConfigApiError(
+                "The access token has no 'oid' claim; cannot scope tasks to the caller."
+            )
+        caller_filter = (
+            f"assignedToId eq '{_escape_odata_literal(caller_id, 'callerId')}'"
+        )
```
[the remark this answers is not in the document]
jainharsheet77: Fixed in 9434c43. `list_project_plan_tasks_for_caller` now resolves the caller's Active role assignments on the plan (via `list_plan_role_assignments(plan_id, subject_id=caller_id, status="Active")`) and expands the task filter to `assignedToId eq '<oid>' or assignedToRoleId eq '<roleId>' ...`, so the role-pooled tasks created by `create_role_assigned_project_plan_task` are no longer excluded. Any caller-supplied filter is preserved as `(<caller filter>) and (<supplied>)`. Pinned by `test_task_caller_scoping_expands_to_caller_direct_and_active_roles` and `test_task_caller_scoping_preserves_caller_supplied_filter`.

solutions/ess-maker-skills/src/mcp/agentconfig/client.py
jainharsheet77: Resolved in 9434c43 as part of inverting the retry default. `_request` now defaults to retry-safe (`idempotent=True` when unset), so `create_agent_config` — which carries no idempotency key — retries transparently again, matching the pre-planner landing-page behavior. The titleId-keyed EmployeeAgents POST is convergent (same titleId => same row), so replaying an ambiguous 5xx is safe. Pinned by `test_create_agent_config_retries_on_ambiguous_5xx`.

solutions/ess-maker-skills/src/mcp/agentconfig/client.py
jainharsheet77: Done in 9434c43. The header-inference is gone entirely. `_request` now uses `retry_safe = True if idempotent is None else idempotent` — the shared default is retry, and only the non-convergent planner creates opt out explicitly (`idempotent=idempotency_key is not None` on create_project_plan, create_project_plan_task and create_role_assigned_project_plan_task). Get-or-create project and attest_plan_role stay retry-safe by default, with a code comment noting why. A newly-added landing-page POST now retries by default instead of silently getting no-retry.

solutions/ess-maker-skills/src/mcp/agentconfig/planner_client.py
jainharsheet77: Done in 9434c43 — the largest part of this push. Extracted a neutral `weve/` core: `weve/weve_client.py` holds `WeveClient` (MSAL/bearer auth, `tid`/`oid` decode, the shared httpx session, the retrying `_request`, and a `_transform_response` identity hook) plus `weve/_odata.py`. `agentconfig/` and `planner/` are now siblings under `src/mcp/`, each composing the core: `AgentConfigClient(WeveClient)` keeps the EmployeeAgents routes and the PascalCase transform; `PlannerClient(PlannerMixin, RolesMixin, WeveClient)` carries only the projects/plans/tasks/roles surface. PlannerClient no longer inherits any landing-page routes. server_planner.py -> planner/server.py, ess-planner is registered in mcp.defaults.json, and both servers share one `weve/.local` MSAL cache + interactive sign-in.

solutions/ess-maker-skills/src/mcp/agentconfig_core/_odata.py
jainharsheet77: Done in 9434c43. Added `_validate_odata_string(value, name)` carrying the shared six-line validation block; `_escape_odata_literal` now returns its escaped result and `_require_odata_id` returns it unescaped (raising on empty). Both delegate to it.

solutions/ess-maker-skills/src/mcp/agentconfig/roles.py
jainharsheet77: Fixed in 9434c43. The class-attr comment now reads `# Provided by the assembled client (PlannerClient).` — `projects_base_url` only exists on PlannerClient.

solutions/ess-maker-skills/src/mcp/agentconfig/_odata.py
jainharsheet77: Fixed in 9434c43. Renamed to `_DEFAULT_AGENTCONFIG_PROJECTS_BASE_URL` so it matches the other underscore-prefixed symbols in the module; planner_client.py imports the underscored name. These helpers are an internal shared core imported across sibling folders on a flat sys.path, so the underscore convention is intentional.

solutions/ess-maker-skills/src/mcp/agentconfig_planner/planner_client.py
```diff
@@ -0,0 +1,64 @@
+# Copyright (c) Microsoft Corporation.
```
deepaksharmacse12, line 1: Do you want to rename this to client.py as discussed over chat thread?

solutions/ess-maker-skills/src/mcp/agentconfig_landing_page/.gitignore
```diff
@@ -0,0 +1,7 @@
+# Compiled Python
```
deepaksharmacse12, line 1: Concerns over the weve name, starting separate discussion over shared chat thread to hear different perspectives.

solutions/ess-maker-skills/src/mcp/agentconfig_planner/planner.py
```diff
@@ -0,0 +1,585 @@
...
+        for role_id in await self._caller_active_role_ids(plan_id, caller_id):
+            clauses.append(
+                f"assignedToRoleId eq '{_escape_odata_literal(role_id, 'roleId')}'"
+            )
+        caller_filter = " or ".join(clauses)
+        merged = dict(query or {})
+        existing = merged.get("filter")
+        merged["filter"] = (
+            f"({caller_filter}) and ({existing})" if existing else caller_filter
+        )
```
deepaksharmacse12, line 443: This lookup is case-sensitive but `_build_query_params` matches on `key.lower()`. If a caller passes `{"Filter": "state eq 'X'"}`, `merged` ends up holding both `Filter` and `filter`, both map to `$filter`, and the last write silently wins — the caller's filter gets dropped with no error.

Worth normalizing the incoming query keys before merging (and rejecting duplicates) so this can't happen.

solutions/ess-maker-skills/src/mcp/agentconfig_core/_odata.py
```diff
@@ -0,0 +1,147 @@
...
+    params: dict[str, str] = {}
+    for key, value in query.items():
+        if value is None:
+            continue
+        option = _QUERY_OPTION_MAP.get(key.lower())
+        if option is None:
+            raise ValueError(f"Unsupported query option: {key}")
+        params[option] = (
+            "true" if value is True else "false" if value is False else str(value)
+        )
```
deepaksharmacse12, line 146: Same class of issue more generally: two friendly keys differing only by case (`filter` / `Filter`) both resolve to `$filter` here and silently overwrite each other. Since the option map is already case-insensitive, I'd raise on a collision rather than let the last one win.

solutions/ess-maker-skills/src/mcp/agentconfig_planner/planner.py
```diff
@@ -0,0 +1,585 @@
...
+                headers=_mutation_headers(etag=active_etag),
+                transform_payload=False,
+            )
+
+        try:
+            return await _perform(etag)
+        except AgentConfigApiError as error:
+            if error.http_status == 412:
+                entity = await refetch()
+                fresh = _entity_scalar(entity, "ETag", "@odata.etag")
```
deepaksharmacse12, line 206: If `refetch()` (or `plan_refetch()` in the 409 branch below) itself throws — a 404 because the entity was concurrently deleted, a 403, or a transient 5xx — that exception propagates and replaces the original 412/409. The caller loses the precondition signal entirely and gets a confusing unrelated error instead.

Suggest wrapping the re-read in a try/except and re-raising the original `error` if the re-read fails.

solutions/ess-maker-skills/src/mcp/agentconfig_planner/planner.py
```diff
@@ -0,0 +1,585 @@
...
+        """Role ids the caller actively holds on this plan.
+
+        Role-pooled tasks are addressed to a role (in ``assignedToRoleId``), not
+        the caller's oid, so scoping "tasks for the caller" to direct
+        assignments alone would hide them. Resolving the caller's Active role
+        assignments for the plan lets the task query expand to those roles.
+        """
+        assignments = await self.list_plan_role_assignments(
+            plan_id, subject_id=caller_id, status="Active"
+        )
```
deepaksharmacse12, line 395: This issues a single unbounded `list_plan_role_assignments` with no `top`/`skiptoken` loop. If the service pages the response, the role set here is partial, and `list_project_plan_tasks_for_caller` then silently omits the caller's pooled tasks for the roles that fell off the first page — which is exactly the silent-omission failure this method exists to prevent. Either page through, or pass an explicit `top` and fail loudly if the result looks truncated.

solutions/ess-maker-skills/src/mcp/planner/planner.py
```diff
@@ -0,0 +1,585 @@
...
+            assignments.get("value")
+            if isinstance(assignments, dict)
+            else assignments
+        )
+        if not isinstance(entities, list):
+            return []
+        role_ids: list[str] = []
+        seen: set[str] = set()
+        for entity in entities:
+            role_id = _entity_scalar(entity, "roleId")
```
deepaksharmacse12: Reading `roleId` here (and filtering on `subjectObjectId` in `roles.py`) while `attest_plan_role` writes `subjectId` / `role` is an asymmetry that's only pinned by hand-written mocks. If the assignment entity actually spells the role field differently, this returns `[]`, role-pooled tasks silently vanish from `list_project_plan_tasks_for_caller`, and the tests still pass.

Has this path been exercised against the live beta API? If so, worth a comment recording that the read model and the attest action body genuinely use different field names.

solutions/ess-maker-skills/src/mcp/agentconfig_planner/roles.py
```diff
@@ -0,0 +1,147 @@
...
+            "target": {"type": "Plan", "id": plan_id},
+            "provider": provider,
+        }
+        return await self._request(
+            "POST",
+            f"{self._role_assignments_collection_url()}/attest",
+            json=body,
+            headers=_mutation_headers(etag=etag, idempotency_key=idempotency_key),
+            transform_payload=False,
+        )
```
deepaksharmacse12, line 146: This POST omits `idempotent=`, so it defaults to retry-safe and an ambiguous 502/504 gets replayed. Every other POST in this PR either opts out or carries an explicit comment justifying why replay is convergent (see `create_agent_configuration_project`).

If attest really is an upsert keyed on (plan, subject, role), please add the same comment here so the omission reads as deliberate. If it isn't, a replayed gateway error will create a duplicate assignment.

solutions/ess-maker-skills/src/mcp/agentconfig_core/base_client.py
```diff
@@ -0,0 +1,419 @@
...
+                ) from error
+
+            except httpx.RequestError as error:
+                last_error = error
+                if retry_safe and attempt < self.max_retries - 1:
+                    await asyncio.sleep((2**attempt) + random.uniform(0, 1))
+                    continue
+                raise
+
+        raise AgentConfigApiError(f"Maximum retries exceeded: {last_error}")
```
deepaksharmacse12, line 419: Retry exhaustion raises with `http_status=None`, so a persistent 429 or 503 reaches the caller unclassifiable — anything downstream that branches on `http_status` (like `_mutate_with_etag_recovery`) can't tell a rate-limit from anything else. Worth carrying the last status through.

Also, on the final attempt the loop still sleeps the full backoff at line 376 before falling out here, so every exhausted retry burns up to ~5s for nothing. Both are pre-existing, but the shared core now propagates them to the planner surface too.

solutions/ess-maker-skills/src/mcp/weve/weve_client.py
```diff
@@ -0,0 +1,419 @@
...
+logging.getLogger("httpcore").setLevel(logging.WARNING)
+
+_CLIENT_ID = "417219b4-3a7d-42a2-bdb1-972bd8281a02"
+_SCOPE = ["https://substrate.office.com/weve/.default"]
+_AUTHORITY = "https://login.microsoftonline.com/organizations"
+# Derived from this shared-core module's own location so every WeveNova MCP
+# (landing-page, planner) shares ONE MSAL cache and ONE interactive sign-in.
+_WEVE_DIR = os.path.dirname(os.path.abspath(__file__))
+_LOCAL_STATE_DIR = os.path.join(_WEVE_DIR, ".local")
+_TOKEN_CACHE_PATH = os.path.join(_LOCAL_STATE_DIR, "msal_token_cache.bin")
```
deepaksharmacse12: Moving the cache from `agentconfig/.local/` to `weve/.local/` is the right end state, but it's a silent breaking change for existing kit users: they get re-prompted for an interactive sign-in with no explanation, and the old cache — which holds refresh tokens — is left orphaned on disk under `agentconfig/.local/`.

Consider a one-time migration (move the old file if present, then delete it), or at minimum a release note.

solutions/ess-maker-skills/.vscode/mcp.defaults.json
```diff
@@ -4,6 +4,11 @@
       "command": "{pythonExecutable}",
       "args": ["server.py"],
       "cwd": "${workspaceFolder}/src/mcp/agentconfig"
+    },
+    "ess-planner": {
```
deepaksharmacse12, line 8: Registering this here means every kit user gets the server started and 23 new tools in their tool list, but nothing in the repo references `ess-planner` — no skill, no prompt, no doc — and no setup step installs `src/mcp/planner/requirements.txt`. It only works today because the planner's deps happen to be identical to `agentconfig`'s, which onboarding step1 installs; that's incidental, not by design.

Is shipping it unreferenced intentional for this PR? If the skill lands separately, I'd hold the defaults entry until then, or add the pip step alongside it.

solutions/ess-maker-skills/src/mcp/planner/planner_client.py
```diff
@@ -0,0 +1,64 @@
...
+import sys
+
+import httpx
+
+# Import the neutral core from the sibling ``weve`` folder. Each MCP server
+# launches with cwd set to its own directory on a flat sys.path (there is no
+# package __init__.py under src/mcp), so make that shared folder importable.
+sys.path.insert(
+    0, os.path.join(os.path.dirname(os.path.abspath(__file__)), "..", "weve")
+)
```
deepaksharmacse12: This `sys.path.insert(0, ...)` is repeated verbatim in four modules (`client.py`, `planner.py`, `planner_client.py`, `roles.py`), unnormalized, and prepends the shared dir ahead of the script dir — with generic top-level module names (`_odata`, `weve_client`, `planner`, `roles`) sitting on a flat path. It works today because nothing collides, but it's brittle as the surface grows. A single bootstrap module, or an actual package, would be sturdier.

Relatedly: `_odata`'s underscore-prefixed names are now a cross-module API consumed by three packages. Worth dropping the underscore to signal that.

deepaksharmacse12 (commented):
Solid refactor overall. The `WeveClient` extraction is clean, landing-page behavior is preserved via the `_transform_response` override, and the `idempotent=` gating for unkeyed creates is well-reasoned and well-tested. Preserving the 412 rather than replaying it is the right call, and the comment explains why. OData literal escaping, HTTPS-only base URLs, `follow_redirects=False`, and the AST contract test forbidding identity/tenant tool arguments are all good defensive choices. All 73 tests pass locally.

A few things I'd want addressed or answered, inline. The ones I care most about:

**Correctness**
- `list_project_plan_tasks_for_caller` looks up `filter` case-sensitively while `_build_query_params` matches case-insensitively, so a mixed-case `Filter` silently drops the caller's filter.
- `_mutate_with_etag_recovery` lets a failing re-read replace the original 412/409, so the caller loses the precondition signal.
- `_caller_active_role_ids` doesn't paginate; a partial role set silently omits pooled tasks.
- `attest_plan_role` retries on ambiguous 5xx with no idempotency key and no comment explaining why that's safe.

**Operational**
- The MSAL cache path moved with no migration — existing users get silently re-prompted, and the old cache (with refresh tokens) is orphaned on disk.
- `ess-planner` is registered in `mcp.defaults.json` for every user but nothing in the repo references it, and no setup step installs its requirements.

**Design**
- The repeated `sys.path.insert` across four modules and the cross-module underscore-prefixed API are brittle as this grows.
- Route shapes and the `roleId`/`subjectObjectId` vs `role`/`subjectId` read/write asymmetry are pinned only against hand-written mocks — has this been run against the live beta API?

jainharsheet77:
### Update: `c518edd` — neutral-core rename + verified fixes

Pushed changes addressing the review feedback on the shared client core and the planner role/ETag handling. Each item below was checked against the AgentConfiguration backend source before changing (or deliberately not changing) behavior. `tests/mcp` (73) pass and ruff is clean.

**Implemented**

- **Neutral-core naming.** Renamed `src/mcp/weve/` → `src/mcp/agentconfig_core/`, `weve_client.py` → `base_client.py`, `WeveClient` → `AgentConfigBaseClient`, `_WEVE_DIR` → `_CORE_DIR`, and the `WeveNova` docstrings → `AgentConfiguration`. The real service literals are intentionally left as-is — the `.../weveb2/api/{beta,v1.1}` base URLs and the `https://substrate.office.com/weve/.default` OAuth scope are server contracts, not naming.

- **Don't mask a 412/409 with the recovery read.** `_mutate_with_etag_recovery` now guards the `refetch()` (412) and `plan_refetch()` (409) recovery reads; if the recovery read itself errors, the *original* precondition/conflict error is re-raised instead of the secondary error, so the caller still gets the actionable stale-ETag / inactive-plan signal.

- **`roleId` → `role` (functional bug).** `_caller_active_role_ids` read `roleId` off each assignment in the list response, but the response projects the role under `role` — only the `$filter` grammar keys it as `roleId`. As written it returned `[]`, so role-pooled tasks were silently dropped from `list_project_plan_tasks_for_caller`. It now reads `role` (falling back to `roleId`). The unit mock returned `{"roleId": ...}`, which hid this — corrected to the real `{"role": ...}` shape.

- **Attest idempotency — documented, no change.** Confirmed against the backend that `attest` is a deterministic-row upsert (a given subject + role + target resolves to one stable grant row) and that an Idempotency-Key replays the cached result, so a retried attest converges rather than duplicating. Added a comment to that effect; no behavior change needed.

**Pushed back**

- **`planner_client.py` → `client.py`.** Keeping the distinct filename. The MCP servers run on a flat `sys.path` (no package `__init__`), so renaming this to `client.py` makes it collide with `agentconfig/client.py` as the top-level module `client`: under the shared pytest interpreter (all of `tests/mcp` runs in one process) the planner server's `from planner_client import PlannerClient` and the client tests would resolve to whichever `client` was imported first. The suite already avoids exactly this by giving each side a distinct top-level name (e.g. `test_planner_server_protocol.py` loads `server.py` under a unique importlib name). Happy to revisit if/when `src/mcp` becomes real packages.

The remaining stylistic notes (paging helper, default-server wiring, cache-path location, `sys.path` underscore) I can take in a follow-up — flag any you consider blocking.

solutions/ess-maker-skills/src/mcp/agentconfig_planner/planner_client.py
jainharsheet77: Keeping this as `planner_client.py` rather than `client.py`. Under `src/mcp` the modules live on a flat `sys.path` (no package `__init__`), so a top-level `client` module would collide with `agentconfig/client.py` — both would register as `sys.modules['client']`. All `tests/mcp` run in a single pytest process (default `prepend` import mode) and collection visits `agentconfig/` before `planner/`, so the planner server's `from planner_client import PlannerClient` and the client tests would bind to the wrong module. The repo already relies on distinct top-level names for this reason (e.g. `test_planner_server_protocol.py` loads `server.py` under a unique importlib name to dodge the `server` collision). Happy to revisit if/when `src/mcp` becomes real packages with `__init__.py`.

solutions/ess-maker-skills/src/mcp/agentconfig_landing_page/.gitignore
jainharsheet77: Done in c518edd. Renamed `weve/` → `agentconfig_core/`, `weve_client.py` → `base_client.py`, class `WeveClient` → `AgentConfigBaseClient`, `_WEVE_DIR` → `_CORE_DIR`, and the `WeveNova` prose → `AgentConfiguration`. The only `weve` tokens left are the real backend API literals — the `weveb2/api/...` request paths and the `weve/.default` OAuth scope — which are fixed by the service and can't be renamed.

solutions/ess-maker-skills/src/mcp/agentconfig_planner/planner.py
jainharsheet77: Good catch — `merged.get("filter")` is case-sensitive, so a caller passing a non-canonical-case key (e.g. `{"Filter": ...}`) would leave both `Filter` and the injected `filter` in the dict, and `_build_query_params` lowercases both onto `$filter` with last-one-wins. In practice this path is only reached from `list_project_plan_tasks_for_caller`, whose callers always pass the canonical lowercase `filter`, so it isn't reachable today. I'm tracking the general hardening (normalize / reject case-variant option keys) as a follow-up rather than widening this rename PR — see the reply on `_odata.py`. Shout if you'd prefer it in this PR.

solutions/ess-maker-skills/src/mcp/agentconfig_core/_odata.py
jainharsheet77: Agreed this is the right place to enforce it — `_build_query_params` currently does `params[option] = ...`, so two keys mapping to the same `$`-option silently last-wins. Pairing this with the planner-side note above; I'd like to land the raise-on-collision as a small follow-up so it ships with a dedicated test rather than folding it into this rename PR. Can pull it forward if you consider it blocking.

solutions/ess-maker-skills/src/mcp/agentconfig_planner/planner.py
jainharsheet77: Done in c518edd. Both recovery re-reads in `_mutate_with_etag_recovery` — the 412 `refetch()` and the 409 `plan_refetch()` — are now wrapped in try/except, and if the re-read itself fails we re-raise the *original* 412/409 error rather than masking it with the secondary failure.

solutions/ess-maker-skills/src/mcp/agentconfig_planner/planner.py
jainharsheet77: This query is intentionally unpaged. It's scoped to a single subject + single plan + `status=Active`, so the result is the set of roles that one subject holds on one plan — a small set bounded by the plan's role catalog (the planner only cares about the three `ATTESTABLE_ROLES`: `WorkdayAdmin`, `ServiceNowAdmin`, `ServiceNowKnowledgeManager`), not a user-growable collection, so the backend returns it in one page. Adding `$top`/`$skiptoken` handling would be dead code for this collection. Happy to add a defensive `$top` as belt-and-suspenders if you'd like, but paging logic isn't warranted here.

solutions/ess-maker-skills/src/mcp/planner/planner.py
jainharsheet77: Done in c518edd, and yes — verified against the AgentConfiguration backend. The role-assignment *response* projects the role name under `role`, while the `$filter` grammar keys it as `roleId` (the value in both cases is the role *name*, e.g. `WorkdayAdmin`). So the old `_entity_scalar(entity, "roleId")` read returned `None` and silently dropped role-pooled tasks — a real latent bug, not just a mock artifact. It now reads `_entity_scalar(entity, "role", "roleId")` (response field first, filter-style key as fallback) and I corrected the mock that was masking it. The `roleId eq '...'` filter assertion stays, since that's the correct filter key.

solutions/ess-maker-skills/src/mcp/agentconfig_planner/roles.py
jainharsheet77: Done (documented) in c518edd. Confirmed against the backend that attest is a deterministic-row upsert — the grant row is keyed on (subject, role, target), so a replayed POST updates the same row rather than creating a duplicate, and an Idempotency-Key replay returns the cached result. It's therefore retry-safe without an explicit `idempotent=` flag; I added a comment on the call documenting exactly that. No behavior change.

solutions/ess-maker-skills/src/mcp/agentconfig_core/base_client.py
jainharsheet77: Both fair, and both are pre-existing in the shared core (this PR only renamed the file). (1) `Maximum retries exceeded` raises with `http_status=None`, so a caller can't tell a retried-out 429/503 apart from an unclassified error — it should carry the last observed status. (2) The 429/503 branch sleeps its backoff even on the final attempt (the `httpx.RequestError` branch already guards with `attempt < max_retries - 1`; the status branch should match). Both are minor observability/latency papercuts with no correctness impact, so I'm deferring them to a small follow-up to keep this PR scoped to the planner surface + rename. Tracked.

solutions/ess-maker-skills/src/mcp/weve/weve_client.py
jainharsheet77: This is an intentional consolidation, not an accidental relocation: both servers now share a single MSAL token cache under `agentconfig_core/.local/` (`_CORE_DIR/.local`) instead of separate per-server caches. The only impact is a one-time token re-acquisition (silent refresh, or an interactive sign-in) the first time after upgrading — the cache is derived state, so there's no data loss. For a preview MCP I'd rather not carry a migration shim for a cache file; I'll add a note to the PR/release description so the re-auth isn't a surprise.

solutions/ess-maker-skills/.vscode/mcp.defaults.json
jainharsheet77: Both observations are fair on #251 in isolation. The `ess-planner` server is the deliverable of this PR; it's registered so the planner skill — which lands separately — can consume it, but you're right that nothing in *this* PR references it yet, and that onboarding only installs `agentconfig/requirements.txt` (planner works today only because its deps are a subset). Two ways to tighten it: (a) add a `pip install -r src/mcp/planner/requirements.txt` step to onboarding, or (b) hold the `ess-planner` defaults entry until the planner skill merges so users don't get the extra tools in the meantime. I lean toward (b) — let me know your preference.

solutions/ess-maker-skills/src/mcp/planner/planner_client.py
jainharsheet77: The `sys.path.insert` bootstrap is duplicated deliberately. There's no package `__init__` under `src/mcp`, and each server launches with its own cwd, so every module has to put its own directory on the path *before* it can import its siblings — a shared bootstrap helper would itself need to already be importable, which is the chicken-and-egg this avoids. The flat, distinct top-level names are load-bearing for the same reason `planner_client.py` isn't `client.py` (see that thread). I can wrap the inserted path in `os.path.normpath` as a cosmetic tidy-up. On the underscore: `_odata` is intentionally marked private-to-the-MCP-tree — it's shared across sibling modules on the flat path, not a published package API, so I'd like to keep the underscore as the 'internal surface' signal.

PR #257, opened 2026-09-01
evaluations: add end-to-end authoring, review, and execution

nehaoss:
## Description

Adds an end-to-end Copilot Studio evaluation lifecycle to the ESS Maker Kit.

This PR includes:

- Agent-grounded and catalogue-grounded test-set generation
- Synchronized Copilot Studio `.mcs.yml` and CSV artifacts
- Automated evaluation quality validation
- Maker editing and judge/SME feedback workflows
- Feedback-only reviewer experience
- `review_requested` and `review_completed` state synchronization
- Safe workspace-to-agent promotion and scoped push
- Connected Copilot Studio profile discovery
- Evaluation run execution, history, and result analysis
- Explicit messages when review state prevents execution
- Structured choice controls for test-set and workflow selections

The branch also incorporates the latest `main` changes while preserving the existing planner routing behavior.

## Related issue

Refs #[ADD-STORY-OR-ISSUE-NUMBER]

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update
- [ ] Refactor / cleanup

## Testing

Ran the focused evaluation workflow test suite covering:

- Evaluation generation and CSV exports
- Automated quality validation
- Review-state reconciliation
- Workspace promotion and cleanup
- Scoped push behavior
- Test-set discovery and execution
- Run history and result analysis
- Maker and reviewer routing

Result: **88 tests passed**.

## Checklist

```text
Validation
- YAML parse: N-A
- AdaptiveDialog kind: N-A
- XML parse: N-A

No files under  samples/  are changed by this PR.
```


apurvabanka:
## Blocking


**1. PR introduces a test failure.** `tests/scripts/test_eval_scenario.py::test_create_command_gives_explicit_evaluation_intent_precedence` passes on the merge base, fails at head. The rewrite of `create.prompt.md` dropped the clause `when the request also contains evaluation file paths`. The PR claims "88 tests passed" — that was a focused subset. Full run: 364 passed / 11 failed, of which 10 installer failures are pre-existing and 1 is new.


**2. Ruff fails.** `F841 Local variable 'list_runs_parser' is assigned to but never used` — `scripts/evaluation_runs.py:707`.


**3. `AttributeError` on Python 3.11.** Repo declares `requires-python = ">=3.11"` / `target-version = "py311"`. In `evaluation_csv._passing_score`, the except branch sets `score = 70` (an `int`) then calls `score.is_integer()` — that method only exists on `int` from 3.12. Any `CompareMeaningGrader` with a `null`/non-numeric `threshold` crashes CSV export, and CSV export runs inside `/setup` (`regenerate_evaluation_exports`), so a single bad remote set breaks setup. Reproduced: `_passing_score(None)` → `AttributeError`. No test covers the fallback path.


## Functional gaps


**4. Review-blocked runs give the wrong message.** `evaluation_runs.main()`'s `run` subcommand calls `list_agent_test_sets(...)` *without* `include_blocked=True` (line ~726), so a review-blocked set is silently filtered out and the user gets `"The selected test set is not active or no longer exists."` instead of `REVIEW_PENDING_GUIDANCE` / `REVIEW_COMPLETION_NOT_PUSHED_GUIDANCE`. This directly contradicts the PR description's "Explicit messages when review state prevents execution." Only `list-sets` is tested for blocking; the `run` path is untested.


**5. Review gate has a hole.** A set with local `review_completed` but no baseline `review.json` (deployed status `None`) is `runnable: True`. Combined with #4, review enforcement depends entirely on a baseline file that `/setup` deletes whenever the remote description has no marker.


**6. `discover_review_sets` can never return workspace sets.** It requires `localStatus == deployedStatus == status`, but `_deployed_review_status` hardcodes `NOT_DEPLOYED` for `source == "Workspace"`. If that's intentional, it isn't documented; if not, workspace-staged sets silently never appear to a reviewer.


**7. Uncaught `KeyError` in `_runtime`.** `config["agent"]["botId"]` / `["folder"]` aren't guarded, so a partial `.local/config.json` produces a raw traceback rather than the `EvaluationRunError` path everything else uses.


**8. Second interactive sign-in.** `resolve_mcs_connection` spins up a separate `PPAdminClient` and calls `authenticate(include_flow=False)` after `PowerPlatformClient.authenticate()` already ran, so `/run` can prompt the maker for a browser sign-in twice.


**9. `cleanup_workspace_set` may leave stale CSVs.** It derives the glob stem via `evaluation_export_stem(destination)` (agent folder), but deletes from the *workspace* exports folder. `_unique_export_stem` disambiguates against sibling folders, so the two locations can resolve to different stems and the workspace CSV survives.


## Documentation / registration gaps


**10. `/run` is unregistered outside the menu.** It's in `menu.prompt.md` but missing from the repo-root `.github/copilot-instructions.md` kit-command list (which drives the wrong-folder redirect) — so `/run` typed at the repo root gets no redirect. `/review` is missing there too.


**11. `solutions/ess-maker-skills/.github/copilot-instructions.md`** — the `src/skills/` grounding line and the CRUD skill-routing table were not extended for `run/SKILL.md` or `review/SKILL.md`.


**12. `solutions/ess-maker-skills/README.md` untouched.** It documents "Generate Evaluation Test Sets" but has nothing for the new generate → tag for review → review → push → run → results lifecycle. The PR checks "I have updated documentation as needed."


## Minor



- `evaluation_runs.list_agent_test_sets` re-parses `review.json` with raw `json.loads` instead of `evaluation_review.parse_review_metadata`, so a malformed file is treated as "no review status" (fails open) rather than raising.

- `push.py`'s staged-description loop uses a bare `except Exception`, which will swallow programming errors as push failures.

- `_case_passed` is evaluated twice per case in `analyze_run_results`.

- PR body still has the `Refs #[ADD-STORY-OR-ISSUE-NUMBER]` placeholder.

CC: @nehaoss 

Fixed below @apurvabanka

• Broken routing test caused by line wrapping.
• Ruff unused-variable warning.
• Python 3.11  score.is_integer()  crash.
• Blocked runs now show the actual review-blocking reason.
• Locally completed but unpushed reviews remain blocked.
• Incomplete config now returns clear  /setup  guidance instead of  KeyError .
• Workspace review discovery wording now explains that only pushed/synchronized sets are reviewable.
• Collision-safe workspace CSV files are cleaned correctly.
• Invalid  review.json  files now fail explicitly.
• Review push catches only expected metadata errors.
• Test-case pass/fail is calculated once.
•  /run  is registered in the root-folder redirect and documented in the README.
• README now explains generation, review, push, execution, history, and results.

jainharsheet77:
## Intent & PM-alignment questions on the evaluation lifecycle

Thanks for this — the eval authoring/review/run lifecycle here is solid. Before sign-off I want to confirm whether a few **observed behaviors match the intended design**, or whether they're gaps to close. These are questions about *intent and PM alignment* — not code requests.

**Scope observation.** This PR changes 36 files, none under `src/skills/planner/`, and the eval skills contain no reference to the plan or the planner. So the eval subsystem and the planner look intentionally decoupled here.

> **Q0 — Was that the intended scope for this PR: the eval lifecycle only, with planner↔eval integration tracked as separate work?**

### 1. Grounding when the agent doesn't exist yet
The dispatcher routes "no matching topic / no configured agent" to **catalogue-grounded** generation, which grounds on the static bundled `ess-catalogue.md`. The vision I understood was: when the agent isn't built yet, ground the eval on the **intent captured by the planner** (the scenarios the sponsor picked at intake).

> **Q1 — Is grounding on the static catalogue (rather than the plan's captured intent) the intended behavior for the greenfield / no-agent case? Is that aligned with what PM described?**

### 2. Generate vs. preview at intent-capture
The planner's Phase 5 renders golden prompts as a **render-only preview** ("writes no files, creates no eval records, pushes nothing"). The vision described *generating* an eval from captured intent — a file, pushed to storage, recallable later.

> **Q2 — Was the eval supposed to be generated and persisted at intent-capture, or is preview-only the intended v1 (with real generation deferred to the post-topics "Generate evals" task)?**

### 3. A single shared world-view of the eval
Today eval **sets** persist to Dataverse and eval **runs** to the Power Platform test-runs API (shared across maker/judge/SME). But the **plan** itself carries no eval reference or last-run state, so there isn't one plan-level "world-view" of the eval that a third person can read after the environment exists.

> **Q3 — Is the shared eval world-view intended to live in Copilot Studio/Dataverse (not the plan)? Given the "everyone shares one plan view" goal, is that aligned with PM?**

### 4. Env / agent hand-off through the plan
The plan models `setup → produces primaryEnvironment → evals consume it`, but on this branch the plan is local-only and the eval run reads the environment/agent from `.local/config.json`, not from a re-fetched shared plan.

> **Q4 — Was "re-fetch the latest plan, see the environment id + agent id, then run the eval" meant to be wired in this PR, or does it intentionally depend on the planner-server sync landing separately?**

### 5. Triggering a run from capture
Running an eval is a separate `/run` flow that requires a pushed set and a configured agent.

> **Q5 — Was intent-capture supposed to offer to trigger a run when the agent is already configured, or is the manual `/run` the intended entry point?**

---

**Net question for PM:** Is this PR intended to be the **eval-lifecycle half only** — with planner-grounded generation, plan-persisted eval state, and the shared world-view explicitly **out of scope** here (to arrive once the planner-server sync merges)? If yes, these are just tracking questions and nothing changes here. If no, these five are the alignment gaps to reconcile before merge.

solutions/ess-maker-skills/src/skills/evaluations/generate/SKILL.md
```diff
@@ -0,0 +1,295 @@
+# Generate Evaluation Test Sets Skill (catalogue-grounded)
+
+Generate Copilot Studio evaluation test sets for the scenario/goal(s) a user
+names, grounded in the **scenario catalogue bundled with this skill** — no
```
jainharsheet77, line 4: Intent question (no code change requested): here the no-agent path grounds on the static bundled catalogue. The vision I understood was that, before an agent exists, the eval should be grounded on the **planner's captured intent** (the scenarios the sponsor picked at intake) rather than a static catalogue. Was catalogue-only grounding the intended behavior for the greenfield case, and is that aligned with PM?

solutions/ess-maker-skills/src/skills/evaluations/dispatcher/SKILL.md
```diff
@@ -0,0 +1,67 @@
...
+
+When one or more topics clearly implement the requested scenario, read
+`src/skills/evaluations/create/SKILL.md` and follow it. Carry the requested
+scenario and matched topic names into that skill as the generation scope.
+
+If several topics contribute to the same scenario, include all of them. Do not
+ask the user to select topics unless the candidates represent genuinely
+different scenarios.
+
+### No matching topic found
```
jainharsheet77, line 59: Intent question (no code change requested): this branch routes "no matching topic / no configured agent" straight to catalogue-grounded generation, with no consultation of the plan. Was the no-agent branch supposed to first look at the plan's captured intent when a plan exists, or is routing to the static catalogue the intended v1? Trying to confirm this matches PM's design.

solutions/ess-maker-skills/src/skills/evaluations/run/SKILL.md
```diff
@@ -0,0 +1,199 @@
...
+
+Run:
+
+```text
+python scripts/evaluation_runs.py list-runs
+```
+
+The script queries the Power Platform `testruns` endpoint and joins each run's
+`testSetId` to the test-set `displayName` returned by the `testsets` endpoint.
+This is the same behavior for makers, judges, and SMEs; no local run mapping is
```
jainharsheet77, line 132: Intent question (no code change requested): run history here is shared through the Power Platform test-runs API — a good shared view for maker/judge/SME. But it lives in Copilot Studio/Dataverse, not on the plan, so the plan has no last-run/state a third person could read as one world-view after the environment exists. Was the shared eval state intended to live only in Dataverse, or was it also meant to be reflected on the plan? Checking alignment with PM's "one plan view" goal.

solutions/ess-maker-skills/scripts/evaluation_csv.py
```diff
@@ -0,0 +1,212 @@
...
+    generated = []
+    for set_folder in sorted(evaluations.iterdir()):
+        if not set_folder.is_dir() or set_folder.name.casefold() == "exports":
+            continue
+        if not any(
+            _load_yaml(path).get("kind") == "EvaluationSet"
+            for path in set_folder.glob("*.mcs.yml")
+        ):
+            continue
+        generated.append(generate_set_csv(set_folder, exports, timestamp))
```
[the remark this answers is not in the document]
nehaoss: Not fixing this: malformed evaluation source should surface explicitly rather than produce a partial-success setup. If desired later, a better enhancement would aggregate and clearly report invalid sets—not silently skip them.

solutions/ess-maker-skills/scripts/evaluation_runs.py
```diff
@@ -0,0 +1,778 @@
...
+    names = _case_names(Path(agent_folder))
+    cases = result.get("testCasesResults")
+    if isinstance(cases, list):
+        result["testCasesResults"] = [
+            {
+                **case,
+                "testCaseName": names.get(
+                    str(case.get("testCaseId", "")),
+                    "Unknown test case",
+                ),
```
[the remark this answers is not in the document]
nehaoss: fixed

solutions/ess-maker-skills/src/skills/evaluations/generate/SKILL.md
nehaoss: Yes, that is the intended behavior. The scenario or goal captured during planner intake defines the evaluation scope. The bundled catalogue only grounds and expands that selected scenario into supported test cases for the greenfield/no-agent path; it does not choose or override the user’s intent. This is aligned with the planned UX.

solutions/ess-maker-skills/src/skills/evaluations/dispatcher/SKILL.md
nehaoss: Good catch. The evaluation flow was designed to work independently, so the current no-topic path uses the scenario named in the request and grounds it against the catalogue. Planner integration is not complete yet. In a planner-driven conversation, it should first reuse the scenarios/goals already captured in the conversational or plan context, then use the catalogue to ground and expand them. It should ask the user only when that context is unavailable. We’ll address this integration gap.

solutions/ess-maker-skills/src/skills/evaluations/run/SKILL.md
nehaoss: Evaluation execution is intentionally independent from the planner.
 A run requires a configured Copilot Studio agent, deployed test sets, and an MCS connection, while the plan currently includes only evaluation generation. 
Full run history remains authoritative in the Power Platform test-runs API and is shared across makers, reviewers, and SMEs.
 PM has not requested run state to be duplicated into the plan, so no plan synchronization is included in this change.

apurvabanka (approved):

PR #259, opened 2026-09-02
Split capability taxonomy so every user-facing command has its own dashboard wedge

amilandi:
## Summary

Split the ADK `adk_capability` taxonomy so every distinct user-facing command has its own slice on the Capability Usage donut. Umbrella labels (a single `evaluations` or `publishing` wedge covering multiple commands) hid the signal PMs need to see which specific action drove a change in usage, and left several user-facing skills with no wedge at all.

## Changes

| Before | After |
| --- | --- |
| `evaluations` (one wedge for create/update/delete/validate) | `evaluation_create`, `evaluation_update`, `evaluation_delete`, `evaluation_validate` |
| `publishing` (one wedge covering both `push.py` and `publish.py`) | `push` (push.py — upload local edits to Dataverse) + `publishing` (publish.py — make changes live) |
| _(no wedge)_ | `onboarding`, `topic_review`, `topic_test`, `workflow_test` |

### push.py rename

`push.py`'s `emit_build_start` / `emit_build_complete` / `emit_agent_deploy` events are re-stamped from `adk_capability="publishing"` to `"push"`, and `push.py` now emits a best-effort `emit_capability_use("push")` at the end of the flow so push gets its own wedge on the Capability Usage donut (mirroring `publish.py`).

> **Aria tile impact:** any tile filtering Build Outcomes / Agent Deploy on `adk_capability="publishing"` needs updating to include both `"publishing"` and `"push"` (or split into two tiles).

### Not included in this PR

- **`discover`** — owned by #238. That PR will add `discover` to `ADK_CAPABILITIES` and wire it into `discover_inventory.py`.
- **`planner`** — owned by a separate feature branch not yet merged. It will add its own taxonomy entry.
- **`flightcheck`** — stays in the taxonomy but is intentionally not part of the Capability Usage donut split; it already has dedicated FlightCheck dashboard panels.

## Tests

- `tests/test_adk_telemetry.py::test_wired_capabilities_are_in_canonical_list` updated for the new granular values.
- The happy-path emit test now uses `evaluation_validate` (was `evaluations`).
- The `test_no_caller_passes_a_noncanonical_capability_to_the_shim` guard from #254 continues to catch drift automatically — it greps `scripts/*.py` and `SKILL.md` for `emit_capability.py <cap>` invocations and asserts every value is in `ADK_CAPABILITIES`.
- `pytest tests/test_adk_telemetry.py` → 67 passed.

## ADO

https://o365exchange.visualstudio.com/O365%20Core/_workitems/edit/7830948

### DIFF
--- a/solutions/ess-maker-skills/scripts/adk_telemetry.py
+++ b/solutions/ess-maker-skills/scripts/adk_telemetry.py
@@ -97,37 +97,61 @@ EVENT_FLIGHTCHECK_ERROR = "adk.flightcheck.error"
 # capability here, also add it to that Aria cube dimension value-list (see the
 # telemetry dashboards story, ADO #7532631) so the new slice renders.
 #
-# One capability per real maker-facing ADK skill / entry point:
+# One capability per real maker-facing ADK skill / entry point. The taxonomy
+# is intentionally granular: every distinct command a maker can run has its
+# own donut slice, so we can see which specific action drove a change in
+# usage. Umbrella labels (a single "evaluations" or "publishing" wedge
+# covering multiple commands) hide that signal, so we split them apart.
 #   setup                   -> first-run environment setup + discovery
 #                              (discover / list_environments are sub-steps of
 #                              this flow and do NOT emit their own capability;
 #                              the adk.agent.create event at the end of setup is
 #                              tracked separately by the "Agents Created" KPI and
 #                              is NOT a capability-donut slice)
+#   onboarding              -> workspace bootstrap after foundation setup
 #   connect                 -> ServiceNow / Workday connection setup
-#   topic_*                 -> topic authoring (create / update / delete)
-#   workflow_*              -> workflow authoring (create / update / delete)
-#   evaluations             -> eval test-set authoring + validation runs
+#   topic_create            -> author a new topic
+#   topic_update            -> modify an existing topic
+#   topic_delete            -> delete a topic
+#   topic_review            -> advisory conformance review of a topic
+#   topic_test              -> debug-and-validate loop for a topic
+#   workflow_create         -> author a new workflow
+#   workflow_update         -> modify an existing workflow
+#   workflow_delete         -> delete a workflow
+#   workflow_test           -> debug a workflow via run history
+#   evaluation_create       -> author eval test cases
+#   evaluation_update       -> modify eval test cases
+#   evaluation_delete       -> delete eval test cases
+#   evaluation_validate     -> quality-score generated eval sets
 #   cleanup                 -> error scan / fix pass
 #   troubleshoot            -> connectivity / auth diagnosis
 #   backup_template_configs -> Workday template-config backup
 #   restore_template_configs-> Workday template-config restore
-#   publishing              -> push / deploy to Copilot Studio
+#   push                    -> upload local changes to Dataverse (staging)
+#   publishing              -> publish so pushed changes go live in runtime
 #   flightcheck             -> pre-deployment readiness check
 ADK_CAPABILITIES = (
     "setup",
+    "onboarding",
     "connect",
     "topic_create",
     "topic_update",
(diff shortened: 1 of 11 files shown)


daeunJe0ng:
**[copilot-review] Review: Request changes (minor)**

Nice cleanup. The per-command split is the right granularity, and documenting the `in {value-list}` -> `not in <blank>` history in the header comment is genuinely helpful. One blocker and a couple of smaller notes.

**Blocking**

- `onboarding` is added to `ADK_CAPABILITIES` and to the `wired` set in `test_wired_capabilities_are_in_canonical_list`, but nothing emits it. No `emit_capability.py onboarding` line was added to `src/skills/onboarding/SKILL.md`, and there's no Python emitter. So the value can't produce a donut wedge, which contradicts the "no wedge -> now has wedge" row for `onboarding` in the PR description. Its siblings in that same row (`topic_review`, `topic_test`, `workflow_test`) each got an emit line. Fix: either add `python scripts/emit_capability.py onboarding` to the onboarding skill, or drop `onboarding` from both the tuple and the `wired` set until it's actually wired.

**Should-fix**

- `test_wired_capabilities_are_in_canonical_list` only checks `wired ⊆ ADK_CAPABILITIES`, so adding a value to both sides always passes and the `wired` set becomes an unverified hand-maintained claim (that's exactly how `onboarding` slipped through). Consider a reverse check that greps the SKILL.md / scripts for `emit_capability.py <cap>` and asserts every non-event capability in `wired` is actually found, so a declared-but-dead value fails CI.

**Question / nit**

- The description says `discover` is "owned by #238. That PR will add `discover` to `ADK_CAPABILITIES`", but #238 is already merged, and `discover` is still not in the canonical list on `main`. If `/discover` emits `emit_capability.py discover` today, it currently normalizes to `unknown`. That's pre-existing and out of scope here, but since this PR curates the taxonomy, worth reconciling and updating the stale "will add" wording to reflect current state.
- Branch is behind `main` (`mergeStateStatus: BEHIND`), so a rebase/merge is needed before merge.

**What's good:** rename to `push` is consistent across `emit_build_start` / `emit_build_complete` / `emit_agent_deploy` plus the new `emit_capability_use("push")`; `evaluations -> evaluation_validate` matches the "validation runs" semantic; telemetry stays fail-open throughout.

amilandi:
Addressed review comments in f686f2a6:

- **Blocking (dead `onboarding`)** — removed from `ADK_CAPABILITIES` tuple and taxonomy doc block. The onboarding/ skill directory was deleted from `main` by the DA-GA cleanup merge, so the value was declared but never emitted. Kept on `main-ca` (#299) where the skill still ships.
- **Should-fix (subset-only wired test)** — added `test_every_canonical_capability_is_actually_emitted` (reverse of the wired-set check). It scans `scripts/**/*.py` and `src/skills/**/SKILL.md` for shim, direct-call, and `adk_capability=` emit sites and fails on any canonical value that isn't wired. Would have caught this branch's stale `onboarding`.
- **Nit (`discover`)** — PR body updated: dropped the "will be added by #238" phrasing; described as a pre-existing gap outside this PR's scope.
- **Nit (behind main)** — merged origin/main into the branch (0 behind now).

`pytest tests/test_adk_telemetry.py tests/scripts/test_push_scope.py` -> 171 passed.

daeunJe0ng (approved):
[copilot-review] Approve.

Re-reviewed at current head. The prior blocking item (dead `onboarding` capability) is resolved, and the reverse-direction sync test now guards against that class of dead-value drift. Taxonomy split is fail-open, single-source, and matches the push/publish code boundary. No code-side blockers.

One operational follow-up (already noted in the PR body): reconcile any Aria Build Outcomes / Agent Deploy tile still filtered on adk_capability="publishing" so push.py rows aren't dropped after the relabel. Tracking that as a work item.

PR #260, opened 2026-09-03
agentconfig-mcp: add shared configuration foundation

asherjh:
Landing-page and planner capabilities share one safe foundation for workspace MCP configuration and AgentConfiguration API access. This keeps authentication, transport, retries, descriptor rendering, and config preservation consistent across both feature surfaces.

- Adds atomic MCP catalog materialization and contextual server configuration.
- Adds shared AgentConfiguration authentication, transport, retry, and OData helpers.
- Configures Dataverse and ServiceNow through reusable descriptors.
- Runs focused MCP foundation tests on `main` and `release/**` branches.

**Validation**
- `python -m pytest tests\mcp\agentconfig_core tests\scripts\test_mcp_config.py -q` (16 passed)
- `python -m ruff check solutions\ess-maker-skills\scripts\mcp_config.py solutions\ess-maker-skills\src\mcp\agentconfig_core tests\mcp\agentconfig_core tests\scripts\test_mcp_config.py --output-format concise`


jainharsheet77:
## Review notes

Context: `release/planner-landing-page` is currently byte-identical to `main` (SHA `28af0c3`, 0/0 ahead-behind), so this is the first PR on the freshly-cut release branch. I compared the extracted "core" against `feature/landing-page-config` and `feature/planner-skill`.

### ❓ Question — was it intentional to leave planner's `base_client.py` changes out of the shared core?

The core here is byte-identical to `feature/landing-page-config`, but `feature/planner-skill` has already evolved the **same** shared file (+34/−15) with two changes this PR omits:

1. **`_format_api_error` extension hook** — planner's `PlannerClient` overrides it to surface `Target` + the nested `Details[]` validation entries from the AgentConfiguration error envelope. This PR keeps that error-formatting inlined inside `_request`, so the override point doesn't exist. Was the richer error surface deliberately deferred, or just not carried over when the core was cut from the landing-page branch?
2. **stdout → stderr sign-in notice** — both surfaces run over stdio (`mcp.run()`), and the shared interactive MSAL sign-in prints `"Opening browser for AgentConfiguration sign-in …"` to **stdout**, which can corrupt the JSON-RPC stream and disconnect the client. Planner already routes this to `stderr` (`file=sys.stderr, flush=True`). Intentionally left for a later PR?

Framing this as a question because the PR's stated goal is "one safe foundation for **both**," and planner's version is a clean superset (adds only these two things, removes nothing). If the omission is intentional (keep this PR strictly landing-page-equivalent and let the planner PR re-introduce them), that's defensible — but it does mean the planner PR will re-touch the shared core. If unintentional, adopting planner's `base_client.py` now avoids that divergence.

### 🟡 Medium — `materialize-defaults` is currently unreachable on the release branch

`mcp_config.py` ships the `materialize-defaults` command plus the defaults/state reconciliation machinery (~⅓ of the file), but **no `.vscode/mcp.defaults.json` ships and nothing invokes it** in this PR — its callers (`setup.prompt.md`, landing-page/planner `SKILL.md`) live only in the feature branches. Run today it errors `Default MCP catalog not found`. Suggest either shipping the defaults file + a caller, or deferring that command to the PR that needs it. (The wired path — `configure` for Dataverse/ServiceNow — is fine and well-tested.)

### 🟢 Low / confirm

- **CI `pull_request` trigger** lost its `branches:` filter, so the full CI (including the FlightCheck suite) now runs on PRs into *every* base branch, not just `main`/`release/**`. If unintended, mirror the push filter: `branches: ["main", "release/**"]`.
- **`_odata.py` has no importer** in this PR (its consumers are the not-yet-included clients), so it lands with no coverage here. Fine for a foundation PR; a small unit test wouldn't hurt.
- Minor: `tests/scripts/__init__.py` exists but `tests/mcp/**` has none — harmless given unique test basenames.

### ✅ Looks good

- Core files (`_odata.py`, `mcp_config.py`, descriptors, `requirements.txt`, `.gitignore`, connect/onboarding doc edits) are byte-identical to `feature/landing-page-config`; `onboarding/step1.md` correctly drops the landing-page-only pip step.
- `configure` path is well covered (shipped Dataverse + ServiceNow descriptors tested end-to-end); atomic writes, `0600` token cache, `.local/` gitignored, duplicate-JSON-key rejection, https-URL validation.
- New `mcp-foundation` CI job's deps resolve (pytest via `requirements-dev.txt`, httpx/msal via core `requirements.txt`); pyproject `pythonpath` makes `import mcp_config` work.
- The `Authorization` header correctly uses a `Bearer` token — it renders as `******` in diffs only because of secret-masking, not a defect.

asherjh:
This foundation is based on the changes merged into the landing-page branch. Landing page uses the shared `Code: Message` handling. The `_format_api_error` extension hook supports planner's custom formatter, so that extension belongs with the planner integration.

The foundation provides reusable MCP materialization support. PR #261 — landing-page-config: add maker configuration workflow supplies the default-server catalog and setup/landing-page callers that activate the landing-page server. That feature-owned wiring ships in the stacked follow-up.

solutions/ess-maker-skills/src/mcp/agentconfig_core/base_client.py
```diff
@@ -0,0 +1,425 @@
...
+    import msal
+
+    cache = _load_msal_cache()
+    app = msal.PublicClientApplication(
+        _CLIENT_ID,
+        authority=_AUTHORITY,
+        token_cache=cache,
+    )
+
+    accounts = app.get_accounts()
```
ccreight, line 142: I don't have a ton of context on this logic so ignore this comment if it's wrong! but from my review skill:

how do we ensure this cached account belongs to the tenant/environment the maker selected?  accounts[0]  can silently reuse an account from another tenant, so requests may target the wrong tenant without ever showing the account picker.

solutions/ess-maker-skills/src/mcp/agentconfig_core/base_client.py
```diff
@@ -0,0 +1,425 @@
...
+        os.chmod(_LOCAL_STATE_DIR, 0o700)
+    except OSError:
+        pass
+
+    flags = os.O_WRONLY | os.O_CREAT | os.O_TRUNC
+    if hasattr(os, "O_BINARY"):
+        flags |= os.O_BINARY
+    descriptor = os.open(_TOKEN_CACHE_PATH, flags, 0o600)
+    try:
+        with os.fdopen(descriptor, "w", encoding="utf-8") as handle:
```
ccreight: same for this comment, ignore if not relevant in your opinion:

can we make this shared cache write atomic and cross-process safe? landing-page and planner can start simultaneously, and both currently truncate/write the same file without locking, which can lose cache updates or leave an unreadable cache.

ccreight (dismissed):

ccreight (approved):

PR #262, opened 2026-09-04
landing-page-config: add context-grounded widget drafts

asherjh:
## Description

Adds typed, schema-valid suggested drafts for Branding, Quick Links, and Starter Prompts on the landing-page maker workflow. Suggestions are grounded in the target agent's topics, workflows, integrations, knowledge sources, evaluations, current landing-page content, and maker guidance. Makers review unpublished proposals in widgets and publish explicitly; exact deterministic edits use direct updates.

Stacked on [PR 261 — Landing-page maker configuration workflow](https://github.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/pull/261).

- Pass typed draft payloads through landing-page widget openers while preserving server baselines and explicit clears.
- Distinguish omitted drafts, supplied suggestions, and explicit empty lists; route exact clears through direct updates.
- Validate each proposed section against its surface schema, with contrast validation applied to generated Branding colors.
- Add schema, protocol, and skill-orchestration coverage for draft payloads and routing.
- Implements Task 7837117 — Pass typed suggested drafts through landing-page widget openers.
- Implements Task 7837118 — Orchestrate suggested landing-page changes through widget drafts.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update
- [ ] Refactor / cleanup

## Testing

- 2026-09-08: `python -m pytest tests/scripts/test_fetch_and_setup.py tests/mcp/agentconfig/test_skill_orchestration.py tests/mcp/agentconfig/test_widget_protocol.py tests/mcp/agentconfig/test_server_contract.py -q` — 39 passed.
- `python -m pytest tests/mcp/agentconfig/test_client.py tests/mcp/agentconfig/test_server_contract.py tests/mcp/agentconfig/test_widget_protocol.py tests/mcp/agentconfig/test_skill_orchestration.py tests/scripts/test_validate_branding.py tests/scripts/test_setup_preserves_title_id.py tests/setup/test_foundation_checklist.py tests/setup/test_foundation_setup_router.py -q` — 78 passed.
- `python -m pytest tests/scripts/test_installer_launch.py -q -k "InstallerMcpDefaults"` — 2 passed.
- `npm test --silent` and `npm run validate --silent` in `tools/ess-maker-profile/extension` — 40 tests passed and package validation succeeded.

## Checklist

- [x] My code follows the existing style
[...]

PR #269, opened 2026-09-11
[1/5] Share authoring identity and deployed-agent discovery

rebova-microsoft:
## Review boundary: slice 1/5, not a release

**This PR targets `users/rebova/org-announcements-prerelease`, a staging/integration branch for the complete Org Announcements feature. It does not target `main` or a release branch and performs no deployment.** Reviewed slices can accumulate there, but promotion waits for the complete feature, the remaining reviews, and the required integration validation. Merging this foundation slice is not approval to ship Org Announcements.

| Review order | Scope |
| --- | --- |
| #269 | Shared authoring identity, deployed-agent discovery, and supporting regression infrastructure |
| #270 | Announcement models and the agent-scoped authoring client |
| #271 | Account-bound Graph audience discovery |
| #272 | Announcement MCP runtime and coverage across both feature providers |
| #273 | Maker entry points, setup, and packaging |

Each downstream PR targets the preceding review branch, not the prerelease branch directly. Each diff is therefore one review slice against its immediate predecessor. The existing landing-page behavior must remain intact at this stage; the staged destination does not waive correctness of the code in this slice.

## Description

Extract a neutral, read-only deployed-agent discovery client and delegate the existing landing-page lookup tools to it. Expose shared tenant-local account matching and privacy-safe credential checks for the subsequent announcement slices, without adding another MCP process.

The downstream consumers are already present in the published stack: #270 adds `OrgAnnouncementsClient(AgentDiscoveryClient)`, #271 uses the shared identity helpers for Graph, and #272 loads both feature providers and extends the shared discovery coverage to both. The announcement loader entry point in this slice prepares that later coverage; it is not called by this slice, which does not yet contain the announcement runtime.

## Current diff and rationale

- Extract shared deployed-agent discovery and identity helpers while preserving the existing account-bound refresh behavior.
- Keep known credential failures actionable without exposing private details: distinguish missing/unreadable/empty token files, malformed credentials, identity mismatch, cache locking/local access failures, and the locally generated sign-in timeout. Unknown library/provider errors remain generic.
- Put safe remediation on the outer error so it survives plain-tool and widget serialization. Retain underlying exception causes where present; this does not claim that chained tracebacks are scrubbed. Preserve credential-source/account binding, error schemas, and retry behavior.
- Add isolated shared-client and discovery coverage, including safeguards around the existing cache/refresh and stdout behavior. The import helper and auth/network isolation keep the affected provider tests independent of real credentials and services. Auth regressions exercise the registered plain-tool and widget handlers.
- Add narrowly scoped, temporary pull-request target filters for the prerelease branch and the review stack. Existing push targets, PR event types, and workflow permissions are unchanged.

## Feature contract

Discovery returns deployed agent identifiers without initializing feature configuration. Tenant/account context comes from authentication rather than tool arguments. Feature providers retain separate MCP processes while reusing the neutral core.

## Dependency and promotion baseline

Bootstrap prerequisite #262 merged into `release/planner-landing-page` on September 10, 2026. This stack was prepared against `cacb1bec056428809f1ccb0383561190d516bee4`, an ancestor of merge commit `8a04f5f40f334e729a3497877edca655730f1be2`; unchanged prerequisite work is excluded from this slice. The future official release target and its final promotion baseline must be confirmed separately.

## Validation

Local validation used Python 3.13.15; no local Python 3.11 run is claimed. Focused offline regressions cover the shared auth paths and both registered MCP response boundaries.

GitHub Actions uses Python 3.11. [CI run 35395516201](https://github.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/actions/runs/35395516201) tracks the auth-diagnostics follow-up at `b5591731667393d2fae78419efe08cb83db8a7ac`; the PR Checks tab shows the current outcome, including any additional run triggered by a description update. The workflow covers Python lint, installer smoke tests, MCP foundation tests, landing-page configuration, and FlightCheck offline tests.

This does not cover live Graph, authoring-service, hosted-widget, or end-to-end integration; those remain required before rollout.

## Temporary CI support and removal plan

`pull_request.branches` filters the target branch, so the existing `main` / `release/**` rules alone do not cover these stacked PRs. The following entries are temporary:

- `users/rebova/org-announcements-prerelease`
- `users/rebova/org-announcements-review-*`

**Before final release promotion, remove both entries and update the matching assertion in `tests/mcp/agentconfig_core/test_ci_contract.py` in the same change.** Preserve the ordinary branch filters, PR event policy, and read-only workflow permissions. The existing `release/**` filter covers a release target using that naming convention; confirm coverage when the final target is chosen. Keep stack coverage in place until that transition.

## Review readiness for this slice

- [x] Relevant offline tests and static checks pass.
- [x] Python 3.11 CI covers this staged target; current results are linked above and shown in Checks.
- [ ] Maintainer review/sign-off for this slice.

## Separate release-promotion gates

- [ ] Remaining stack reviews and full-feature integration.
- [ ] Authorized live integration and end-to-end validation.
- [ ] Final promotion baseline confirmed and temporary CI filters/assertion cleaned up.

This remains a staged implementation review, not release approval.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,11 @@ on:
   push:
     branches: [ "main", "release/**" ]
   pull_request:
-    branches: [ "main", "release/**" ]
+    branches:
+      - "main"
+      - "release/**"
+      - "users/rebova/org-announcements-prerelease"
+      - "users/rebova/org-announcements-review-*"
     types: [ "opened", "synchronize", "reopened", "edited" ]
 
 permissions:
@@ -103,6 +107,7 @@ jobs:
         python -m pytest
         tests/mcp/agentconfig_core
         tests/scripts/test_mcp_config.py
+        tests/scripts/test_auth.py
         -q
 
   landing-page-config:
--- a/solutions/ess-maker-skills/src/mcp/agentconfig_core/_odata.py
+++ b/solutions/ess-maker-skills/src/mcp/agentconfig_core/_odata.py
@@ -70,6 +70,14 @@ def _escape_odata_literal(value: str, name: str) -> str:
     return _validate_odata_string(value, name).replace("'", "''")
 
 
+def _validate_title_id(title_id: str) -> str:
+    """Validate the opaque EmployeeAgents key shared by authoring surfaces."""
+    _validate_odata_string(title_id, "titleId")
+    if len(title_id) > 256:
+        raise ValueError("titleId must not exceed 256 characters")
+    return title_id
+
+
 def _require_odata_id(value: str, name: str) -> str:
     """Validate a non-empty, control-char-free id and encode it as an OData key."""
     return urllib.parse.quote(_escape_odata_literal(value, name), safe="")
--- a/solutions/ess-maker-skills/src/mcp/agentconfig_core/agent_discovery.py
+++ b/solutions/ess-maker-skills/src/mcp/agentconfig_core/agent_discovery.py
@@ -1,0 +1,81 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""Read-only deployed-agent discovery shared by feature-owned MCP providers."""
+
+from __future__ import annotations
+
+from typing import Any
+
+from base_client import AgentConfigApiError, AgentConfigBaseClient
+
+
+_MAX_SEARCH_LENGTH = 256
+
+
+def _convert_key_case(value: Any, *, upper: bool) -> Any:
+    if isinstance(value, list):
+        return [_convert_key_case(item, upper=upper) for item in value]
+    if not isinstance(value, dict):
+        return value
+    converted: dict[str, Any] = {}
+    for key, item in value.items():
+        if key and key[0].isalpha():
+            first = key[0].upper() if upper else key[0].lower()
+            converted_key = first + key[1:]
+        else:
+            converted_key = key
+        converted[converted_key] = _convert_key_case(item, upper=upper)
+    return converted
+
+
+def _to_api_payload(value: Any) -> Any:
+    return _convert_key_case(value, upper=True)
+
+
+def _to_tool_payload(value: Any) -> Any:
+    return _convert_key_case(value, upper=False)
+
+
+def _unwrap_agent_collection(payload: Any) -> list[dict[str, Any]]:
+    if isinstance(payload, list):
+        return payload
+    if isinstance(payload, dict) and isinstance(payload.get("value"), list):
+        return payload["value"]
+    raise AgentConfigApiError(
+        "AgentConfiguration API returned an invalid collection response"
+    )
+
+
+class AgentDiscoveryClient(AgentConfigBaseClient):
+    """Discover deployed titleIds without initializing any feature configuration.
+
+    Feature clients retain their own base URL, transport, and authentication.
+    Explicit response conversion keeps discovery independent of each feature's
+    canonical payload casing and collection routes.
+    """
+
+    def _agent_collection_path(self) -> str:
+        return f"tenants('{self.tenant_id}')/EmployeeAgents"
+
+    async def list_agent_configs(self) -> list[dict[str, Any]]:
+        payload = await self._request(
+            "GET", self._agent_collection_path(), transform_payload=False
+        )
+        return _unwrap_agent_collection(_to_tool_payload(payload))
+
+    async def search_agents(self, search_string: str) -> list[dict[str, Any]]:
+        if not isinstance(search_string, str) or not search_string.strip():
+            raise ValueError("searchString must be a non-empty string")
+        normalized = search_string.strip()
+        if len(normalized) > _MAX_SEARCH_LENGTH:
+            raise ValueError(
+                f"searchString must not exceed {_MAX_SEARCH_LENGTH} characters"
+            )
+        payload = await self._request(
+            "POST",
+            f"{self._agent_collection_path()}/SearchAgents",
+            json={"SearchString": normalized},
+            transform_payload=False,
+        )
+        return _unwrap_agent_collection(_to_tool_payload(payload))
--- a/solutions/ess-maker-skills/src/mcp/agentconfig_core/base_client.py
+++ b/solutions/ess-maker-skills/src/mcp/agentconfig_core/base_client.py
@@ -42,6 +42,7 @@ import webbrowser
 from typing import Any, Optional
 
 import httpx
+from portalocker.exceptions import LockException
 
 from _tenant_context import configured_tenant_id
 from _token_cache import create_token_cache
@@ -85,15 +86,11 @@ def _resolve_token() -> str:
 
 def _read_token_file(token_file: str) -> str:
     if not os.path.isfile(token_file):
-        raise ValueError(
-            f"AGENTCONFIG_ACCESS_TOKEN_FILE={token_file!r} does not exist"
-        )
+        raise ValueError("AGENTCONFIG_ACCESS_TOKEN_FILE does not exist")
     with open(token_file, "r", encoding="utf-8") as handle:
         token = handle.read().strip()
     if not token:
-        raise ValueError(
-            f"AGENTCONFIG_ACCESS_TOKEN_FILE={token_file!r} is empty"
-        )
+        raise ValueError("AGENTCONFIG_ACCESS_TOKEN_FILE is empty")
     return token
 
 
@@ -157,18 +154,19 @@ def acquire_token_msal_interactive(expected_tenant_id: str | None = None) -> str
         result = _acquire_token_interactive_form_post(app)
 
     if "access_token" not in result:
-        error = result.get("error", "unknown_error")
-        description = result.get("error_description", "")
-        raise ValueError(f"MSAL sign-in failed ({error}): {description}")
+        raise ValueError("AgentConfiguration sign-in failed. Sign in again and retry.")
     return result["access_token"]
 
 
-def _refresh_msal_token(tenant_id: str, object_id: str) -> str:
+def account_for_identity(app: Any, tenant_id: str, object_id: str) -> Any | None:
+    """Find one MSAL home account through its tenant-local profile.
+
+    get_accounts() groups tenant profiles by home account. Its exposed local id
+    may belong to another tenant (notably for guests), so match the original
+    tenant profile in the shared cache before selecting a grouped account.
+    """
     import msal
 
-    app = _create_msal_app(tenant_id)
-    # get_accounts() groups profiles by home account and can expose another
-    # tenant's local id. Resolve the original tenant profile through the cache.
     home_account_ids = {
         account["home_account_id"]
         for account in app.token_cache.search(msal.TokenCache.CredentialType.ACCOUNT)
@@ -182,13 +180,19 @@ def _refresh_msal_token(tenant_id: str, object_id: str) -> str:
         account for account in app.get_accounts()
         if account.get("home_account_id") in home_account_ids
     ]
-    if len(accounts) != 1:
+    return accounts[0] if len(accounts) == 1 else None
+
+
+def _refresh_msal_token(tenant_id: str, object_id: str) -> str:
+    app = _create_msal_app(tenant_id)
+    account = account_for_identity(app, tenant_id, object_id)
+    if account is None:
         raise AgentConfigApiError(
             "The original account is unavailable for token refresh. "
             "Sign in again as that account and retry.",
             http_status=401,
         )
-    result = app.acquire_token_silent(_SCOPE, account=accounts[0], force_refresh=True)
+    result = app.acquire_token_silent(_SCOPE, account=account, force_refresh=True)
     if not result or not result.get("access_token"):
         raise AgentConfigApiError(
             "Could not refresh the access token for the original account. "
@@ -250,11 +254,14 @@ def _decode_jwt_payload(token: str) -> dict[str, Any]:
     payload_segment = parts[1]
     padded = payload_segment + "=" * (-len(payload_segment) % 4)
     try:
-        return json.loads(base64.urlsafe_b64decode(padded))
+        payload = json.loads(base64.urlsafe_b64decode(padded))
     except (binascii.Error, json.JSONDecodeError, UnicodeDecodeError) as error:
         raise ValueError(
-            f"Could not decode AGENTCONFIG_ACCESS_TOKEN payload: {error}"
+            "Could not decode AGENTCONFIG_ACCESS_TOKEN payload"
         ) from error
+    if not isinstance(payload, dict):
+        raise ValueError("AGENTCONFIG_ACCESS_TOKEN payload must be an object")
+    return payload
 
 
 def _decode_tenant_id_from_jwt(token: str) -> str:
@@ -285,6 +292,17 @@ def _decode_object_id_from_jwt(token: str) -> Optional[str]:
     if isinstance(object_id, str) and object_id:
         return object_id
     return None
+
+
+def validate_token_identity(token: str, tenant_id: str, object_id: str | None) -> str:
+    """Check resource-token context; services still validate and authorize tokens."""
+    _validate_token_tenant(token, tenant_id)
+    if not object_id:
+        raise ValueError("The authoring account cannot be identified without an oid claim")
+    actual = _decode_object_id_from_jwt(token)
+    if actual is None or actual.casefold() != object_id.casefold():
+        raise ValueError("The token does not identify the intended authoring account")
+    return token
 
 
 class AgentConfigBaseClient:
@@ -326,6 +344,11 @@ class AgentConfigBaseClient:
             f"tenant_id={self.tenant_id!r}>"
         )
 
+    @property
+    def object_id(self) -> str | None:
+        """Captured tenant-local principal context, never a tool argument."""
+        return self._object_id
+
     def _transform_response(self, payload: Any) -> Any:
         """Surface-specific response key transform; identity in the neutral core.
 
@@ -360,7 +383,7 @@ class AgentConfigBaseClient:
         self._client = None
 
     def _acquire_replacement_token(self) -> str:
-        if self._object_id is None:
+        if self.object_id is None:
             raise AgentConfigApiError(
                 "The original account cannot be identified for token refresh. "
                 "Provide a token with an object id and recreate the client.",
@@ -377,23 +400,21 @@ class AgentConfigBaseClient:
                     http_status=401,
                 )
         else:
-            token = _refresh_msal_token(self.tenant_id, self._object_id)
-
-        _validate_token_tenant(token, self.tenant_id)
-        object_id = _decode_object_id_from_jwt(token)
-        if object_id is None or object_id.casefold() != self._object_id.casefold():
-            raise AgentConfigApiError(
-                "The replacement token does not identify the original account. "
-                "Provide a matching token and retry.",
-                http_status=401,
-            )
-        return token
+            token = _refresh_msal_token(self.tenant_id, self.object_id)
+        return validate_token_identity(token, self.tenant_id, self.object_id)
 
     async def _refresh_token(self, rejected_token: str) -> None:
         async with self._token_lock:
             if self._token != rejected_token:
                 return
-            token = await asyncio.to_thread(self._acquire_replacement_token)
+            try:
+                token = await asyncio.to_thread(self._acquire_replacement_token)
+            except (ValueError, LockException, OSError) as error:
+                raise AgentConfigApiError(
+                    "Could not renew credentials for the original account. "
+                    "Provide matching credentials and retry.",
+                    http_status=401,
+                ) from error
             if token == rejected_token:
                 raise AgentConfigApiError(
                     "No replacement access token is available. "
--- a/solutions/ess-maker-skills/src/mcp/agentconfig_landing_page/client.py
+++ b/solutions/ess-maker-skills/src/mcp/agentconfig_landing_page/client.py
@@ -28,52 +28,20 @@ sys.path.insert(
 from _odata import (  # noqa: E402
     _require_odata_id,
     _validate_https_base_url,
-    _validate_odata_string,
+    _validate_title_id,
 )
-from base_client import AgentConfigApiError, AgentConfigBaseClient  # noqa: E402
+from agent_discovery import (  # noqa: E402
+    AgentDiscoveryClient,
+    _to_api_payload,
+    _to_tool_payload,
+)
+from base_client import AgentConfigApiError as AgentConfigApiError  # noqa: E402
 
 
 DEFAULT_AGENTCONFIG_BASE_URL = "https://substrate.office.com/weveb2/api/v1.1"
-_MAX_TITLE_ID_LENGTH = 256
-_MAX_SEARCH_LENGTH = 256
 
 
-def _validate_title_id(title_id: str) -> str:
-    _validate_odata_string(title_id, "titleId")
-    if len(title_id) > _MAX_TITLE_ID_LENGTH:
-        raise ValueError(
-            f"titleId must not exceed {_MAX_TITLE_ID_LENGTH} characters"
-        )
-    return title_id
-
-
-def _convert_key_case(value: Any, *, upper: bool) -> Any:
-    """Recursively convert the first character of JSON object keys."""
-    if isinstance(value, list):
-        return [_convert_key_case(item, upper=upper) for item in value]
-    if not isinstance(value, dict):
-        return value
-
-    converted: dict[str, Any] = {}
-    for key, item in value.items():
-        if key and key[0].isalpha():
-            first = key[0].upper() if upper else key[0].lower()
-            converted_key = first + key[1:]
-        else:
-            converted_key = key
-        converted[converted_key] = _convert_key_case(item, upper=upper)
-    return converted
-
-
-def _to_api_payload(value: Any) -> Any:
-    return _convert_key_case(value, upper=True)
-
-
-def _to_tool_payload(value: Any) -> Any:
-    return _convert_key_case(value, upper=False)
-
-
-class AgentConfigClient(AgentConfigBaseClient):
+class AgentConfigClient(AgentDiscoveryClient):
     """Async client for production EmployeeAgents list/search/create/get/PATCH.
 
     Inherits auth, token decode, the httpx session, and the retrying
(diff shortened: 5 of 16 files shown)


rebova-microsoft:
Review order: #269 (shared core/discovery) → #270 (contracts/client) → #271 (Graph audience) → #272 (MCP runtime) → #273 (maker/setup/packaging).

Each PR targets the preceding review branch; #269 targets `users/rebova/org-announcements-prerelease`. The prerelease branch remains at the shared baseline until the chunks are reviewed and merged. The complete local feature is at the #273 head.

Bootstrap PR #262 has merged, and the pinned base is in the release branch ancestry. Promotion to the future official release branch and `main` is separate. Paired widget/backend integration, live testing, and end-to-end validation remain outstanding; these draft PRs are not release approval.

rebova-microsoft:

rebova-microsoft (commented):
## File-by-file walkthrough

These 16 explanatory comments describe what each changed file does and why it belongs in this shared-foundation slice. They are reading aids, not an approval or a request for changes, and do not resolve the existing review feedback.

Scope: 4 runtime files, 1 CI file, and 11 test/support files. Of the 929 added lines, 774 are under `tests/`. This PR extracts shared discovery and identity primitives; it does not yet add the announcement provider or widget.

Suggested reading order: `agent_discovery.py` alongside the landing-page client deletions, then `base_client.py` with its regression cases, then the test-import support and CI changes.

solutions/ess-maker-skills/src/mcp/agentconfig_core/_odata.py
```diff
@@ -70,6 +70,14 @@ def _escape_odata_literal(value: str, name: str) -> str:
     return _validate_odata_string(value, name).replace("'", "''")
 
 
+def _validate_title_id(title_id: str) -> str:
```
rebova-microsoft, line 73: **Walkthrough — shared `titleId` validation.**

This moves the existing landing-page title-ID rule into the shared helpers: reuse the established string validation and reject identifiers longer than 256 characters. `titleId` stays an opaque deployed-agent identifier, not a GUID requirement.

The purpose is one rule for both feature clients. This validator does not replace OData key escaping; callers still pass validated IDs through the existing encoding helper when constructing an agent URL.

solutions/ess-maker-skills/src/mcp/agentconfig_core/agent_discovery.py
```diff
@@ -0,0 +1,81 @@
...
+    if isinstance(payload, list):
+        return payload
+    if isinstance(payload, dict) and isinstance(payload.get("value"), list):
+        return payload["value"]
+    raise AgentConfigApiError(
+        "AgentConfiguration API returned an invalid collection response"
+    )
+
+
+class AgentDiscoveryClient(AgentConfigBaseClient):
```
rebova-microsoft, line 50: **Walkthrough — extract discovery once, without adding a server.**

This new Python client contains the agent-list/search behavior moved out of the landing-page client, along with the existing payload-casing helpers and collection unwrapping. It inherits authentication, HTTP transport, and retries from `AgentConfigBaseClient`.

The operations remain `GET tenants('{tenantId}')/EmployeeAgents` and `POST .../SearchAgents`. Search trims its input and rejects empty or overlong strings. Responses may be a bare list or `{value: [...]}`; an unexpected shape raises an error rather than becoming an empty success.

The separate `_agent_collection_path` and explicit discovery conversion keep lookup independent of a feature's own collection route and canonical payload format. Landing-page consumes this shared code in this slice; announcements consumes it later. No third MCP process or configuration-initialization write is introduced.

solutions/ess-maker-skills/src/mcp/agentconfig_landing_page/client.py
```diff
@@ -28,52 +28,20 @@
...
-def _to_api_payload(value: Any) -> Any:
-    return _convert_key_case(value, upper=True)
 
-
-def _to_tool_payload(value: Any) -> Any:
-    return _convert_key_case(value, upper=False)
-
-
-class AgentConfigClient(AgentConfigBaseClient):
+class AgentConfigClient(AgentDiscoveryClient):
```
rebova-microsoft, line 44: **Walkthrough — the deleted discovery code moved to the shared layer.**

`AgentConfigClient` now inherits from `AgentDiscoveryClient`, which itself inherits from `AgentConfigBaseClient`. This removes local copies of list/search, key-casing helpers, and title-ID validation; it does not remove those capabilities.

Landing-page-specific create/get/update operations stay here. The client retains its base URL and normal payload conversion, and `_collection_path()` delegates to the shared agent-collection path.

Read this deletion alongside `agentconfig_core/agent_discovery.py`: the intent is to preserve existing landing-page behavior while allowing a second feature provider to reuse discovery without calling the landing-page MCP process.

solutions/ess-maker-skills/src/mcp/agentconfig_core/base_client.py
```diff
@@ -157,18 +154,19 @@ def acquire_token_msal_interactive(expected_tenant_id: str | None = None) -> str
...
     if "access_token" not in result:
-        error = result.get("error", "unknown_error")
-        description = result.get("error_description", "")
-        raise ValueError(f"MSAL sign-in failed ({error}): {description}")
+        raise ValueError("AgentConfiguration sign-in failed. Sign in again and retry.")
     return result["access_token"]
 
 
-def _refresh_msal_token(tenant_id: str, object_id: str) -> str:
+def account_for_identity(app: Any, tenant_id: str, object_id: str) -> Any | None:
```
rebova-microsoft, line 161: **Walkthrough — reusable identity checks and controlled credential failures.**

This is the main behavioral file in the slice; it is not a new login/token-cache implementation.

- `account_for_identity` extracts the tenant-profile matching already used by token refresh. MSAL groups profiles by home account, so matching the tenant-local cached profile matters for identities such as guest accounts. The helper returns an account only when the match is unambiguous.
- `object_id` exposes the account captured by the authoring client. `validate_token_identity` centralizes the tenant/account comparison so later resource clients can reuse it. Backend token validation and authorization remain authoritative.
- Decoded token payloads must be JSON objects; malformed payloads are rejected explicitly.
- Missing/empty token-file messages and MSAL failure messages stop echoing private paths or raw provider details. During refresh, validation, cache-lock, and filesystem failures become a controlled 401 `AgentConfigApiError`, retaining the original exception as its cause.

The account-preserving refresh mechanism was already upstream. This slice extracts its matching/checking logic and changes error handling. The separate existing review thread about an unreadable token file during initial construction remains open; this walkthrough does not claim that path is fixed.

.github/workflows/ci.yml
```diff
@@ -7,7 +7,11 @@ on:
   push:
     branches: [ "main", "release/**" ]
   pull_request:
-    branches: [ "main", "release/**" ]
+    branches:
+      - "main"
+      - "release/**"
+      - "users/rebova/org-announcements-prerelease"
```
rebova-microsoft, line 13: **Walkthrough — run CI for this review stack, without opening every feature branch.**

The PR-target filter adds exactly our prerelease branch and `users/rebova/org-announcements-review-*`. Without those entries, the stacked PRs would miss CI because they target each other rather than `main` or `release/**`.

The other change in this file adds the existing `tests/scripts/test_auth.py` to the foundation job because authentication is a shared dependency. Existing push filters, PR event types, and permissions are retained. The announcement-specific CI job is not part of this first slice; it arrives with the announcement modules in subsequent chunks.

tests/mcp/_mcp_modules.py
```diff
@@ -0,0 +1,132 @@
...
+    "drafts",
+    "graph_directory_client",
+    "telemetry",
+    "server",
+)
+
+_UNSET = object()
+
+
+def load_mcp_modules(
```
rebova-microsoft, line 59: **Walkthrough — test-only isolation for flat Python server modules.**

Both provider folders use names such as `client.py`, `drafts.py`, and `server.py`. Separate MCP processes can import their own siblings normally, but a combined pytest process can otherwise reuse the first module cached as `client` or `server` for the wrong provider.

This helper loads modules from explicit paths under provider-specific aliases, temporarily makes their sibling imports resolve, and restores the previous plain-name entries in `sys.modules` afterward. Cached aliases let tests that monkeypatch the same provider share its module objects. Production imports and the global pytest import mode are unchanged.

The announcement loader is defined for later slices but is not called before those files exist. This is the main test-support complexity to inspect. Warning suppression is currently broad during module execution; the existing review thread asking for a narrower warning-category filter is separate from this explanation.

tests/mcp/conftest.py
```diff
@@ -0,0 +1,49 @@
...
+    Path(__file__).resolve().parents[2]
+    / "solutions" / "ess-maker-skills" / "src" / "mcp" / "agentconfig_core"
+)
+sys.path.insert(0, str(CORE_DIR))
+
+import _tenant_context  # noqa: E402
+
+
+@pytest.fixture(autouse=True)
+def isolate_mcp_identity(monkeypatch, tmp_path) -> None:
```
rebova-microsoft, line 27: **Walkthrough — keep these tests independent of the developer's identity and environment.**

This automatic pytest fixture redirects maker configuration to a temporary test path and removes credential environment overrides. It also makes unmocked tenant discovery, MSAL client creation, and HTTP transport fail explicitly.

Tests must therefore provide their own fake identity and transport instead of accidentally reading local maker configuration, signing in, or contacting a real service. Monkeypatching is restored after each test. This changes test isolation, not production authentication behavior.

tests/mcp/agentconfig/test_client.py
```diff
@@ -23,9 +23,15 @@
...
-sys.path.insert(0, str(AGENTCONFIG_DIR))
+# Sibling MCP servers share the top-level names ``client``/``server``, so the
+# modules are loaded through the shared isolated importer rather than by a plain
+# ``import`` off ``sys.path``. See tests/mcp/_mcp_modules.py.
+sys.path.insert(0, str(Path(__file__).resolve().parents[1]))
 
-import client as agentconfig_client  # noqa: E402
+from _mcp_modules import load_landing_page_modules  # noqa: E402
+
+_LANDING_MODULES = load_landing_page_modules()
```
rebova-microsoft, line 33: **Walkthrough — preserve the landing-page client suite while changing how it imports the client.**

The only substantive change here replaces `import client` with `load_landing_page_modules()["client"]`. Existing test assertions remain in place.

This makes the suite reliably exercise the landing-page client even when another provider's `client.py` is loaded in the same pytest process. The surrounding explanation points to the shared test loader; it is not a production client rewrite.

tests/mcp/agentconfig/test_server_contract.py
```diff
@@ -32,11 +29,15 @@
...
+# modules are loaded through the shared isolated importer rather than by a plain
+# ``import`` off ``sys.path``. See tests/mcp/_mcp_modules.py.
+sys.path.insert(0, str(Path(__file__).resolve().parents[1]))
+
+from _mcp_modules import load_landing_page_modules  # noqa: E402
 
-with warnings.catch_warnings():
-    warnings.simplefilter("ignore", IncompleteFieldDefinitionWarning)
-    import server as agentconfig_server  # noqa: E402
+_LANDING_MODULES = load_landing_page_modules()
```
rebova-microsoft, line 39: **Walkthrough — keep tool-contract assertions attached to the correct provider.**

The suite now obtains the landing-page server through the isolated loader instead of importing the plain top-level name `server`. Existing contract checks remain; the local import/warning wrapper is removed because module loading is handled centrally.

This prevents collection order from making the landing-page suite inspect another feature's server once the announcement runtime joins the stack. The centralized warning-filter behavior is covered by the separate review discussion on `_mcp_modules.py`.

tests/mcp/agentconfig/test_widget_protocol.py
```diff
@@ -26,11 +24,15 @@
...
+# modules are loaded through the shared isolated importer rather than by a plain
+# ``import`` off ``sys.path``. See tests/mcp/_mcp_modules.py.
+sys.path.insert(0, str(Path(__file__).resolve().parents[1]))
 
-with warnings.catch_warnings():
-    warnings.simplefilter("ignore", IncompleteFieldDefinitionWarning)
-    import server as agentconfig_server  # noqa: E402
+from _mcp_modules import load_landing_page_modules  # noqa: E402
+
+_LANDING_MODULES = load_landing_page_modules()
```
rebova-microsoft, line 34: **Walkthrough — same landing-page widget checks, isolated server import.**

This makes the same test-loader adaptation as the other landing-page suites: replace the direct `server` import and local warning wrapper with the explicitly loaded landing-page module.

The widget protocol assertions themselves are retained. The reason this file belongs in the foundation slice is to keep those tests running against the intended provider when multiple flat MCP server folders are imported together; it does not change a widget's protocol or UI.

tests/mcp/agentconfig_core/test_agent_discovery.py
```diff
@@ -0,0 +1,138 @@
...
+]
+
+
+@pytest.fixture
+def token(monkeypatch):
+    encoded = base64.urlsafe_b64encode(json.dumps({"tid": TENANT}).encode()).decode()
+    monkeypatch.setenv("AGENTCONFIG_ACCESS_TOKEN", f"header.{encoded}.signature")
+
+
+def test_both_features_inherit_the_same_discovery_implementation():
```
rebova-microsoft, line 40: **Walkthrough — protect behavior while discovery moves out of the landing-page client.**

The new suite checks inherited methods, list/search request routes and payload casing, bare versus wrapped collections, invalid queries before transport, malformed-response errors, and provider-tool delegation without initializing configuration.

A slice-boundary detail: `CLIENTS` currently contains only the landing-page client, despite this test's forward-looking "both features" name. The announcement provider does not exist in PR #269; its consumer coverage joins later. That keeps this commit independently runnable instead of relying on future files.

tests/mcp/agentconfig_core/test_base_client.py
```diff
@@ -108,6 +107,65 @@ def acquire_token_by_auth_code_flow(self, flow, response):
     )
 
 
+def test_the_sign_in_notice_goes_to_stderr_not_stdout(monkeypatch) -> None:
```
rebova-microsoft, line 110: **Walkthrough — why this is the largest added test block.**

The expanded cases cover shared behavior both feature clients rely on: credential-source precedence, configured-tenant matching before a request, silent refresh of the original identity, rejected/mismatched replacement credentials, concurrent 401s sharing one refresh, privacy-safe surfaced failures, and checkout-anchored imports.

The sign-in tests also ensure human-readable browser notices go to stderr rather than corrupting MCP's JSON-RPC stdout stream. Tests use fake servers, accounts, tokens, and HTTP transports.

Much of this protects behavior already in the upstream foundation; the line count does not imply that this PR adds hundreds of lines of authentication runtime. The new assertions complement the account-helper extraction and error normalization in `base_client.py`.

tests/mcp/agentconfig_core/test_ci_contract.py
```diff
@@ -0,0 +1,26 @@
...
+# Licensed under the MIT License.
+
+"""Keep review-stack CI support narrow without changing upstream event policy."""
+
+from pathlib import Path
+
+import yaml
+
+
+def test_ci_accepts_only_the_approved_review_targets_in_addition_to_upstream():
```
rebova-microsoft, line 11: **Walkthrough — make the narrowly approved CI boundary explicit.**

This test asserts the exact push and PR-target branch lists, the existing PR event types, and `contents: read` permissions. Its purpose is to catch accidental broadening of repository-wide CI policy while adding support for this particular stack.

It uses `yaml.BaseLoader` so the workflow key `on` remains a string rather than being interpreted as a YAML 1.1 boolean. These assertions are intentionally policy-specific; a deliberate future policy change should update them alongside the workflow.

tests/mcp/agentconfig_core/test_tenant_context.py
```diff
@@ -147,3 +149,71 @@ def test_invalid_config_encoding_warns_and_falls_back(config_path, caplog) -> No
 
     assert _tenant_context.configured_tenant_id() is None
     assert "configuration could not be read" in caplog.text
+
+
+@pytest.mark.parametrize("launch_folder", ["agentconfig_landing_page"])
+def test_discovers_configured_environment_from_feature_launch_directory(
```
rebova-microsoft, line 155: **Walkthrough — tenant discovery must work from a feature's launch directory.**

This extends coverage for running from the landing-page MCP directory, parsing mocked Dataverse authentication challenges into a concrete tenant, and ensuring runtime installation paths include the shared dependencies.

The earlier configuration-path assertion is adjusted because the new automatic fixture deliberately redirects `_CONFIG_PATH` into a temporary location. The test still verifies the solution root and imported `auth.py` belong to this checkout, not a sibling worktree.

Only the landing-page launch directory is exercised in this first slice; future announcement files are not required. The tenant-discovery implementation and dependency declarations already exist in the upstream baseline and are not rewritten here.

tests/mcp/agentconfig_core/test_token_cache.py
```diff
@@ -546,3 +546,4 @@ def test_cache_and_directory_have_private_posix_permissions(tmp_path: Path) -> N
 
     assert stat.S_IMODE(directory.stat().st_mode) == 0o700
     assert stat.S_IMODE(path.stat().st_mode) == 0o600
+    assert stat.S_IMODE(Path(str(path) + ".lockfile").stat().st_mode) == 0o600
```
rebova-microsoft, line 549: **Walkthrough — one additional permissions assertion.**

The existing POSIX permissions test already checks the cache directory (`0700`) and cache file (`0600`). This line extends the check to the synchronization lock file, also requiring `0600`.

The token-cache implementation is unchanged in this PR. This is a small extension of the existing private-file coverage, not a new cache design.

tests/mcp/agentconfig_core/test_token_refresh.py
```diff
@@ -414,8 +414,11 @@ def test_unavailable_original_file_does_not_fall_back_to_another_source(
 
     async def run():
         try:
-            with pytest.raises(ValueError, match="does not exist"):
+            with pytest.raises(base_client.AgentConfigApiError) as caught:
```
rebova-microsoft, line 417: **Walkthrough — match the controlled refresh-error contract.**

When the original token file disappears, refresh now surfaces a 401 `AgentConfigApiError` rather than a raw `ValueError`. The updated assertion checks the HTTP status, retains the underlying `ValueError` as the cause, and confirms the private file path is absent from the surfaced message.

The existing scenario still guards against silently switching from that original file to a different credential source. This test change follows the normalization in `_refresh_token`; it does not relax identity preservation or assert that initial token-file construction errors use the same path.

tests/mcp/_mcp_modules.py
```diff
@@ -0,0 +1,132 @@
...
+                sys.modules[name] = module
+                try:
+                    with warnings.catch_warnings():
+                        # Importing a FastMCP server transitively triggers
+                        # pydantic-settings' IncompleteFieldDefinitionWarning,
+                        # which the suite's ``filterwarnings = error`` would
+                        # otherwise turn into a collection failure. Suppressed
+                        # here, once, instead of at every call site.
+                        warnings.simplefilter("ignore")
+                        spec.loader.exec_module(module)
```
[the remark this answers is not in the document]
rebova-microsoft: Addressed in c1afef5c390bb3e1edb951a93d1e8b3195a0c514. The loader now ignores only `pydantic_settings.exceptions.IncompleteFieldDefinitionWarning` within its existing `catch_warnings` context. Unrelated categories retain the caller’s warning policy.

Added focused loader regressions for the expected category, `UserWarning`, and `RuntimeWarning` under both recording and warnings-as-errors policies. They also cover cached loads, cleanup after failure, restoration of plain module bindings, and restoration of the caller’s warning filters. CI for the follow-up commit is green.

solutions/ess-maker-skills/src/mcp/agentconfig_core/base_client.py
```diff
@@ -85,15 +86,11 @@ def _resolve_token() -> str:
 
 def _read_token_file(token_file: str) -> str:
     if not os.path.isfile(token_file):
-        raise ValueError(
-            f"AGENTCONFIG_ACCESS_TOKEN_FILE={token_file!r} does not exist"
-        )
+        raise ValueError("AGENTCONFIG_ACCESS_TOKEN_FILE does not exist")
     with open(token_file, "r", encoding="utf-8") as handle:
         token = handle.read().strip()
```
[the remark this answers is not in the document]
rebova-microsoft: Addressed in c1afef5c390bb3e1edb951a93d1e8b3195a0c514. `_read_token_file` now catches `OSError` specifically around open/read and raises a fixed, path-free `ValueError`, preserving the original exception as its cause. The returned message does not expose the configured path; the retained cause is still available for debugging.

Regression coverage exercises permission/read failures and disappearance after the existence check through the helper, token resolution, and initial client construction. It confirms no fallback to an environment token or MSAL, no request sent, and unchanged successful/missing/empty-file behavior. CI for the follow-up commit is green.

.github/workflows/ci.yml
dcowen91, line 13: I don't think we want to add user branches for CI.

apurvabanka:
**Blocking**



1. **Dead, broken forward reference.** `tests/mcp/_mcp_modules.py` ships `ORG_ANNOUNCEMENTS_MODULES` + `load_org_announcements_modules()`, but `src/mcp/agentconfig_org_announcements/` doesn't exist on this branch. Calling it raises `FileNotFoundError`. Nothing references it. Its whole docstring justifies fixing a `client`/`server` name collision that cannot occur in this slice — only one server exists.

2. **CI branch allow-list pinned by a test.** `ci.yml` hardcodes `users/rebova/org-announcements-prerelease` and `users/rebova/org-announcements-review-*`, and `test_ci_contract.py` *asserts* that exact list. Once the stack lands and those branches go away, removing them fails the test. No TODO or removal plan.



**Design / layering**



1. **The "shared" core has no second consumer.** `AgentDiscoveryClient`, `account_for_identity`, `validate_token_identity`, and the `object_id` property are used only by the landing page (or internally). `AgentDiscoveryClient` also has no `__init__`, so it can't be instantiated standalone — the sharing claim is untested. `CLIENTS = [...]` and `parametrize("modules", [LANDING])` in `test_agent_discovery.py` are single-element scaffolding for slice 2.

2. **Foundation job now depends on a feature server.** `test_agent_discovery.py` / `test_mcp_modules.py` live under `tests/mcp/agentconfig_core/` but import the landing-page FastMCP server. They only work because `requirements-dev.txt` happens to carry `mcp`; the job installs only `agentconfig_core/requirements.txt`.

3. **New hard dep on an extra.** `base_client.py` now does `from portalocker.exceptions import LockException` at module top. `portalocker` is only a `msal-extensions[portalocker]` extra — if that extra ever drops, *every* MCP surface fails at import, not just refresh.



**Diagnostics**



1. **Privacy stripping removes actionable detail with no replacement.** MSAL `error`/`error_description`, the JWT decode cause, and the token-file path are all gone, and `_refresh_token` now collapses `ValueError`/`OSError`/`LockException` into one generic 401. A missing token file during refresh now reads as an auth failure. This is a local developer tool where the path is the maker's own — consider `logger.debug` for the detail.



**Scope**



1. Title says "share identity and discovery," but the diff also bundles error-message hardening, token-file `OSError` handling, a lockfile-permission assertion, stdout/stderr print guards (testing code already on base), and network isolation in a new `conftest.py`. Several are independently reviewable.



**Minor:** `_mcp_modules` gets loaded under two identities (`_mcp_modules` via `sys.path` vs `tests.mcp._mcp_modules` in `test_mcp_modules.py`); on partial load failure the loader pops only the failing alias, leaving earlier ones cached; `_MAX_TITLE_ID_LENGTH` became a bare `256` in `_odata.py`; `validate_token_identity`'s `if not object_id` branch is unreachable in-slice.

CC: @rebova-microsoft

.github/workflows/ci.yml
rebova-microsoft: Agreed that these user-branch entries should not ship in the final release. They are temporary support for reviewing this stack: #269 targets `users/rebova/org-announcements-prerelease`, and the downstream PRs target the preceding review branch rather than `main` or `release/**`.

The `pull_request.branches` filter matches the PR target branch, so the existing filters alone would skip CI for these PRs. The `push` trigger is unchanged and still restricted to `main` and `release/**`.

I also confirmed this is working: [CI run 34908684999](https://github.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/actions/runs/34908684999) ran on September 14 as a `pull_request` event for #269, associated with the current head `c1afef5c` and targeting the prerelease branch. All five CI jobs executed and passed, including their actual lint/test steps; these are not inherited checks from another branch.

I will remove both temporary user-branch filters before promoting the changes to the final release branch. Once the PRs target a `release/**` branch, the existing release filter covers them.

rebova-microsoft:
Thanks @apurvabanka. I have updated the description to make the staging boundary and cleanup plan much more explicit.

**#269 is slice 1 of a five-PR stack, merging only into `users/rebova/org-announcements-prerelease`.** That branch collects the reviewed pieces; it is not promoted to the final release branch or `main` until the complete feature and required integration validation are ready. This does not excuse a regression in the existing landing-page behavior, but it changes how to interpret groundwork whose consumers arrive in later slices.

Specific pointers for the forward-reference/sharing questions:

- [#270](https://github.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/pull/270) contains `OrgAnnouncementsClient(AgentDiscoveryClient)`.
- [#271](https://github.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/pull/271) uses `account_for_identity` and `validate_token_identity` in the Graph client.
- [#272](https://github.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/pull/272) calls the announcement loader and extends `test_agent_discovery.py` to both clients/providers. In #269 the announcement loader entry point is unused and its target directory is absent; invoking it at this stage would indeed fail. It is preparation for that published downstream slice, not a claim that the announcement runtime is already available here.
- `AgentDiscoveryClient` inherits `AgentConfigBaseClient.__init__`, including the required `base_url` and `logger_name` arguments. Not defining a separate constructor does not prevent instantiation.

**CI cleanup:** agreed that changing only the workflow would leave a failing assertion. Before final release promotion, I will remove both temporary user-branch target filters and update `test_ci_contract.py` together, retaining the existing event/permissions safeguards. That commitment is now in the description. The filters remain while the PRs target the staged branches; [run 34908684999](https://github.com/microsoft/Employee-Self-Service-Agent-Developer-Kit/actions/runs/34908684999) confirms CI runs for #269's current head against the prerelease target.

**Dependencies/layering:** the foundation job installs both `requirements-dev.txt` and the core requirements; `mcp` is explicitly in the former. The discovery regression does exercise the landing-page server, so separating core-only and provider-contract coverage is a layering discussion, not an absent CI dependency. `test_mcp_modules.py` exercises the loader with synthetic modules rather than importing the landing-page server. The core requirements also explicitly request `msal-extensions[portalocker]`; the currently used 1.3.1 package declares the `portalocker` dependency for that extra.

**Diagnostics and remaining observations:** the JWT decode, token-file read exception, and normalized refresh paths retain chained causes, while their returned messages are sanitized. The MSAL failure message is coarser, so I am not claiming all diagnostic detail is preserved. The diagnostics tradeoff, loader import identities/partial-load caching, and remaining minor observations still merit individual triage; the staging explanation does not resolve them. I am not proposing to restore raw authentication/provider payloads or treating those concerns as closed.

The description also now explains why the shared-auth boundary changes and isolation/regression support are included alongside the extraction. This update is description/discussion only; no implementation changes or review resolutions are being made.

rebova-microsoft:
@apurvabanka I dug into the remaining points and found a real issue with the auth diagnostics: the underlying cause was preserved in Python, but the MCP response only showed the generic outer message.

Addressed that in b559173. Known credential failures now explain what needs attention without exposing paths, tokens, or raw provider messages. The local sign-in timeout also has its own message. Account selection and retry behavior are unchanged.

The importer concerns didn't reproduce as bugs: both helper imports share the same provider-module cache, and partial-load retries behave consistently. I'm leaving that code alone rather than adding a refactor to this PR.

The temporary CI filters and their matching assertion are still scheduled for removal before release promotion. Ready for another look when you have a chance.

dcowen91 (approved):

apurvabanka (approved):

PR #285, opened 2026-09-15
feat(setup): add durable native ALM import

nkemms:
## Summary

This change adds a durable native package-import path for creating a Dev agent or explicitly replacing one exact target. It separates the remote package mutation from local workspace attachment and makes interrupted operations safe to resume.

## Description

A maker can choose a package, confirm a create or exact-target replacement, and retain enough operation evidence to verify the result without replaying the mutation. Replacement is limited to the confirmed agent, redirects and automatic mutation retries are disabled, and successful imports hand off to the existing Dev attachment path. The package-import route remains an advanced explicit option rather than the default setup choice.

## Outcome

Ready for review after #276. The change is replay-safe, preserves setup provenance, and keeps package mutation separate from local workspace projection.

## Testing

Live testing covered fresh creation, exact-target replacement, retained-identity verification, resume from a saved operation receipt without a second import, and successful Dev workspace attachment. Focused import, attachment, routing, and AgentBuilder tests passed locally; GitHub checks pass except for the same two Linux filename-casing assertions corrected in #287, where the combined stack is green.


amilandi (dismissed):

srideshpande (approved):
Changes are already reviewed. Approving as it is a new iteration.

PR #287, opened 2026-09-15
feat(setup): add MOS starter import path

nkemms:
## Summary

This change lets a maker with no existing agent choose an entitled Employee Self-Service starter and create a new editable Dev agent. It completes the setup stack with guarded creation, workspace-local sign-in reuse, occupied-workspace routing, and reliable resume behavior.

## Description

Setup reads the native starter catalog, presents the available products, and creates only the package the maker explicitly confirms. The create request is protected against accidental replay, the agent's ALM setting is persisted and verified, and the result is handed to the shared Dev attachment path. If the current workspace already contains an agent, setup offers to continue with it, create a sibling workspace for another agent, or cancel; known older setup state is migrated without requesting a redundant agent URL.

## Outcome

Ready for review. Review the stack in order: #277 → #276 → #285 → #286 → #287; this PR is the shareable entry point for the complete setup journey.

## Testing

Live testing confirmed entitled-package discovery, HTTP 201 starter creation with returned agent identity, persisted ALM configuration, and a final attached workspace reporting `workspace-ready` and `connectReady: true`. Existing-agent resume also confirmed one-time cached-account selection, direct recorded-target validation, legacy-state migration, and workspace refresh. All current GitHub checks pass; the lifecycle suite completed with 570 tests passing and one platform-specific symlink test skipped, and CodeQL passed.


amilandi:
Reviewed with a specific eye on setup robustness/consistency. I've noticed some issues with this skill recently. This PR addresses it directly, not just as a side effect of adding the MOS starter path. Highlights:

- **Occupied-workspace gate** on `/setup` with explicit Continue / new-workspace / Cancel, no preselect.
- **New workspace = detached Git worktree** in a sibling folder; refuses to run on a dirty tree and never overwrites in place.
- **One-time account selection** via new `cached-accounts` subcommand, threaded as `--account` through every subsequent script. No more repeat MS sign-in prompts.
- **Explicit Resume vs. Continue** label depending on canonical-state completeness.
- **Legacy schema-1 completed state auto-migrates** instead of hard-erroring.
- **"Never infer progress from conversation history"** rule + host-native task list.

MOS starter path itself looks good: mandatory confirmation before create, ALM enable followed by read-back verification, uncertain-response branch stops setup instead of guessing, attach reuses the existing `workspace-ready` / `connectReady: true` completion contract.

Non-blocking questions:
1. Occupied-workspace gate treats `.local/config.json` as a signal — will a `/flightcheck`-only workspace trip it?
2. `prepare_fresh_workspace.py` requires a clean git tree; consider surfacing that expectation in the maker-facing copy so a maker with edited topic YAMLs isn't surprised after choosing **Create and open a new workspace**.
3. Legacy migration only covers *completed* schema-1 state — was in-progress schema-1 ever writable?

LGTM otherwise.

amilandi (dismissed):

amilandi (approved):

PR #289, opened 2026-09-17
ESS Diagnostics: interactive transcript-diagnosis skill (replaces Python tool)

RamananVr:
## Summary

Replaces the standalone ESS Diagnostics **Python tool** with a single self-contained **Claude skill** (`tools/ESS-Diagnostics-Skills/SKILL.md`) that walks a Field Engineer (FDE) through a Copilot Studio / PVA transcript interactively and produces evidence-cited outputs. The entire Python pipeline (`src/`, `tests/`, `pyproject.toml`) is removed.

### What the skill does
- **Step 0 — hard gate:** stops and asks the FDE to confirm the transcript file path and state the specific problem before parsing anything (won't proceed without a problem statement).
- **Per-turn interactive walkthrough:** for each turn it shows a structured summary (intent, search query + rewrite, and **all** search results with their source text) plus a one-line verdict strip, then **hard-stops** with a section menu. The FDE can **drill into any section** (full evidence + the check's reasoning), **override a verdict** (with reason, carried into the report), `continue`, or `run all` to finish without pausing.
- **5 checks:** intent recognition, search query issued (+ rewrite), search-result topical relevance, knowledge grounding consistency, final answer vs. retrieved content/guardrails. Divergence = first failing check; a root cause is composed from it.
- **Real-schema parse map:** derived from real transcript exports — events are `Trace` discriminated by `p.data.kind` (`LlmIntentRecognized`, `PluginStart`, `PluginResponse`, `AnalyticsAiMetricsSignalTraceData`, `KnowledgeTraceData`); user turns are `SynchronousIncomingActivity`; grounding/rewrite come from `completionState` / `rewrittenMessage` / `verifiedSearchResults[].rankScore`. Missing fields degrade to N/A rather than guessing.

### Outputs (three files, written to the OS temp dir — never the repo, since transcripts contain employee PII)
1. **`<name>-debug-report.md`** — human-readable diagnosis: per-turn verdicts with evidence, FDE overrides, divergence point, root cause.
2. **`<name>-normalized.json`** — compact **diagnostic** view (only the fields the checks use; no verdicts).
3. **`<name>-transcript.json`** — **faithful, lossless** full transcript converted to JSON (every event and field preserved, `_turn`/`_index` annotated), produced by the bundled helper `scripts/transcript_to_json.py`.

### Transcript helper is Python (ported from Node.js)
The bundled faithful-dump helper is `scripts/transcript_to_json.py` — a stdlib-only Python script, so the toolkit stays Python-only with no Node.js dependency. It began as a Node.js helper (`transcript-to-json.js`) and was ported to Python within this PR as an **exact behavioral drop-in**: identical CLI contract (`<transcript.txt> [outFile.json]`), same `_turn`/`_index` annotation, same temp-dir output path, same `Error:`/exit-1 failures, and content byte-identical output (`ensure_ascii=False` preserves raw Unicode). The Node.js version has been removed and `SKILL.md` now invokes the Python helper.

## How to invoke the skill

From a Claude Code session in the repo, either invoke it by name:

```
/ESS-Diagnostics-Skills:diagnose-transcript
```

or point Claude at it directly:

> Use the ESS diagnostics skill on `C:\path\to\Transcript_xxxx.txt`

The skill then (1) stops and asks for the file path + the specific problem, (2) parses and walks each turn — pausing so you can drill in, override a verdict, `continue`, or `run all`, and (3) writes the three output files to your OS temp dir and reports their absolute paths.

To convert a transcript to faithful JSON directly (also used internally by the skill) — use `python` on Windows, `python3` on POSIX:

```
python tools/ESS-Diagnostics-Skills/scripts/transcript_to_json.py "<path-to-transcript.txt>"
```

## Validation

Dry-run against a real (redacted) transcript confirmed the Step 0 gate, per-turn hard stops, real-schema extraction, drill-down/override, and temp-dir outputs. The run surfaced a genuine **search relevancy/ranking defect** — the correct KB was retrieved but under-ranked (all `rankScore=0`) and not cited — validating that the checks + FDE override mechanism work. No automated test suite (the skill is prose); the Python helper was validated manually (run against a real transcript, output diffed byte-for-byte against the original Node.js helper, and error-path spot-checked).

## Notes
- No PII-bearing artifacts are written to the repo; all diagnosis outputs go to the OS temp directory.
- Planning docs live under `tools/ESS-Diagnostics-Skills/docs/plans/`, consolidated into two files that each cover both efforts (the v4 skill and the Node→Python helper port) in a Part 1 / Part 2 split: `2026-09-17-ess-diagnostics-design.md` (combined design) and `2026-09-17-ess-diagnostics-implementation.md` (combined implementation plan).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tools/ESS-Diagnostics-Skills/SKILL.md
```diff
@@ -0,0 +1,389 @@
...
+file), running a 5-check diagnostic on each turn with evidence quoted straight
+from the transcript. It produces an evidence-cited markdown Debug Report plus a
+normalized-transcript JSON artifact. This skill replaces the earlier standalone
+Python diagnostics tool.
+
+## Rules
+
+- This skill is **read-only**. The ONLY writes it may make are the two output
+  files: the markdown Debug Report and the normalized-transcript JSON. Never
+  modify the transcript, agent files, or anything else.
```
AdwaitRaut: Rules here say "the ONLY writes it may make are the two output files," but line 262 says three, line 269 says "both files," and the README says two without ever mentioning `-transcript.json`. An LLM will follow the hard rule and skip the faithful dump while Step 4 promises the FDE three paths. Pick three and fix all four spots.

tools/ESS-Diagnostics-Skills/SKILL.md
```diff
@@ -0,0 +1,389 @@
+# Diagnose an ESS Transcript
```
AdwaitRaut, line 1: Open question, not blocking — how do you see people invoking this? Other skills sit at `src/skills/<name>/` with a `.github/prompts/<cmd>.prompt.md` wiring the slash command; this one's under `tools/` without one, and the README just says "via its slash/skill mechanism." Real slash command later, or point Claude at the path? Either's fine, just worth a README line.

tools/ESS-Diagnostics-Skills/docs/plans/2026-09-17-ess-diagnostics-implementation.md
```diff
@@ -0,0 +1,415 @@
...
+
+**Step 2:** CRLF-normalize, verify, `git add`, commit: `docs: update README for v4 interactive + real-schema behavior`.
+
+---
+
+### Task 7: Dry-run validation against the real transcript
+
+**Files:** none (validation only).
+
+**Step 1:** Follow the revised SKILL.md against `C:\Users\rarame\Downloads\Transcript_a9b8b841-6a4f-4675-9e79-92fc6df74be9.txt`. Confirm:
```
AdwaitRaut: Minor: real alias still in the example paths (`C:\Users\rarame\...`, also 306/316/317) — swap in a placeholder next time you're in here.

AdwaitRaut (commented):
The per-turn walkthrough is solid. N/A-on-failed-precondition and the FDE override path are the right calls.

One fix before merge: the output contract disagrees with itself across the Rules block, Step 4, and the README (two files vs three). Open question on invocation plus a few nits below — none of those need to hold it up.

tools/ESS-Diagnostics-Skills/scripts/transcript_to_json.py
```diff
@@ -0,0 +1,130 @@
...
+        out_dir = Path(tempfile.gettempdir()) / "ess-diagnostics" / base
+        try:
+            out_dir.mkdir(parents=True, exist_ok=True)
+        except OSError as e:
+            fail("could not create output directory " + str(out_dir) + ": " + str(e))
+        out_file = out_dir / (base + "-transcript.json")
+
+    try:
+        out_file.write_text(
+            json.dumps(annotated, indent=2, ensure_ascii=False), encoding="utf-8"
```
AdwaitRaut, line 139: Non-blocking: on POSIX `gettempdir()` is `/tmp`, so you get 0755/0644 and other local users can read the output. Windows `%TEMP%` is per-user, so the documented path is fine — but since PII is the reason for writing here, `chmod` 0700/0600 on POSIX would close it.

tools/ESS-Diagnostics-Skills/scripts/transcript_to_json.py
```diff
@@ -0,0 +1,130 @@
+#!/usr/bin/env python3
```
AdwaitRaut, line 1: Missing the standard header (everything else under `tools/` has it) — `# Copyright (c) Microsoft Corporation.` / `# Licensed under the MIT License.`, below the shebang.

tools/ESS-Diagnostics-Skills/scripts/transcript_to_json.py
```diff
@@ -0,0 +1,130 @@
...
+            annotated.append({"_turn": turn, "_index": i, **e})
+        else:
+            annotated.append({"_turn": turn, "_index": i, "_value": e})
+
+    base = src_path.name
+    dot = base.rfind(".")
+    if dot > 0:
+        base = base[:dot]
+
+    if len(argv) >= 3 and argv[2]:
```
AdwaitRaut, line 104: Two small ones, take or leave: `out_file` is never checked against `src_path`, so passing the transcript as the output arg overwrites it. And the lossless claim doesn't quite hold — `{"_turn": t, **e}` lets an event's own `_turn` win (line 91), and non-dicts come back wrapped in `_value`. Neither shows up in a real export, so only the first looks worth fixing.

tools/ESS-Diagnostics-Skills/SKILL.md
RamananVr: Fixed in 0ec9a3a. Standardized on three files everywhere: the Rules block now lists all three (Debug Report, normalized JSON, faithful transcript JSON), Step 4 already said "three," and I changed "both files" -> "all three." The README's Outputs list and read-only guarantee now say three and document `-transcript.json`. Good catch - an LLM would've followed the hard rule and skipped the faithful dump.

tools/ESS-Diagnostics-Skills/SKILL.md
RamananVr: Added a README line: for now it's "point Claude at the skill dir" (no wired slash command). Left a real .github/prompts slash command as a follow-up rather than in this PR.

tools/ESS-Diagnostics-Skills/docs/plans/2026-09-17-ess-diagnostics-implementation.md
RamananVr: Done — swapped the real user path prefix for a placeholder (`C:\path\to\`) across all occurrences in the implementation plan (0ec9a3a).

tools/ESS-Diagnostics-Skills/scripts/transcript_to_json.py
RamananVr: Added in 0ec9a3a: on POSIX the per-run temp dir is chmod 0700 and the output file 0600; no-op on Windows (where %TEMP% is already per-user). Since PII is the reason for the temp dir, agreed this was worth closing.

tools/ESS-Diagnostics-Skills/scripts/transcript_to_json.py
RamananVr: Added `# Copyright (c) Microsoft Corporation.` / `# Licensed under the MIT License.` below the shebang, matching the other tools (0ec9a3a).

tools/ESS-Diagnostics-Skills/scripts/transcript_to_json.py
RamananVr: Fixed the overwrite risk in 0ec9a3a — it now refuses to write when the output path resolves to the source transcript. Left the `_turn` collision / `_value` wrapping as-is since neither appears in a real export, as you noted.

AdwaitRaut (approved):

